PR #15, opened 2022-04-19
Add more examples to the documentation

pablogsal:
Add more example files to the documentation so users can see different
flamegraph files so they can evaluate the tool in different scenarios
that involve native allocations.


godlygeek (approved):
LGTM (needed to fix the import order for isort and I fixed up "SQlite" to "SQLite" in one place)

godlygeek (approved):
LGTM

PR #16, opened 2022-04-20
docs: Improve documentation structure

godlygeek:
In addition to some rephrasing and polishing of the wording of the
documentation, overhaul the structure. Remove the dedicated pages
showing how to run each of the CLI commands, instead merging them into
a section on the most closely related page. Create a new page for the
`run` subcommand that describes all its possible modes, but link to
a separate page for detailed descriptions about "live" mode. Move the
"live mode" page into the reporters section, since it's essentially
a reporter.


godlygeek:
Docs generated from this PR can be viewed [here](https://godlygeek.github.io/memray/getting_started.html), which will certainly be much easier than trying to read this diff 😅

Issue #12, opened 2022-04-18
Should we have a `memray` entry point?


godlygeek:
Currently we only have `memray3.X` as an entry point - should we also have `memray`? This will be more confusing for people with multiple copies of `memray` installed, but easier for people with only one, and it would make the usage examples from `-h` line up with how we actually expect it to be used.

pablogsal:
Yeah, I find a bit annoying having to type memray3.x all the time if I am in a virtual environment. So I am +1 on thos

PR #20, opened 2022-04-20
Add a `memray` entry point

pablogsal:
Closes: #12

Currently we only have `memray3.X` as an entry point but having also
`memray` as an entry point will make it easier for people with only one
Python version installed and it would make the usage examples from -h
line up with how we actually expect it to be used.

This commit also updates the docs and the README file to simplify the
examples with the new entry point.


godlygeek (approved):
LGTM - I removed some no longer needed parentheticals, and spotted a place where we were unnecessarily using implicit string literal concatenation in `setup.lpy`

PR #21, opened 2022-04-20
Use a version of the example SQLITE flamegraph that has symbols

pablogsal:
The old example that we were using for the SQLITE code was executed in a
system without symbols for libsqlite, making not possible to resolve
some entries in the flame graph.


godlygeek (approved):
Ooh yeah, it's much cooler when you can see the b-tree rebalancing!

PR #31, opened 2022-04-21
Fetch line numbers for Python frames lazily

godlygeek:
Up until now we've been greedily capturing the line number of each
Python frame as we first encounter it, but this is unnecessarily
inefficient. Since we've already been assuming that it's safe to call
`PyFrame_GetLineNumber` without the GIL held (and we'd already need to
change this code should that assumption become false in a future CPython
version), we can delay calculating the Python line number until we're
about to emit the frame. Since many frames are entered without ever
being emitted (because no allocation is performed under them), this
drastically reduces the number of calls to `PyFrame_GetLineNumber`,
which is a deceptively slow function.

### DIFF
--- a/src/memray/_memray/tracking_api.cpp
+++ b/src/memray/_memray/tracking_api.cpp
@@ -80,11 +80,17 @@ class PythonStackTracker
     {
     }
 
+    enum class FrameState {
+        NOT_EMITTED = 0,
+        EMITTED_BUT_LINE_NUMBER_MAY_HAVE_CHANGED = 1,
+        EMITTED_AND_LINE_NUMBER_HAS_NOT_CHANGED = 2,
+    };
+
     struct LazilyEmittedFrame
     {
         PyFrameObject* frame;
         RawFrame raw_frame_record;
-        bool emitted;
+        FrameState state;
     };
 
   public:
@@ -96,10 +102,8 @@ class PythonStackTracker
     void clear();
 
     static PythonStackTracker& get();
-    void emitPendingPops();
-    void emitPendingPushes();
-    int getCurrentPythonLineNumber();
-    void setMostRecentFrameLineNumber(int lineno);
+    void emitPendingPushesAndPops();
+    void invalidateMostRecentFrameLineNumber();
     int pushPythonFrame(PyFrameObject* frame);
     void popPythonFrame(PyFrameObject* frame);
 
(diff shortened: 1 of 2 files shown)


godlygeek:
Flipping this to Draft - this is subtly broken in a way that our test suite doesn't catch. I'll need to circle back to this - and add a new test...

godlygeek:
OK, I think this is now _actually_ correct... And the new test case covers the problem that the original version of this had, so that we know we won't regress on it.

pablogsal:
Experiments executed for `python -m memray run  -o /dev/null -m test test_list test_list test_list test_list test_list test_list test_list test_list test_list test_list` for a total duration of 1.2595737380001992 seconds in baseline.

## With this PR

```
      Calls to PyFrame_GetLineNumber: 25550
     0.03%     0.03%  python  python3.10  [.] PyFrame_GetLineNumber                                                                                ▒
```

## Baseline


```
      Calls to PyFrame_GetLineNumber: 270090
     0.39%     0.39%  python  python3.10                               [.] PyFrame_GetLineNumber                                                   ◆
```

This means that in general the most we can save (that we are indeed saving it) is 0.39% of the total runtime (this is assuming no `PYTHONMALLOC=malloc` where most savings will be there since calls to `PyFrame_GetLineNumber` in `pushPythonFrame` are a bigger percentage of the runtime because our hooks are called less).

For the case where `PYTHONMALLOC=malloc` for a total duration of 3.0231499819997225 seconds in baseline:


```
      Calls to PyFrame_GetLineNumber: 21057170
+    5.26%     5.26%  python  python3.10                               [.] PyFrame_GetLineNumber
```

## Baseline

```
      Calls to PyFrame_GetLineNumber: 21125440
+    7.23%     7.23%  python  python3.10                               [.] PyFrame_GetLineNumber
```

This means that in the latter we are saving ~2% maximum. This is consistent with the timing results of `2.8712007330000233` (PR) against `2.935535135000009` (baseline).

## PR

```
  Test duration: 2.8712007330000233
  Time (mean ± σ):      3.058 s ±  0.051 s    [User: 3.008 s, System: 0.056 s]
  Range (min … max):    2.983 s …  3.167 s    10 runs
```

## Baseline

```
    Test duration: 2.935535135000009
  Time (mean ± σ):      3.120 s ±  0.046 s    [User: 3.070 s, System: 0.056 s]
  Range (min … max):    3.064 s …  3.210 s    10 runs
```

That is around 3.5% of the difference.

pablogsal (approved):

PR #37, opened 2022-04-21
Remove duplicated words

jugmac00:


godlygeek:
Thanks! Per our [contribution licensing](https://github.com/bloomberg/memray/blob/main/README.md#contribution-licensing) documentation, would you mind adding a `Signed-Off-By:` tag to your commit message to indicate that you have read and agree to our [Developer's Certificate of Origin](https://github.com/bloomberg/.github/blob/main/DCO.md)?

That should be as easy as doing:
```shell
$ git commit --amend -C HEAD -s
$ git push -f
```
from a checkout of this branch.

jugmac00:
@godlygeek Sorrý, must have missed the requirement. -> updated my PR

godlygeek:
OK, sorry, one more thing...

You've made two changes. One of them, to `src/memray/_memray/elf_shenanigans.cpp`, looks perfectly good and I'm happy to accept it.

The other one is to `src/vendor/libbacktrace/ltmain.sh`. That file isn't part of Memray's code base, exactly - it's a vendored copy of the `libbacktrace` code from here: https://github.com/ianlancetaylor/libbacktrace/blob/4d2dd0b172f2c9192f83ba93425f868f2a13c553/ltmain.sh#L570

I'd rather not patch our copy of their code for something so minor as a typo, so I'd like you to remove that change and only make the first one (or I can do that for you, if you'd like).

I'm guessing that even the owners of `libbacktrace` won't want to accept that change, because that file seems to be generated code: https://github.com/ianlancetaylor/libbacktrace/blob/4d2dd0b172f2c9192f83ba93425f868f2a13c553/ltmain.sh#L1

That was generated by something in https://git.savannah.gnu.org/cgit/libtool.git/tree/ but as far as I can see that typo no longer exists in the latest version of libtool, so I suspect that this has already been fixed upstream, and just needs to trickle down into other projects over time. https://git.savannah.gnu.org/cgit/libtool.git/tree/bootstrap#n1369 seems like it might be the latest version of that code, and it doesn't contain the typo.

jugmac00:
Updated once more :-)

pablogsal (approved):

pablogsal:
Thanks @jugmac00 for your contribution 🎉

Issue #25, opened 2022-04-21
Add documentation on how to use memray as an API


pablogsal:
One of our coolest features is being able to use `memray` as an API. We should add documentation:

- [ ] Add API docs with all the relevant classes and enums.
- [ ] Add A tutorial on how to use the `Tracker`, `FileReader` and other utilities.
- [ ] Update the `README`.

corleyma:
Seconding the desire for this.  I'm eager to integrate memray as a profiling option in our python workflow orchestrator, so any guidance on API usage would be a big help.

pradyunsg:
For the folks interested and eager to try things out before @godlygeek has a chance to write down the API docs, you can look at the tests we have, for the public tracking API: https://github.com/bloomberg/memray/blob/main/tests/integration/test_api.py

The `FileDestination` will write to a file, which you can then pass to the various reporters on the CLI to generate the reports. Do make sure that the file passed into the `FileDestination` doesn't exist already. :)

PR #42, opened 2022-04-21
Finalize and document Tracker API

godlygeek:
Closes: #25 

Rendered documentation can be previewed [here](https://godlygeek.github.io/memray/api.html).

@pablogsal Let me know whether there's anything else that you'd like to see documented here. 

Now is probably our last chance to rename some of this before the API actually gets users. I'm wondering whether we should have gone with `force=True` or `overwrite=True` for the parameter to tell a `FileDestination` to overwrite an existing file, instead of `exist_ok=True`. I'm also thinking that the name `SocketDestination` is a bit odd for something that configures how to start a _server_, rather than a _client_...

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,19 @@ By default, the live command will present the main thread of the program. You ca
 
 <img src="https://github.com/bloomberg/memray/blob/main/docs/_static/images/live_different_thread.png?raw=true" align="center"/>
 
+# API
+
+In addition to tracking Python processes from a CLI using `memray run`, it is also possible to programmatically enable tracking within a running Python program.
+
+```py
+import memray
+
+with memray.Tracker("output_file.bin"):
+    print("Allocations will be tracked until the with block ends")
+```
+
+For details, see the [API documentation](https://bloomberg.github.io/memray/api.html).
+
 # License
 
 Memray is Apache-2.0 licensed, as found in the [LICENSE](LICENSE) file.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,0 +1,24 @@
+.. module:: memray
+
+Memray API
+==========
+
+Memray exposes an API that can be used to programmatically activate or
+deactivate tracking of a Python process's memory allocations. You do this by
+creating a `Tracker` object and using it as a context manager in a ``with``
+statement. While the body of the ``with`` statement runs, tracking will be
+enabled, with output being sent a destination you specify when creating the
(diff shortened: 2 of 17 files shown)


pablogsal:
> @pablogsal Let me know whether there's anything else that you'd like to see documented here.

I wonder if we want to document the `FileReader` and related classes, as they are part of the API and those are the ones that allow extension. On the other hand is unclear if we want to freeze all these APIs just now...

pablogsal:
> I'm wondering whether we should have gone with `force=True` or `overwrite=True` for the parameter to tell a `FileDestination` to overwrite an existing file, instead of `exist_ok=True`. I'm also thinking that the name `SocketDestination` is a bit odd for something that configures how to start a _server_, rather than a _client_...

I agree with both, those are good points. What about if we rename `exists_ok` to `overwrite`?

Regarding `SocketDestination` is a bit trickier because we are indeed sending this to a socket destination, but the whole enterprise then makes the whole program act as a server. I think that is ok-ish unless some other native speakers think otherwise.

godlygeek:
> I wonder if we want to document the `FileReader` and related classes, as they are part of the API and those are the ones that allow extension.

I've gone with "no" for now, in part because we haven't yet gotten any requests for this, but also because I'm not as familiar with those classes and I can speak as well to whether we're willing to commit to interface stability on them, or whether there's things that still need to be revisited.

> Regarding `SocketDestination` is a bit trickier because we are indeed sending this to a socket destination,

Yes, but the destination that we send to isn't identified by the host and port that you give. You do `Tracker(SocketDestination(port=1234))`, and we don't send any data to port 1234. Instead, we listen on port 1234 for an incoming connection, which gets assigned an ephemeral port, and we send data to that ephemeral port instead.

I dunno - I find the name a bit misleading, but I also can't really think of a way to fix it. I guess we could call it `SocketServer` instead...

> I agree with both, those are good points. What about if we rename `exists_ok` to `overwrite`?

Ack, I'll make that change.

pablogsal:
> I guess we could call it SocketServer instead...

`SocketServer` works for me

godlygeek:
After some offline discussion, we've settled on leaving `SocketDestination` for the name of that class, but changing its constructor parameters to make it clearer that it's spawning a server, renaming
```py
class memray.SocketDestination(port: int, host: Optional[str] = '127.0.0.1')
```
to
```py
class memray.SocketDestination(server_port: int, address: Optional[str] = '127.0.0.1')
```

pablogsal (approved):

Issue #30, opened 2022-04-21
Make memray reproducible


inglor:
## Bug Report

**Current Behavior**
memray as it stands is not reproducible [1]

**Input Code**

The below are the reproducible tests Arch Linux infrastructure automatically perform on each package (periodically). As you see from the diffscope the Build-ID is different on the .so package.

https://reproducible.archlinux.org/api/v0/builds/267424/diffoscope
https://reproducible.archlinux.org/api/v0/builds/267424/log

**Expected behavior/code** 
Expected behavior for this is for memray to be reproducible (byte-for-byte) 

**Environment**

Package is build according to PKGBUILD from Arch Linux repos [2]

**Possible Solution**

Not entirely sure but I think this is because it's build from within cython and doesn't read the Build_Date or something

**Additional context/Screenshots**

https://reproducible-builds.org/docs/source-date-epoch/

[1]: https://reproducible-builds.org
[2]: https://archlinux.org/packages/community/x86_64/memray/

pablogsal:
I have investigated and seems that the reason the build ids are different is due to some Cython symbols being migrated when pip does an isolated build as the path is different. Everything else seems the same:
![Screenshot 2022-04-21 at 12 15 13](https://user-images.githubusercontent.com/11718525/164448210-ba835534-9a2d-4f1a-ab0b-70c7334a224c.png)
![Screenshot 2022-04-21 at 12 15 25](https://user-images.githubusercontent.com/11718525/164448219-5ec92d8f-8a6d-4f28-94fb-f9b446cedfab.png)
![Screenshot 2022-04-21 at 12 15 33](https://user-images.githubusercontent.com/11718525/164448226-db2ba8f1-87a6-436b-816b-b46df335b1c7.png)

inglor:
Hmm seems like this is a bug for `cython` then instead of `memray`. Should I raise an issue somewhere else and link it here?

pablogsal:
> Hmm seems like this is a bug for `cython` then instead of `memray`. Should I raise an issue somewhere else and link it here?

I am not sure is a bug in Cython. The problem is that when `gcc` builds DWARF it burns the source location and the source location is different with `pip` isolated installation because it uses a different temporary directory. This problem should also appear with the vendored patched `libbacktrace` as the build system builds it in a temporary directory but for some reason that's not a problem, it seems (likely because it doesn't include DWARF by default).

inglor:
[...]

PR #47, opened 2022-04-22
Add `--wheel` to the `make dist` target

pablogsal:
Closes: #30

When we build the distribution without "--wheel", `build_ext` gets
called during "python -m build" from the temporary build directory,
while with "--wheel" the files built during "make build" are copied into
the created wheel. This allows the build to be more reproducible as in
this way the DWARF included in the extension to not change due to the
temporary directory path being part of it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ build-js: $(generated_js_files)  ## Build package assets in-place
 
 .PHONY: dist
 dist:  ## Generate Python distribution files
-	$(PYTHON) -m build .
+	$(PYTHON) -m build --wheel .
 
 .PHONY: install-sdist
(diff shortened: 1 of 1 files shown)


lkollar (approved):

Makefile
```diff
@@ -42,7 +42,7 @@ build-js: $(generated_js_files)  ## Build package assets in-place
 
 .PHONY: dist
 dist:  ## Generate Python distribution files
-	$(PYTHON) -m build .
+	$(PYTHON) -m build --wheel .
```
pradyunsg, line 45: You likely need a `--sdist` as well, otherwise this will only generate a wheel.

PR #58, opened 2022-04-24
Reduce result file size by 20% in non-native mode

pablogsal:
Here are a bunch of optimizations that reduce the result file size by 20%:

- Reduce file size in non-native mode by specializing native allocations
- Ensure our enumerations only use the size of a char

### DIFF
--- a/src/memray/_memray/record_reader.cpp
+++ b/src/memray/_memray/record_reader.cpp
@@ -169,7 +169,7 @@ bool
 RecordReader::parseNativeFrameIndex()
 {
     UnresolvedNativeFrame frame{};
-    if (!d_input->read(reinterpret_cast<char*>(&frame), sizeof(UnresolvedNativeFrame))) {
+    if (!d_input->read(reinterpret_cast<char*>(&frame), sizeof(frame))) {
         return false;
     }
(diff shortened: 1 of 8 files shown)


src/memray/_memray/record_reader.cpp
```diff
@@ -194,7 +194,31 @@ RecordReader::parseAllocationRecord()
...
+            .frame_index = stack.empty() ? 0 : stack.back(),
+            .native_segment_generation = d_symbol_resolver.currentSegmentGeneration()});
+    return true;
+}
+
+bool
+RecordReader::parseNativeAllocationRecord()
+{
+    NativeAllocationRecord record;
+    if (!d_input->read(reinterpret_cast<char*>(&record), sizeof(NativeAllocationRecord))) {
```
godlygeek, line 211: I'd prefer `sizeof(record)` since we're modifying this code - the `(&obj, sizeof(obj))` pattern is recognizable and eliminates the risk of updating the type of `obj` in one place and forgetting to update the `sizeof` to match.

godlygeek (changes requested):
Looks good, but I see a few minor nits to fix

godlygeek (approved):
LGTM

Issue #43, opened 2022-04-22
Crashes while processing large output binary


delica1:
We have an application which uses a lot of memory > 30GB and we're trying to understand where the memory is used. The output binary file is 150GB+ and when try to create a table /HTML / summary it crashes with out of memory errors because it seems to try to process the entire large file in memory at once.

godlygeek:
Well, that's a fun stress test. Yeah, I'm guessing that the `d_allocation_records.reserve(stats.n_allocations);` call in `RecordReader::RecordReader` is failing due to the large number of allocated records.

So... we'll have to see what can be done about that.

godlygeek:
If you still have that file, would you kindly mind running:
```
memray parse output.new.bin | awk '{print $1}' | sort | uniq -c
```
for me?

That'll read the output file and give us a count of different types of records in it. (`memray parse` is a (currently) undocumented command that pretty-prints the contents of a binary capture file as plain text; there are no guarantees about its format remaining stable across releases, it is meant as a debugging tool.)

I'm curious exactly how many allocations we're dealing with here.

godlygeek:
I wouldn't be shocked if `memray run` takes several minutes to run on a 150 GB file, either.

delica1:
took about 2.5 hours :)

2546564914 ALLOCATION
  19631 FRAME_ID
820219592 FRAME_POP
862137699 FRAME_PUSH
      1 HEADER
 174981 MEMORY
      1 THREAD

godlygeek:
A new version was released yesterday that both writes smaller output files and doesn't attempt to load all allocation records into memory simultaneously. Give Memray 1.1.0 a shot (`pip install --upgrade memray` if you're using `pip`) and see if things work better for you now.

rski:
I have a 50mb binary file, doing memray flamegraph on it always crashes like this:

```
read(4, "\"\"\"\nRead and write ZIP files.\n\nX"..., 8192) = 8192
close(4)                                = 0
write(2, "    import binascii\n", 20    import binascii
)   = 20
write(2, "ImportError: /usr/lib64/python3."..., 152ImportError: /usr/lib64/python3.9/lib-dynload/binascii.cpython-39-x86_64-linux-gnu.so: failed to map segment from shared object: Cannot allocate memory
) = 152
[...]

PR #59, opened 2022-04-25
Record reader refactoring

godlygeek:
Relates-to: #43 

Some initial refactoring of the record reader to get us closer to being able to process records in a streaming fashion, without needing to hold all allocations in memory simultaneously.

We can land this as is, or hold off and base further work off of this.


src/memray/_memray.pyx
```diff
@@ -296,10 +296,22 @@ cdef class FileReader:
 
     cdef void _populate_allocations(self):
         cdef RecordReader* reader = self._get_reader()
-        while reader.nextRecord() not in (
-                RecordResult.RecordResultEndOfFile,
-                RecordResult.RecordResultError):
-            continue
+
+        stats = self._header["stats"]
+        self._allocations.reserve(stats["n_allocations"])
```
pablogsal, line 301: I think we have changed this back and forth from the C++ to the Cython later already 3 times 😅

src/memray/_memray/snapshot.cpp
```diff
@@ -143,68 +143,66 @@ reduceSnapshotAllocations(const allocations_t& records, size_t snapshot_index, b
     return aggregator.getSnapshotAllocations(merge_threads);
 }
 
-HighWatermark
-getHighWatermark(const allocations_t& records)
+void
+HighWatermarkFinder::updatePeak(size_t index)
```
pablogsal, line 147: You can mark this as `noexcept`

pablogsal (approved):
LGTM

I just have one small nit. 👍

pablogsal (approved):

PR #61, opened 2022-04-25
Add `-c` option: program passed as string

### LINKED ISSUE #44: Add run -c mode to allow program passed in as string
## Feature Request

`python3` has `-c` option which allows passing an arbitrary program string. This is useful for quick ad-hoc experiments in the shell (especially with `--live` mode).

E.g. `python3 -c 'print(len([str(i) for i in range(1000000000)]))'`

Would be nice if `memray run` had a similar mode?

**Describe the solution you'd like** 

`memray run -c` could have the same logic as `python3 -c`

**Describe alternatives you've considered** 

Currently the only alternative seems to put the program in a temporary script and run via `memray run script.py`

Perhaps there is some standard Python module one can use to run a program string so we could reuse the `-m` option, but I don't know of such module.

Another possible alternative is to use process substitution to avoid a temporary file, but it doesn't seem to work: 

```
$ memray run <(echo 'print(len([str(i) for i in range(1000000000)]))')
Writing profile results into /dev/fd/memray-63.945915.bin
Could not create output file /dev/fd/memray-63.945915.bin: No such file or directory
```
-- fair enough, but if we try `memray run -o output.bin <(echo 'print(len([str(i) for i in range(1000000000)]))')`, it seems to exit immediately without any captures, as if it didn't wait for the script to finish writing.

(it works as expected for python3)
```
$ python3  <(echo 'print(len([str(i) for i in range(10000000)]))')
10000000
```

P.S. thanks for the great tool! I was seriously impressed that it 'just worked' after `pip install` :)

tal66:
Hi, this closes #44.

can now run: 
- `memray run -c 'print("my python commands")'`  -> a py file will be created with this string as its content. then `arg.script` will be replaced with this file's name. after that, memray runs as before.
- `memray run -c <my_file>.py` -> parser.error with help message to remove `-c` option

(EDIT: no longer creating a file)

### DIFF
--- a/src/memray/commands/run.py
+++ b/src/memray/commands/run.py
@@ -134,7 +134,7 @@ def _run_with_file_output(args: argparse.Namespace) -> None:
     if args.output is None:
         script_name = args.script
         if args.run_as_cmd:
-            script_name = "cmd_script"
+            script_name = "string"
 
         output = f"memray-{os.path.basename(script_name)}.{os.getpid()}.bin"
         filename = os.path.join(os.path.dirname(script_name), output)
@@ -268,7 +268,7 @@ class RunCommand:
             parser.error("The --live-port argument requires --live-remote")
         if args.follow_fork is True and (args.live_mode or args.live_remote_mode):
             parser.error("--follow-fork cannot be used with the live TUI")
-        if args.run_as_cmd and args.script.endswith(".py"):
+        if args.run_as_cmd and pathlib.Path(args.script).exists():
             parser.error("remove the option -c to run a file")
 
         self.validate_target_file(args)
--- a/src/memray/commands/run.py
+++ b/src/memray/commands/run.py
@@ -51,6 +51,8 @@ def _run_tracker(
         try:
             if args.run_as_module:
                 runpy.run_module(args.script, run_name="__main__", alter_sys=True)
+            elif args.run_as_cmd:
+                exec(args.script, {"__name__": "__main__"})
             else:
                 runpy.run_path(args.script, run_name="__main__")
         finally:
@@ -62,6 +64,7 @@ def _child_process(
     port: int,
     native: bool,
     run_as_module: bool,
+    run_as_cmd: bool,
     quiet: bool,
     script: str,
     script_args: List[str],
@@ -69,6 +72,7 @@ def _child_process(
     args = argparse.Namespace(
         native=native,
         run_as_module=run_as_module,
+        run_as_cmd=run_as_cmd,
         quiet=quiet,
         script=script,
         script_args=script_args,
@@ -84,7 +88,7 @@ def _run_child_process_and_attach(args: argparse.Namespace) -> None:
         raise MemrayCommandError(f"Invalid port: {port}", exit_code=1)
 
     arguments = (
-        f"{port},{args.native},{args.run_as_module},{args.quiet},"
+        f"{port},{args.native},{args.run_as_module},{args.run_as_cmd},{args.quiet},"
         f'"{args.script}",{args.script_args}'
     )
     tracked_app_cmd = [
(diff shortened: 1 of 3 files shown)


tal66:
Hi, some tests failed because I changed output to the console. didn't expect it to be tested :)
e.g 
```
assert error:
E         - Only Python files can be executed under memray
E         + Only Python files (or valid Python commands using -c) can be executed under memray
```
I can fix the tests, but see if in general this looks fine to you

godlygeek:
I'd prefer a different approach here. Instead of writing the command to a file, I'd prefer to add another case to `_run_tracker` here:
https://github.com/bloomberg/memray/blob/73e563f464bab5654162bc0c561d52f81ead6eb0/src/memray/commands/run.py#L52-L55

That's where we currently handle either running a file or running a module. I'd add another case for running a command, and do:
```py
            if args.cmd:
                exec(args.cmd, {})
            elif args.run_as_module:
                runpy.run_module(args.script, run_name="__main__", alter_sys=True)
            else:
                runpy.run_path(args.script, run_name="__main__")

**Comment by tal66** [comment]
Hi, can you approve running the workflows on the last push? (used --force)

**Comment by pablogsal** [comment]
> Hi, can you approve running the workflows on the last push? (used --force)

Done!

**Comment by tal66** [comment]
I just changed the signoff now, that was the only problem in the checks

File: src/memray/commands/run.py [hunk 1]
```diff
@@ -51,6 +51,8 @@ def _run_tracker(
         try:
             if args.run_as_module:
                 runpy.run_module(args.script, run_name="__main__", alter_sys=True)
+            elif args.run_as_cmd:
+                exec(args.script, {"__name__": "__main__"})
```
godlygeek, line 55: @pablogsal I don't really love this, but I can't think of any better option. It's a bit weird to set `__name__ = "__main__"` even when `sys.modules["__main__"]` is not being set to a module containing this code, for one thing.

That said, it's probably good enough, and I doubt most code would even notice that it's not _really_ running in the main module. Can you spot any cleaner way to do this?

src/memray/commands/run.py
```diff
@@ -128,8 +132,12 @@ def _run_with_socket_output(args: argparse.Namespace) -> None:
 
 def _run_with_file_output(args: argparse.Namespace) -> None:
     if args.output is None:
-        output = f"memray-{os.path.basename(args.script)}.{os.getpid()}.bin"
-        filename = os.path.join(os.path.dirname(args.script), output)
+        script_name = args.script
+        if args.run_as_cmd:
+            script_name = "cmd_script"
```
godlygeek, line 137: The interpreter calls a command given with `-c` "<string>" - you can see it in tracebacks like:
```
~>python -c "1/0"
  File "<string>", line 1, in <module>
ZeroDivisionError: division by zero
```

I don't think I want to include the `<` and `>` in the name here, since those are shell metacharacters that can cause havoc if people forget to quote them, but I think I'd prefer calling this "string" instead of "cmd_script", to at least somewhat match what the interpreter is doing.
```suggestion
            script_name = "string"
```

src/memray/commands/run.py
```diff
@@ -237,18 +252,24 @@ def validate_target_file(self, args: argparse.Namespace) -> None:
...
+                "Only valid Python files or commands can be executed under memray",
+                exit_code=1,
             )
 
     def run(self, args: argparse.Namespace, parser: argparse.ArgumentParser) -> None:
         if args.live_port is not None and not args.live_remote_mode:
             parser.error("The --live-port argument requires --live-remote")
         if args.follow_fork is True and (args.live_mode or args.live_remote_mode):
             parser.error("--follow-fork cannot be used with the live TUI")
+        if args.run_as_cmd and args.script.endswith(".py"):
```
godlygeek, line 271: I like the user friendliness of this, but it's not technically correct - a command _can_ end with `.py` and still be valid - for instance:
```shell
python -c 'from types import SimpleNamespace as NS; ns = NS(py=1); ns.py'
```
is a perfectly valid command to run. Instead, how about if we use:
```suggestion
        if args.run_as_cmd and pathlib.Path(args.script).exists():
```
to check if the script they've given is the name of a file that exists on disk? Technically that file name could _also_ be a valid Python command to execute, but it's even less likely, and it would also handle cases where the filename doesn't end in `.py`

godlygeek (commented):
Sorry for taking so long to review this. This looks very good to me, though I have a few thoughts about it.

I'm happy to make these changes myself on top of what has already been committed here, but I'm curious to hear other people's thoughts on them.

src/memray/commands/run.py
tal66: could be `string`, `command`, `cmd`, not an issue

src/memray/commands/run.py
tal66: yeah, could happen with a shebang. i use it less nowadays. anyway, exists() doesn't throw exceptions, so seems ok.

godlygeek (approved):
Looks good to me - thank you for the contribution, @tal66!

PR #62, opened 2022-04-25
Decode capture files in two passes

godlygeek:
Instead, read the file twice. The first time through, collect the memory
records and find the high water mark. After that, the user can choose to
do a pass that actually yields allocation records.

Note that with this design it no longer makes sense to warn if the input
file was closed before we reached EOF, since now that's a normal
consequence of retrieving the high water mark allocations: the second
pass over the file closes it once it has read up through the high water
mark index.

Closes: #43

### DIFF
--- a/src/memray/_memray.pyx
+++ b/src/memray/_memray.pyx
@@ -337,7 +337,8 @@ cdef class FileReader:
 
         # Initial pass to populate _header, _high_watermark, and _memory_records.
         cdef shared_ptr[RecordReader] reader_sp = make_shared[RecordReader](
-            unique_ptr[FileSource](new FileSource(self._path))
+            unique_ptr[FileSource](new FileSource(self._path)),
+            False
         )
         cdef RecordReader* reader = reader_sp.get()
 
--- a/src/memray/_memray.pyx
+++ b/src/memray/_memray.pyx
@@ -22,12 +22,15 @@ from _memray.sink cimport SocketSink
 from _memray.snapshot cimport HighWatermark
 from _memray.snapshot cimport HighWatermarkFinder
 from _memray.snapshot cimport Py_GetSnapshotAllocationRecords
+from _memray.snapshot cimport Py_ListFromSnapshotAllocationRecords
+from _memray.snapshot cimport SnapshotAllocationAggregator
 from _memray.socket_reader_thread cimport BackgroundSocketReader
 from _memray.source cimport FileSource
 from _memray.source cimport SocketSource
 from _memray.tracking_api cimport Tracker as NativeTracker
 from _memray.tracking_api cimport install_trace_function
 from libcpp cimport bool
+from libcpp.limits cimport numeric_limits
 from libcpp.memory cimport make_shared
 from libcpp.memory cimport make_unique
 from libcpp.memory cimport shared_ptr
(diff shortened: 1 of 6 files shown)


godlygeek:
With this branch, I'm able to successfully generate a flame graph from a 125 GB output file.

pablogsal:
A couple of comments:

* Can you run some quick benchmarks of this PR against the current main branch when generating flame graphs for less crazy files? I recall that when we moved to hold everything in memory it was quite a significant speedup so we should know how much we are going to miss.

* We could do some optimizations here: the second pass just needs to deal with allocation records. Every other record (like frames and other stuff) can just be skipped. I suspect dealing with the frame map is one of the major operations so we can avoid it in the second pass.

* We should think about showing some progress to the user. I doubt many people will wait for those 20 minutes if we don't tell them what's going on. We could do this in a separate PR thought.

pablogsal:
> That's certainly not the case as things are written today - we're returning `Allocation` objects created by the second `RecordReader` to Python land, and if the user calls `.stack_trace()` on one of them, it accesses the `RecordReader` it was created with, passing it the frame index, which the `RecordReader` looks up in its internal tables.

Maybe I am missing something but here is how I imagine this:

1. We do a first pass, we insert all frame ids, stack traces we construct the frame tree while calculating the *high watermark point*.
2. We already have all the tables and data structures created. Technically, if we return an allocation at this point (which we do not since we are just calculating the high water mark) the stack trace could be resolved because the data is on the reader tables.
3. When we do the second pass to calculate the *high watermatk records* technically we do not need to create the tables again because we already have done that in the first pass. We just need to copy over those vectors, tree and maps to the second reader (if we use a second reader) and just ignore the records we read that are not allocations. Here ignoring the record mean reading them but not inserting them anyware. Having to recreate the frame tree in the second pass is going to be super costly an unnecesary.

Of course, the specifics on how we just skip the records elegantly are to be discussed, but we can have some attribute that just say "don't create these" and the reader just skips the insertion on the data structures. We can have an alternate constructor that receives these from the second reader and using this would said the "second pass" flag. Do you see any problem with this approach?

pablogsal:
> Even if in "second pass" mode it only needs to parse `ALLOCATION` and `ALLOCATION_WITH_NATIVE`, that means it still needs `d_stack_traces` and `d_symbol_resolver.currentSegmentGeneration()`, and getting those means it also needs `FRAME_PUSH`, `FRAME_POP`, and `MEMORY_MAP_START`. I don't think this would save as much as you think...

My guess is that the tree is the most expensive operation (from previous benchmarking).

godlygeek:
> My guess is that the tree is the most expensive operation (from previous benchmarking) so I still have hopes.

This would save the work of writing the tree, but we'd still need to read from the tree, and if I recall correctly, that was the one where we spent more time.

I have a different proposal that I think should be possible: I think I can combine `HighWatermarkFinder` and `SnapshotAllocationAggregator` into a single class that, in one pass over the records, finds the set of allocation records that represent the high water mark, the set of allocation records that represent the leaked records, and the number of bytes that were in use at the high water mark.

I think that's both less work than what you're proposing and more efficient. I'm gonna spend a bit of time experimenting on that...

godlygeek:
After some offline discussions, we realized the reverse of this proposal:

> 3\. When we do the second pass to calculate the _high watermark records_ technically we do not need to create the tables again because we already have done that in the first pass.

is much easier to implement: instead of collecting these tables in the first pass and reusing them in the second pass, we can avoid collecting them in the first pass, where we know nothing needs them, and only collect them in the second pass where we know that stack traces will be needed for each allocation.

I've pushed a commit that does that, which gives some reasonable performance gain over the initial two-pass approach.

src/memray/_memray.pyx
```diff
@@ -318,97 +321,130 @@ def start_thread_trace(frame, event, arg):
...
-        self._populate_allocations()
+        try:
+            self._file = open(file_name)
+        except OSError as exc:
+            raise OSError(f"Could not open file {file_name}: {exc.strerror}") from None
 
-    cdef void _populate_allocations(self):
-        cdef RecordReader* reader = self._get_reader()
+        self._path = "/proc/self/fd/" + str(self._file.fileno())
+        self._populate_metadata()
```
pablogsal, line 336: This is a bit misleading since we have metadata in the file and the method actually populates a ton of extra stuff. (I know, naming, amiright? 😄 )

src/memray/_memray.pyx
```diff
@@ -318,97 +321,130 @@ def start_thread_trace(frame, event, arg):
...
 
+    def __dealloc__(self):
+        self.close()
+
+    cdef object _populate_metadata(self):
+        """Initial pass to populate _header, _high_watermark, and _memory_records."""
+        cdef HighWatermarkFinder finder
```
pablogsal, line 343: Quite a big nit: Move this closer to the loop where is used

src/memray/_memray.pyx
```diff
@@ -318,97 +321,130 @@ def start_thread_trace(frame, event, arg):
...
     def __exit__(self, exc_type, exc_value, exc_traceback):
         self.close()
 
-    def __dealloc__(self):
-        self._reader.reset()
-    
-    def _yield_allocations(self, size_t index, merge_threads):
-        for elem in Py_GetSnapshotAllocationRecords(
-            self._allocations, index, merge_threads):
+    def _yield_allocations(self, size_t max_records, bool merge_threads):
```
pablogsal, line 390: Let's rename this as this now is *agregating* the allocations (so technically cannot be used also in the `get_allocation_records()`.

src/memray/_memray.pyx
```diff
@@ -318,97 +321,130 @@ def start_thread_trace(frame, event, arg):
...
+            self._high_watermark.index + 1,
+            merge_threads,
+        )
 
     def get_leaked_allocation_records(self, merge_threads=True):
-        self._ensure_reader_is_open()
-        cdef size_t snapshot_index = self._allocations.size() - 1
-        yield from self._yield_allocations(snapshot_index, merge_threads)
+        self._ensure_not_closed()
+        yield from self._yield_allocations(<size_t>(-1), merge_threads)
```
pablogsal, line 426: Please, add a comment for those that are not part of the selected cabal of "we know how unsigned overflow works" as we are :)

src/memray/_memray.pyx
```diff
@@ -318,97 +321,130 @@ def start_thread_trace(frame, event, arg):
...
-        return self._high_watermark.get()
+        reader.close()
 
     def get_high_watermark_allocation_records(self, merge_threads=True):
-        self._ensure_reader_is_open()
-        cdef HighWatermark* watermark = self._get_high_watermark()
-        yield from self._yield_allocations(watermark.index, merge_threads)
+        self._ensure_not_closed()
+        yield from self._yield_allocations(
+            self._high_watermark.index + 1,
```
pablogsal, line 420: Could you add a comment about why the `+1`?

src/memray/_memray/record_reader.cpp
```diff
@@ -115,6 +119,9 @@ RecordReader::parseFramePush()
     if (!d_input->read(reinterpret_cast<char*>(&record), sizeof(record))) {
         return false;
     }
+    if (!d_track_stacks) {
```
pablogsal, line 122: exc

src/memray/_memray/record_reader.cpp
```diff
@@ -370,6 +400,10 @@ RecordReader::nextRecord()
 PyObject*
 RecordReader::Py_GetStackFrame(unsigned int index, size_t max_stacks)
 {
+    if (!d_track_stacks) {
```
pablogsal, line 403: 👍

pablogsal (dismissed):
I left a bunch of nits, otherwise LGTM

Great job 🎉

src/memray/_memray.pyx
godlygeek: Would you prefer `numeric_limits[size_t].max()`? It's an extra `cimport` but might be more readable.

godlygeek: On reflection, I prefer that. I'll go with that.

src/memray/_memray.pyx
godlygeek: Suggestions welcome. It populates the memory records and the header, and finds the high water mark.

Or I can just inline it into the `__cinit__` and avoid giving it a name at all... any objections to that approach?

pablogsal: Inlining in the `__cinit__` is good 👍

src/memray/_memray.pyx
godlygeek: This was already aggregating the allocations, and already wasn't used in `get_allocation_records()`. I didn't change anything about that.

I'm not sure what a better name would be. I'll change it to
```py
def _yield_unfreed_allocations(self, size_t allocations_to_process, bool merge_threads):
```

I don't think that's perfect, but I'm struggling to come up with anything better. If you've got something, feel free to fixup.

godlygeek:
Nits addressed, take another look when you can.

pablogsal (approved):

PR #67, opened 2022-04-29
Improve capture file space efficiency

godlygeek:
The net effect of all of these changes is a reduction in capture file size of approximately 90%, depending on the test being measured.

### DIFF
--- a/src/memray/_memray/record_reader.cpp
+++ b/src/memray/_memray/record_reader.cpp
@@ -161,9 +161,10 @@ RecordReader::processFramePush(const FramePush& record)
 }
 
 bool
-RecordReader::parseFramePop(FramePop* record)
+RecordReader::parseFramePop(FramePop* record, unsigned int flags)
 {
-    return d_input->read(reinterpret_cast<char*>(record), sizeof(*record));
+    record->count = flags + 1;
+    return true;
 }
 
 bool
@@ -441,7 +442,7 @@ RecordReader::nextRecord()
             } break;
             case RecordType::FRAME_POP: {
                 FramePop record;
-                if (!parseFramePop(&record) || !processFramePop(record)) {
+                if (!parseFramePop(&record, record_type_and_flags.flags) || !processFramePop(record)) {
                     if (d_input->is_open()) LOG(ERROR) << "Failed to process frame pop";
                     return RecordResult::ERROR;
                 }
@@ -705,11 +706,11 @@ RecordReader::dumpAllRecords()
                 printf("FRAME_POP ");
 
                 FramePop record;
-                if (!parseFramePop(&record)) {
+                if (!parseFramePop(&record, record_type_and_flags.flags)) {
                     Py_RETURN_NONE;
                 }
 
-                printf("count=%u\n", record.count);
+                printf("count=%zd\n", record.count);
             } break;
             case RecordType::FRAME_INDEX: {
                 printf("FRAME_ID ");
--- a/src/memray/_memray/record_reader.cpp
+++ b/src/memray/_memray/record_reader.cpp
@@ -151,7 +151,7 @@ RecordReader::isOpen() const noexcept
 bool
 RecordReader::parseFramePush(FramePush* record)
 {
-    return readIntegralDelta(&d_last_python_frame_id, &record->frame_id);
+    return readIntegralDelta(&d_last.python_frame_id, &record->frame_id);
 }
 
 bool
@@ -160,7 +160,7 @@ RecordReader::processFramePush(const FramePush& record)
     if (!d_track_stacks) {
         return true;
     }
-    thread_id_t tid = d_current_thread;
+    thread_id_t tid = d_last.thread_id;
     auto [it, inserted] = d_stack_traces.emplace(tid, stack_t{});
     auto& stack = it->second;
     if (inserted) {
@@ -185,7 +185,7 @@ RecordReader::processFramePop(const FramePop& record)
     if (!d_track_stacks) {
         return true;
     }
-    thread_id_t tid = d_current_thread;
+    thread_id_t tid = d_last.thread_id;
 
     assert(!d_stack_traces[tid].empty());
     auto count = record.count;
@@ -199,10 +199,10 @@ RecordReader::processFramePop(const FramePop& record)
 bool
 RecordReader::parseFrameIndex(tracking_api::pyframe_map_val_t* pyframe_val)
 {
-    return readIntegralDelta(&d_last_python_frame_id, &pyframe_val->first)
+    return readIntegralDelta(&d_last.python_frame_id, &pyframe_val->first)
            && d_input->getline(pyframe_val->second.function_name, '\0')
            && d_input->getline(pyframe_val->second.filename, '\0')
-           && readIntegralDelta(&d_last_python_line_number, &pyframe_val->second.lineno);
+           && readIntegralDelta(&d_last.python_line_number, &pyframe_val->second.lineno);
 }
 
 bool
@@ -222,8 +222,8 @@ RecordReader::processFrameIndex(const tracking_api::pyframe_map_val_t& pyframe_v
 bool
 RecordReader::parseNativeFrameIndex(UnresolvedNativeFrame* frame)
 {
-    return readIntegralDelta(&d_last_instruction_pointer, &frame->ip)
-           && readIntegralDelta(&d_last_native_frame_id, &frame->index);
+    return readIntegralDelta(&d_last.instruction_pointer, &frame->ip)
+           && readIntegralDelta(&d_last.native_frame_id, &frame->index);
 }
 
 bool
@@ -242,7 +242,7 @@ RecordReader::parseAllocationRecord(AllocationRecord* record, unsigned int flags
 {
     record->allocator = static_cast<hooks::Allocator>(flags);
 
-    if (!readIntegralDelta(&d_last_data_pointer, &record->address)) {
+    if (!readIntegralDelta(&d_last.data_pointer, &record->address)) {
         return false;
     }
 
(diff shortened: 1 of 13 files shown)


godlygeek:
If you want it to not segfault, that costs extra.

pablogsal:
> If you want it to not segfault, that costs extra.

How much extra? I think I have some spare change... 💰

godlygeek:
Ah, dumb bug - I was failing to initialize `d_lastTid` in the record writer, and so the test suite occasionally got "lucky" enough to construct a new `RecordWriter` in the same spot as the last one, and pick up the correct thread ID for `d_lastTid` accidentally. If that happened, it would incorrectly think it had already emitted a `CONTEXT_SWITCH` record and that it didn't need to.

godlygeek:
The net effect of all of these changes is a reduction in capture file size of approximately 70%, depending on the test being measured.

src/memray/_memray/record_reader.cpp
```diff
@@ -243,22 +254,42 @@ RecordReader::parseNativeAllocationRecord()
...
+           && d_input->read(reinterpret_cast<char*>(addr), sizeof(*addr));
+}
 
+bool
+RecordReader::processSegmentHeader(const std::string& filename, size_t num_segments, uintptr_t addr)
+{
     std::vector<Segment> segments(num_segments);
     for (size_t i = 0; i < num_segments; i++) {
+        RecordType record_type;
+        if (!d_input->read(reinterpret_cast<char*>(&record_type), sizeof(record_type))
```
pablogsal, line 352: Technically this is *parsing* instead of *processing*, no? I don't think is a problem and not sure if is worth moving this into another function, but is a bit surprising given the pattern you are entablishing.

To be honest, I would do nothing here but wanted to point out in case you want to discuss :)

src/memray/_memray/record_writer.h
```diff
@@ -58,47 +87,166 @@ bool inline RecordWriter::writeString(const char* the_string)
...
+bool inline RecordWriter::writeVarint(size_t rest)
+{
+    unsigned char next_7_bits = rest & 0x7f;
+    rest >>= 7;
+    while (rest) {
+        next_7_bits |= 0x80;
+        if (!writeSimpleType(next_7_bits)) {
+            return false;
+        }
+        next_7_bits = rest & 0x7f;
```
pablogsal, line 99:
```suggestion
        next_7_bits = rest & 0x7fULL;
```

src/memray/_memray/record_writer.h
```diff
@@ -58,47 +87,166 @@ bool inline RecordWriter::writeString(const char* the_string)
...
+        rest >>= 7;
+    }
+
+    return writeSimpleType(next_7_bits);
+}
+
+bool inline RecordWriter::writeSignedVarint(ssize_t val)
+{
+    // protobuf style "zig-zag" encoding
```
pablogsal, line 108: Can you add a link for reference?

src/memray/_memray/records.h
```diff
@@ -20,7 +20,7 @@ using frame_id_t = size_t;
 using thread_id_t = unsigned long;
 using millis_t = long long;
 
-enum class RecordType {
+enum class RecordType : unsigned char {
```
pablogsal, line 23: 👀

src/memray/_memray/records.h
```diff
@@ -36,6 +36,14 @@ enum class RecordType {
     CONTEXT_SWITCH = 12,
 };
 
+struct RecordTypeAndFlags
```
pablogsal, line 39: Can we have an assert somewhere that this will be able to hold the `RecordType` enum if we add new record types in the future? This can be a nightmare if someone just adds the record that doesn't fit :(

src/memray/_memray/record_writer.h
```diff
@@ -59,6 +63,13 @@ class RecordWriter
     HeaderRecord d_header{};
     TrackerStats d_stats{};
     thread_id_t d_lastTid{};
+    uintptr_t d_lastInstructionPointer{};
```
pablogsal, line 66: A bit of a nit, but I think we should encapsulate all of these into some `state` struct for clarity.

src/memray/_memray/record_writer.h
```diff
@@ -92,6 +103,34 @@ bool inline RecordWriter::writeVarint(size_t rest)
...
+
+template<typename T>
+bool inline RecordWriter::writeIntegralDelta(T* prev, T new_val)
+{
+    ssize_t delta = new_val - *prev;
+    *prev = new_val;
+    return writeSignedVarint(delta);
+}
+
+bool inline RecordWriter::writeStringDelta(std::string* prev, const char* new_str)
```
pablogsal, line 122: I wonder if this makes a lot of sense. We don't emit segment headers and frame indexes that often and the complexity of writing these is quite big compared with varints. Did you capture by any chance stats on how efficient this is (how long the suffixes tend to be)?

src/memray/_memray/record_reader.cpp
godlygeek: Yeah, it's a bit weird, but I couldn't come up with anything nicer. The `dumpAllRecords` function wants to just parse segment records as it encounters them, and the `nextRecord()` function wants to greedily slurp up every `SEGMENT` record it encounters immediately after a `SEGMENT_HEADER`. The structure I'm moving it to in this PR is the simplest one I could come up with that retains the existing behavior for `nextRecord` (enter a mode where only `SEGMENT` records are accepted) and for `dumpAllRecords` (don't do that). If you see a nicer option, I'm all ears.

src/memray/_memray/record_reader.cpp
```diff
@@ -108,6 +108,35 @@ RecordReader::readVarint(size_t* val)
...
+    if (!readVarint(&zigzag_val)) {
+        return false;
+    }
+
+    *val = static_cast<ssize_t>((zigzag_val >> 1) ^ (~(zigzag_val & 1) + 1));
+    return true;
+}
+
+bool
+RecordReader::readStringDelta(std::string* cache, std::string* new_val)
```
pablogsal, line 124: `cache` is a bit weird. What about `last_string` or similar? Technically is a cache of 1 element but if you don't know what's going on in general it can be a bit confusing

pablogsal:
Can you add a changelog entry?

src/memray/_memray/record_reader.cpp
pablogsal: Right. The only think I can think of is to factor this part into `parseSegments()` where the caller passes the fector and passing the vector in to this function that calls `d_symbol_resolver.addSegments`

tests/integration/test_tracing.py
```diff
@@ -67,12 +67,7 @@ def test_traceback(tmpdir):
     assert len(frees) == 1
     (free,) = frees
     traceback = list(free.stack_trace())
-    assert traceback[-4:] == [
```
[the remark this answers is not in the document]
godlygeek: I'll make it raise a `NotImplementedError`, if that works for you?

src/memray/_memray/record_writer.h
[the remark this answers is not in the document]
godlygeek: Yeah, BDE style. The one piece of BDE style that I really like is that it makes is obvious at the call site whether something is an input parameter or an output parameters, and seeing the `&` there at the call site makes it clear that it's going to be modified.

src/memray/_memray/record_writer.h
```diff
@@ -58,47 +87,166 @@ bool inline RecordWriter::writeString(const char* the_string)
     return d_sink->writeAll(the_string, strlen(the_string) + 1);
 }
 
+bool inline RecordWriter::writeVarint(size_t rest)
+{
+    unsigned char next_7_bits = rest & 0x7f;
```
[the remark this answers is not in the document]
godlygeek: Discussed offline - `ULL` isn't really quite right, since that's too large for a `size_t` on 32-bit. I used `0x7fULL` in the decoder, and Pablo was proposing it here in the encoder for symmetry. Instead, I'll update the reader to do a `static_cast<size_t>` in the case where it needs to be promoted.

src/memray/_memray/records.h
godlygeek: Sure, but the `assert` will only trigger in debug builds, and the warning shows up in all builds - repeatedly, since it shows up once per file that includes `records.h`, which turns out to be most of them.

But, ok, adding the assert is easy enough.

src/memray/_memray/record_writer.h
godlygeek: I skipped it for function names based on the low utility, but it works pretty well for file names. The prefixes tend to be reasonably long - often we see long runs of `/usr/lib/python3.10/site-packages/...` where this encoding lets us avoid sending that prefix (the `site-packages` directory) over and over.

_BUT_, you're right that we don't send strings very often, so even though this does a pretty good job of compressing the strings, it makes little difference to the overall size of the file. And you're right that it more complex than the varints - I made more mistakes while writing it, and it took longer to debug.

For my test file generated with:
```
memray run --native -m test test_list test_ast
```
the total size of all filenames in the capture file (without any delta compression) is ~200kb, and the total file is about 80mb, so even if we were able to elide them entirely, it would save at most 0.2% of the total file size.

Alright, that's enough to convince me - I'm gonna drop the delta encoding of strings. I'll keep it in a branch should we decide to resurrect it, but for now, even though it works well for file names, we don't write file names often enough to justify the complexity.

src/memray/_memray/record_reader.cpp
godlygeek: I'm just gonna leave this alone. I agree that it could be nicer, but I think it's "ok" as it is, and anything else seems like it would be more complex. We can always clean this up further in another PR.

godlygeek:
I think I've addressed all the changes you've asked for except for [this one](https://github.com/bloomberg/memray/pull/67#discussion_r865867841) - I'll just leave that for later cleanup.

https://github.com/bloomberg/memray/compare/4ad3a79b2f7597b33f7e839de3f17244bc2a498f..pull/67/head should be the delta between what you last reviewed and the latest version.

pablogsal (approved):

PR #73, opened 2022-05-03
Adjusting docs flamegraph description to follow the depicted graph

ochaloup:
I'm proposing a quickfix for the documentation.

As reading the documention to memray flamegraphs
https://bloomberg.github.io/memray/flamegraph.html
I consider the information that says that the graph shows the second biggest memory consumption going from function `a()` is not correct as the `a()`[1] does not allocates the memory. The second biggest memory allocator should be `e()`.

[1]
```
def a(n):
    return [b(n), h(n)]
```


godlygeek (approved):
Looks good to me! Thank you for the contribution!

Issue #74, opened 2022-05-03
Failure building in Alpine Linux


mikets-gc:
I'm trying to install memray in Alpine based container. For example `python:3.8-alpine`.
After installing the needed packages:
` apk update & apk add gcc g++ make libunwind-dev musl-dev`
compilation errors appear. First, there is missing definition of `pvalloc` function, probably removed in gcc 10 which is available in Alpine.
I tried to add manually the function prototype in `/usr/include/malloc.h` - after that another errors appears:
```
      src/memray/_memray/hooks.h:31:30: error: redefinition of 'memray::hooks::SymbolHook<void* (*)(void*, long unsigned int, int, int, int, long int)> memray::hooks::mmap'
         31 |     FOR_EACH_HOOKED_FUNCTION(mmap64)                                                                    \
            |                              ^~~~~~
      src/memray/_memray/hooks.cpp:70:64: note: in definition of macro 'FOR_EACH_HOOKED_FUNCTION'
         70 | #define FOR_EACH_HOOKED_FUNCTION(f) SymbolHook<decltype(&::f)> f(#f, &::f);
            |                                                                ^
      src/memray/_memray/hooks.cpp:71:1: note: in expansion of macro 'MEMRAY_HOOKED_FUNCTIONS'
         71 | MEMRAY_HOOKED_FUNCTIONS
            | ^~~~~~~~~~~~~~~~~~~~~~~
      src/memray/_memray/hooks.h:30:30: note: 'memray::hooks::SymbolHook<void* (*)(void*, long unsigned int, int, int, int, long int)> memray::hooks::mmap' previously declared here
         30 |     FOR_EACH_HOOKED_FUNCTION(mmap)                                                                      \
            |                              ^~~~
      src/memray/_memray/hooks.cpp:70:64: note: in definition of macro 'FOR_EACH_HOOKED_FUNCTION'
         70 | #define FOR_EACH_HOOKED_FUNCTION(f) SymbolHook<decltype(&::f)> f(#f, &::f);
            |                                                                ^
      src/memray/_memray/hooks.cpp:71:1: note: in expansion of macro 'MEMRAY_HOOKED_FUNCTIONS'
         71 | MEMRAY_HOOKED_FUNCTIONS
            | ^~~~~~~~~~~~~~~~~~~~~~~
      In file included from src/memray/_memray/elf_utils.h:7,
                       from src/memray/_memray/hooks.h:10,
                       from src/memray/_memray/hooks.cpp:4:
[...]

PR #75, opened 2022-05-03
Add Alpine Linux support

pablogsal:
Closes: #74

### DIFF
--- a/src/memray/_memray/alloc.h
+++ b/src/memray/_memray/alloc.h
@@ -1,0 +1,12 @@
+#include <malloc.h>
+#include <stdlib.h>
+
+#ifndef __GLIBC__
+extern "C" {
+void*
+pvalloc(size_t size)
+{
+    return NULL;
+}
+}
+#endif
--- a/src/memray/_memray/elf_utils.h
+++ b/src/memray/_memray/elf_utils.h
@@ -38,18 +38,32 @@ using Sym = ElfW(Sym);
 using Sxword = ElfW(Sxword);
 using Xword = ElfW(Xword);
 
+template<class T, class U>
+T
+ensureRelocatedAddress(const Addr base, const U addr)
+{
+    // Depending on the libc version, some values may be already relocated
+    // or not. So we need to check first if the relocation already happened
+    // and make it ourselves if that is not the case.
+    auto the_addr = reinterpret_cast<Addr>(addr);
+    if (the_addr < base) {
+        return reinterpret_cast<T>(base + the_addr);
+    }
+    return reinterpret_cast<T>(the_addr);
+}
+
 template<typename T, Sxword AddrTag, Sxword SizeTag>
 struct DynamicInfoTable
 {
     T* table = nullptr;
     Xword size = {};
 
-    explicit DynamicInfoTable(const Dyn* dynamic_section)
+    explicit DynamicInfoTable(const Addr base, const Dyn* dynamic_section)
     {
         // Obtain the table address and the size from the tags in the dynamic section
         for (; dynamic_section->d_tag != DT_NULL; ++dynamic_section) {
             if (dynamic_section->d_tag == AddrTag) {
-                table = reinterpret_cast<T*>(dynamic_section->d_un.d_ptr);
+                table = ensureRelocatedAddress<T*>(base, dynamic_section->d_un.d_ptr);
             } else if (dynamic_section->d_tag == SizeTag) {
                 size = dynamic_section->d_un.d_val;
             }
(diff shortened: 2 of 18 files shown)


src/memray/_memray/tracking_api.h
```diff
@@ -23,14 +23,35 @@
 #include "record_writer.h"
 #include "records.h"
 
-#ifdef MEMRAY_TLS_MODEL
+// Only define use a different TLS model than the default if we are using glibc
+// as is the only implementation that can give us some space in the static TLS
+// array to place our TLS with confidence. In general, in other implementations
+// this may result in a relocation error at compilation time or at load time.
```
godlygeek, line 29: I _think_ `local-dynamic` should be safe even on musl.

src/memray/_memray/hooks.cpp
```diff
@@ -173,7 +175,11 @@ memalign(size_t alignment, size_t size) noexcept
 {
     assert(hooks::memalign);
 
-    void* ret = hooks::memalign(alignment, size);
+    void* ret;
+    {
+        tracking_api::RecursionGuard guard;
```
godlygeek, line 180: Is this an unrelated change? Or is this necessary because `memalign` in musl calls into a function that we've intercepted?

src/memray/_memray/alloc.h
```diff
@@ -0,0 +1,10 @@
+#include <malloc.h>
+#include <stdlib.h>
+
+#ifndef __GLIBC__
+void*
+pvalloc(size_t size)
```
godlygeek, line 6: This would break and require a change to our source if musl ever added `pvalloc`. But fixing that requires some sort of configure-like system, or feature test macros, or something... 😕

src/memray/_memray/hooks.cpp
pablogsal: Is exactly because of that. `muslc` doesn't use private symbols in this case. I didn't want to add this everywhere to not impact performance.

src/memray/_memray/alloc.h
pablogsal: If you want to handle this differently, I am open to suggestions :)

pablogsal: Technically we just need this so the test suite works, but maybe there are other better ways to do this...

godlygeek: The test suite probably shouldn't depend on a function that even the glibc docs document as "obsolete", to be fair...

We're also not intercepting `aligned_alloc` right now, which the glibc docs suggest as a replacement for the obsolete `valloc`... I opened #78 for that.

src/memray/_memray/hooks.cpp
godlygeek: Hm - would it be better to just move the `RecursionGuard` handling up from `trackAllocationImpl` and `trackDeallocationImpl` into each of the hook functions? 

It would be more consistent, at the cost of a few lines of copy pasted code.

> We only do this (for now) in this call to avoid the extra overhead of the recursion guard in the rest of the allocators (that are more commonly used by far).

> I didn't want to add this everywhere to not impact performance.

What performance are you worried about? This is only hit if the hooks are installed, and in the happy path the call to `valloc` succeeds and we call into `trackAllocation` and it sets the `RecursionGuard` itself. If we hoist the creation of the `RecursionGuard` out of `trackAllocationImpl` and `trackDeallocationImpl` and into the hook functions, then it gets used the same number of times as it used to, so it shouldn't have any performance impact at all. We're already turning it on, we just want to turn it on a bit earlier than we have been.

Or am I missing something?

src/memray/_memray/alloc.h
godlygeek: Alright, I can't immediately think of any better idea than this, but - doesn't this need an `extern "C"` on it?

src/memray/_memray/elf_utils.h
```diff
@@ -38,18 +38,32 @@ using Sym = ElfW(Sym);
...
+ensureRelocatedAddress(const Addr base, const U addr)
+{
+    // Depending on the libc version, some values may be already relocated
+    // or not. So we need to check first if the relocation already happened
+    // and make it ourselves if that is not the case.
+    auto the_addr = reinterpret_cast<Addr>(addr);
+    if (the_addr < base) {
+        return reinterpret_cast<T>(base + the_addr);
+    }
+    return reinterpret_cast<T>(the_addr);
```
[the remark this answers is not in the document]
pablogsal: Nothing as far as I know, but it is extremely unlikely. Take into account that `base` is the address of the dynamic section of the elf object and just the load point for 64 bits is already 0x40000 (and that's the entry point, you need to consider anything after it and before the dynamic section table).

src/memray/_memray/alloc.h
pablogsal: We *could* just not care about `pvalloc` to be honest....

src/memray/_memray/hooks.cpp
godlygeek: OK. I think this is a bit ugly, and I'd like to see if we can come up with something more elegant, but that can be left for another PR.

src/memray/_memray/alloc.h
godlygeek: I think it's fine, let's just add the `extern "C"`

godlygeek (approved):
LGTM - The RecursionGuard handling is a bit ugly, but we can clean it up in a follow up PR. This definitely moves the ball forward.

Issue #78, opened 2022-05-05
Hook `aligned_alloc`


godlygeek:
We're currently not hooking the C11 `aligned_alloc` function, which means we might be missing some memory allocations.

PR #79, opened 2022-05-06
Hook `aligned_alloc`

pablogsal:
Closes: #78

### DIFF
--- a/src/memray/_memray/alloc.pxd
+++ b/src/memray/_memray/alloc.pxd
@@ -3,7 +3,7 @@ cdef extern from "alloc.h" nogil:
     void free (void *ptr)
     void *malloc (size_t size)
     int posix_memalign(void** memptr, size_t alignment, size_t size)
-    void* aligned_alloc(long unsigned int alignment, long unsigned int size)
+    void* aligned_alloc(size_t alignment, size_t size)
     void *realloc (void *ptr, size_t newsize)
     void* valloc(size_t size)
     void* memalign(size_t alignment, size_t size)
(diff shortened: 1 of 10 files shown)


pablogsal:
Welp, seems that some old libc versions don't have `aligned_alloc` :(

The problem is that is not very easy to add macro guards to this:

https://github.com/bloomberg/memray/blob/e9eb395e6254222b29291078bd3b609b35d99b9d/src/memray/_memray/hooks.h#L19

pablogsal:
@godlygeek Btw, I am getting this as well:

```
        src/memray/_memray/records.h:56:30: warning: ‘memray::tracking_api::RecordTypeAndFlags::record_type’ is too small to hold all values of ‘enum class memray::tracking_api::RecordType’
             RecordType record_type : 4;
```

src/memray/_memray/alloc.pxd
```diff
@@ -3,6 +3,7 @@ cdef extern from "alloc.h" nogil:
     void free (void *ptr)
     void *malloc (size_t size)
     int posix_memalign(void** memptr, size_t alignment, size_t size)
+    void* aligned_alloc(long unsigned int alignment, long unsigned int size)
```
godlygeek, line 6:
```suggestion
    void* aligned_alloc(size_t alignment, size_t size)
```

godlygeek (approved):
I spotted one minor mistake - the `.pxd` still had this as `long unsigned int` instead of `size_t`.

I'll fix that up, but otherwise LGTM - I'll squash and land this as long as CI passes after my change.

PR #80, opened 2022-05-06
Simplify the CI steps

pablogsal:
We have currently a lot of redundancy in the CI as we are testing
multiple Python versions in 32 and 64 bits but we are also testing
independently the 64 bit version for all these Python versions.

To simplify the CI, add a specific 32 bit step using only the latest
Python version (since there should not be anything that would be
different in 32 vs 64 bit regarding Python versions) and only run the
wheel steps on releases and pushes to the main branch.


pablogsal:
Closing as seems that the Docker 32 bits github action target doesn't work as expected

PR #82, opened 2022-05-06
Leverage lz4 to add file compression on the boundaries

pablogsal:
- Vendor a copy of lz4 stream
- Update the `FileSink` and `FileSource` to use `lz4`
- Update project files to account for the dependency on `lz4`

As a bonus, this seem to make the reader 5% faster:

```
Benchmark 1: memray parse /tmp/lel.bin  > /dev/null
  Time (mean ± σ):      1.051 s ±  0.018 s    [User: 1.023 s, System: 0.026 s]
  Range (min … max):    1.016 s …  1.081 s    10 runs

Benchmark 1: memray parse /tmp/lel2.bin  > /dev/null
  Time (mean ± σ):     990.6 ms ±   9.5 ms    [User: 968.2 ms, System: 21.5 ms]
  Range (min … max):   969.1 ms … 999.5 ms    10 runs
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,8 +57,9 @@ system where you are doing the installation.
 If you wish to build Memray from source you need the following binary dependencies in your system:
 
 - libunwind
+- liblz4
 
-Check your package manager on how to install these dependencies (for example `apt-get install libunwind-dev` in Debian-based systems).
+Check your package manager on how to install these dependencies (for example `apt-get install libunwind-dev liblz4-dev` in Debian-based systems).
 
 Once you have the binary dependencies installed, you can clone the repository and follow with the normal building process:
 
--- a/src/memray/_memray/sink.cpp
+++ b/src/memray/_memray/sink.cpp
@@ -14,6 +14,7 @@
 #include <Python.h>
 
 #include "exceptions.h"
+#include "lz4_stream.h"
 #include "sink.h"
 
 namespace memray::io {
@@ -68,8 +69,10 @@ FileSink::writeAll(const char* data, size_t length)
     return true;
 }
 
-FileSink::FileSink(const std::string& file_name, bool overwrite)
-: d_fileNameStem(removeSuffix(file_name, "." + std::to_string(::getpid())))
+FileSink::FileSink(const std::string& file_name, bool overwrite, bool compress)
+: d_filename(file_name)
+, d_fileNameStem(removeSuffix(file_name, "." + std::to_string(::getpid())))
+, d_compress(compress)
 {
     int flags = O_CREAT | O_RDWR | O_TRUNC | O_CLOEXEC;
     if (!overwrite) {
@@ -167,7 +170,7 @@ std::unique_ptr<Sink>
 FileSink::cloneInChildProcess()
 {
     std::string file_name = d_fileNameStem + "." + std::to_string(::getpid());
-    return std::make_unique<FileSink>(file_name, true);
+    return std::make_unique<FileSink>(file_name, true, d_compress);
 }
 
 FileSink::~FileSink()
@@ -180,6 +183,38 @@ FileSink::~FileSink()
     }
     if (d_fd != -1) {
         ::close(d_fd);
+    }
+
+    if (d_compress) {
+        std::ifstream in_file(d_filename);
+        std::string tmp_filename = d_filename + ".lz4.tmp";
+        std::ofstream out_file(tmp_filename);
+        bool success = true;
+
+        // lz4_stream is using exceptions rather than failbit/badbit
+        try {
+            lz4_stream::ostream lz4_stream(out_file);
+            std::vector<char> buf(16 * 1024);
+            while (in_file) {
+                in_file.read(&buf[0], buf.size());
+                lz4_stream.write(&buf[0], in_file.gcount());
+            }
+        } catch (...) {
+            success = false;
+        }
+
+        out_file.close();
+        if (!in_file.eof() || !out_file) {
+            success = false;
+        }
+
+        if (!success) {
+            std::cerr << "Failed to compress input file" << std::endl;
+            ::unlink(tmp_filename.c_str());
+        } else if (0 != std::rename(tmp_filename.c_str(), d_filename.c_str())) {
+            std::perror("Error moving compressed file back to original name");
+            ::unlink(tmp_filename.c_str());
+        }
     }
 }
 
--- a/src/memray/_memray/sink.cpp
+++ b/src/memray/_memray/sink.cpp
@@ -173,6 +173,41 @@ FileSink::cloneInChildProcess()
     return std::make_unique<FileSink>(file_name, true, d_compress);
 }
 
+void
+FileSink::compress() noexcept
+{
+    std::ifstream in_file(d_filename);
+    std::string tmp_filename = d_filename + ".lz4.tmp";
+    std::ofstream out_file(tmp_filename);
+    bool success = true;
+    constexpr size_t bufsize = 4 * 1024;
+
+    // lz4_stream is using exceptions rather than failbit/badbit
+    try {
+        lz4_stream::ostream lz4_stream(out_file);
+        std::vector<char> buf(bufsize);
+        while (in_file) {
+            in_file.read(&buf[0], buf.size());
+            lz4_stream.write(&buf[0], in_file.gcount());
+        }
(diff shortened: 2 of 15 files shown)


pablogsal:
Some stats. For `memray run  -fo /tmp/lel2.bin -m test test_list test_ast test_compile test_dict test_set` it reduces the file size from **196M** to **51M** (73%).

src/memray/_memray/sink.cpp
```diff
@@ -167,7 +170,7 @@ std::unique_ptr<Sink>
 FileSink::cloneInChildProcess()
 {
     std::string file_name = d_fileNameStem + "." + std::to_string(::getpid());
-    return std::make_unique<FileSink>(file_name, true);
+    return std::make_unique<FileSink>(file_name, true, d_compress);
```
godlygeek, line 173: I'm wondering if we might want to hardcode `false` instead of passing `d_compress` here. We might get killed in the middle of compressing the file if we're a `multiprocessing` worker, and that'll leave both the compressed file and the uncompressed file on disk if so. Ignoring the setting has the better failure mode...

src/memray/_memray/sink.cpp
```diff
@@ -181,6 +184,20 @@ FileSink::~FileSink()
...
+
+    if (d_compress) {
+        std::ifstream in_file(d_filename);
+        std::string tmp_filename = d_filename + ".lz4";
+        std::ofstream out_file(tmp_filename);
+        lz4_stream::ostream lz4_stream(out_file);
+        std::copy(
+                std::istreambuf_iterator<char>(in_file),
+                std::istreambuf_iterator<char>(),
+                std::ostreambuf_iterator<char>(lz4_stream));
```
godlygeek, line 196: Surely it's inefficient to copy one character at a time...

src/memray/_memray/source.cpp
```diff
@@ -25,22 +25,32 @@ namespace memray::io {
...
-    if (!d_stream) {
+    d_raw_stream.open(d_file_name, std::ios::binary | std::ios::in);
+    if (!d_raw_stream) {
         throw IoError{"Could not open file " + file_name + ": " + std::string(strerror(errno))};
     }
+    uint32_t header;
+    d_raw_stream.read(reinterpret_cast<char*>(&header), 4);
+    d_raw_stream.seekg(0, std::ios::beg);
+
+    if (header == 0x184D2204) {
```
godlygeek, line 36: Instead of reading into a `uint32_t`, it would be better to read into a `char[4]` so that we're endianness agnostic. Or alternatively, check for `htonl(header) == 0x04224D18` I suppose.

src/memray/_memray/source.cpp
```diff
@@ -25,22 +25,32 @@ namespace memray::io {
...
+        d_stream = std::make_unique<lz4_stream::istream>(d_raw_stream);
+    } else {
+        d_stream = std::make_unique<std::ifstream>();
+        dynamic_cast<std::ifstream*>(d_stream.get())->open(d_file_name, std::ios::binary | std::ios::in);
```
godlygeek, line 40: There's a TOCTOU bug here. You're reopening the file after checking the magic, by which point it could have been rewritten with a different header.

src/memray/_memray/source.cpp
```diff
@@ -56,13 +66,13 @@ FileSource::close()
 void
 FileSource::_close()
 {
-    d_stream.close();
+    d_raw_stream.close();
```
godlygeek, line 69: Shouldn't this be `d_stream->close()`?

godlygeek: It seems strange to me that you're closing the raw stream but leaving `d_stream` alone...

README.md
```diff
@@ -57,6 +57,7 @@ system where you are doing the installation.
 If you wish to build Memray from source you need the following binary dependencies in your system:
 
 - libunwind
+- lz4
```
godlygeek, line 60: I think we should say `liblz4` for consistency.

README.md
```diff
@@ -57,6 +57,7 @@ system where you are doing the installation.
...
 
 Check your package manager on how to install these dependencies (for example `apt-get install libunwind-dev` in Debian-based systems).
```
godlygeek, line 62: We should update this example to `apt-get install libunwind-dev liblz4-dev`

src/memray/_memray/sink.cpp
pablogsal: But compression only will trigger if the tracker is destroyed which either happens at the end of the lifetime of the child or if there is a context manager somewhere. I cannot imagine a situation where we will get killed mid compression in a natural way... What do you think?

src/memray/_memray/sink.cpp
pablogsal: Haha, sure, but I will use the "this was a draft so get me out of the jail card" 😉

src/memray/_memray/source.cpp
pablogsal: AFAIK the lz4 stream doesn't have `close()` and just acts as a facade to the other stream

src/memray/_memray/sink.cpp
godlygeek: That's a fair point - the normal case for a `multiprocessing` worker will be that the Tracker destructor doesn't fire, because we never return to the frames in the parent where the context manager was entered, and so we _never_ try to compress.

Fair point.

src/memray/_memray/source.cpp
godlygeek: OK. That's annoying, but I'm gonna leave it the way you have it rather than introduce more complexity here.

godlygeek:
@pablogsal I've pushed fixup commits that get this into a state where I'm happy landing it.

Review them, see if you spot any problems I introduced, and if not, flip this to "ready for review" and I'll squash the fixups, approve, and merge.

src/memray/_memray/sink.cpp
```diff
@@ -189,13 +189,31 @@ FileSink::~FileSink()
...
         std::ofstream out_file(tmp_filename);
-        lz4_stream::ostream lz4_stream(out_file);
-        std::copy(
-                std::istreambuf_iterator<char>(in_file),
-                std::istreambuf_iterator<char>(),
-                std::ostreambuf_iterator<char>(lz4_stream));
-        if (std::rename(tmp_filename.c_str(), d_filename.c_str())) {
+        bool success = true;
+
+        // lz4_stream is using exceptions rather than failbit/badbit
```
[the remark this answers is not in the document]
pablogsal: Btw, from the notes of `std::copy`:

> In practice, implementations of std::copy avoid multiple assignments and use bulk copy functions such as [std::memmove](https://en.cppreference.com/w/cpp/string/byte/memmove) if the value type is [TriviallyCopyable](https://en.cppreference.com/w/cpp/named_req/TriviallyCopyable) and the iterator types satisfy [LegacyContiguousIterator](https://en.cppreference.com/w/cpp/named_req/ContiguousIterator).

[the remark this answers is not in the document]
godlygeek: Our value type is `TriviallyCopyable`, but our iterator type is not a `LegacyContiguousIterator`.

> A `LegacyContiguousIterator` is a `LegacyIterator` whose logically adjacent elements are also physically adjacent in memory.

That's absolutely not true for an `istream_iterator` or an `ostream_iterator`; there's (optionally) an in-memory buffer, but not all elements are adjacent in memory - not all elements are guaranteed to be in memory at all.

src/memray/_memray/sink.cpp
```diff
@@ -189,13 +189,31 @@ FileSink::~FileSink()
...
+        try {
+            lz4_stream::ostream lz4_stream(out_file);
+            std::vector<char> buf(16 * 1024);
```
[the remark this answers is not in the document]
godlygeek: Would you prefer I just use `PAGE_SIZE`? Equally arbitrary, but already defined 😄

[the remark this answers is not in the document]
godlygeek: I've just gone with `constexpr size_t bufsize = 4 * 1024;`, knocking it down to 4 kb/1 page as well.

godlygeek (approved):
I've addressed your comments on my updates - feel free to land if you're happy with the current state.

PR #83, opened 2022-05-07
Find leaks and high water mark in one pass

godlygeek:
Rather than making two passes over the file, make a single pass that
tracks both the high water mark and any allocations or deallocations
since the high water mark. From the allocations and deallocations since
the high water mark we can determine if a new high water mark has been
hit, and we can determine what was leaked at the end by aggregating
these extra allocations and deallocations on top of the high water mark
allocations.

I believe that with this scheme we need to process any given allocation
record at most twice.


godlygeek:
We've decided that the complexity of this approach outweighs its benefits.

PR #87, opened 2022-05-08
Check if the dynamic section is relocated against the ELF files

pablogsal:
Our trick to see if the dynamic section is relocated may fail in some
extreme cases if the library is loaded at a very low memory address and
the offset to the dynamic section is very very big.

To avoid the heuristic of checking against the base of the dynamic
section to decide addresses are an offset or virtual addresses, we can
check against the same value in the original ELF file (which is always
an offset) and decide then if it was or not relocated.


godlygeek:
Let's wait and see if we ever get a user report that this can fail in a real system.

PR #91, opened 2022-05-12
Use LTO and hide our symbols

godlygeek:
This seems to achieve some minor tracking speedups on the order of 2%.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -123,6 +123,8 @@ jobs:
         python3 -m pip install --upgrade pip
         python3 -m pip install -r requirements-test.txt
         python3 -m pip install -e .
+      env:
+        MEMRAY_MINIMIZE_INLINING: 1
     - name: Run Valgrind
       run: make valgrind
     - name: Run Helgrind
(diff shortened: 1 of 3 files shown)


godlygeek:
Hm, OK, prior to 3.9 `PyMODINIT_FUNC` didn't include `__attribute__ ((visibility ("default")))` - it was added in https://github.com/python/cpython/commit/0b60f64e4343913b4931dc27379d9808e5b78fe1.

That's sad.

godlygeek:
I also had to update the Valgrind suppressions to account for all the extra inlining caused by LTO - in particular, the atomic operations that Helgrind complains about are all now inlined, making them much harder to recognize.

I wonder if we should be running Valgrind and Helgrind against build that has minimal optimizations and inlining, instead...

godlygeek:
@pablogsal What's do you think about this PR?

pablogsal:
> @pablogsal What's do you think about this PR?

I measured a bigger speedup in my system (6.54%). So I think this is worth it as is basically "free". Let's do it 🚀

pablogsal (approved):

godlygeek:
I took another swing at this, and I'm happier with it - rather than updating all the Valgrind suppressions to deal with extra inlining caused by `-flto`, I updated the Valgrind/Helgrind CI action to set a flag to avoid the extra inlining. That solves the problem of needing to make our suppressions more generic than they'd otherwise need to be, and potentially ignoring real problems as a result.

PR #92, opened 2022-05-13
Allow tracing the Python allocators

pablogsal:
The Python allocators don't generally generate allocation events to the
system allocator. This is how Python normally runs and is directly
responsible of keeping it fast and this translates to profiling as well.
Sometimes, users may still be interested on the requests that the
interpreter does to the Python allocators and although this generates a
lot more data and makes profiling slower, it may help in some
situations. This acts also as an alternative way to run with
`PYTHONMALLOC=malloc` but without collating the data with the system
allocator.

### DIFF
--- a/src/memray/commands/run.py
+++ b/src/memray/commands/run.py
@@ -33,6 +33,7 @@ def _run_tracker(
     args: argparse.Namespace,
     post_run_message: Optional[str] = None,
     follow_fork: bool = False,
+    trace_python_allocators: bool = False,
 ) -> None:
     sys.argv = [args.script, *args.script_args]
     if args.run_as_module:
@@ -41,6 +42,8 @@ def _run_tracker(
         kwargs = {}
         if follow_fork:
             kwargs["follow_fork"] = True
+        if trace_python_allocators:
+            kwargs["trace_python_allocators"] = True
         tracker = Tracker(destination=destination, native_traces=args.native, **kwargs)
     except OSError as error:
         raise MemrayCommandError(str(error), exit_code=1)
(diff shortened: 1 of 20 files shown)


godlygeek:
This fails catastrophically when combined with `--follow-fork`. Given:
```shell-session
$ cat multiprocessing_test.py
import mmap
import multiprocessing


def worker(n):
    print(n, mmap.mmap(-1, 1))


with multiprocessing.Pool(5) as pool:
    pool.map(worker, range(5))
```
trying to run:
```shell-session
memray run --trace-python-allocators --follow-fork multiprocessing_test.py
```
loops forever creating an infinite number of 4 KB files, thanks to the child processes failing to start properly in the child at-fork handler.

godlygeek:
That seems to be easily fixed by having `registerPymallocHooks()` check whether the hooks are already registered, in order to avoid mistaking our own hooks for the original allocators.

src/memray/commands/run.py
```diff
@@ -215,6 +219,12 @@ def prepare_parser(self, parser: argparse.ArgumentParser) -> None:
             help="Record allocations in child processes forked from the tracked script",
             default=False,
         )
+        parser.add_argument(
+            "--trace-python-allocators",
```
godlygeek, line 223: I wonder if we should shorten this to `--trace-pymalloc` or even just `--pymalloc`? `--trace-python-allocators` is a lot to remember/type.

I almost wonder if we should just call it `--leaks`, since the time when it's most useful is in conjunction with `memray flamegraph --leaks`, but that's probably too reductive...

I think `--trace-pymalloc` might be the best possible name, on balance.

pablogsal: Technically `pymalloc` != Python allocators. You could track the python allocators where they in turn do not call pymaalloc (they could call something else like malloc).

godlygeek: I see your point, but worry that no one will use the option if we call it `--trace-python-allocators` 😄

godlygeek (approved):
LGTM after my tweaks, and Pablo says he's happy with those tweaks.

PR #94, opened 2022-05-14
Add the option to do statistical sampling of allocations

pablogsal:
Look ma, no hands!

⚠️ For evaluation purposes only ... for now ⚠️


src/memray/_memray/tracking_api.cpp
```diff
@@ -290,6 +328,7 @@ Tracker::Tracker(
 
     d_background_thread = std::make_unique<BackgroundThread>(d_writer, memory_interval);
     d_background_thread->start();
+    d_sampler.setSamplingInterval(sampling_interval);
```
godlygeek, line 331: We're only calling this in the thread where the `Tracker` is created. Unless I'm missing something, other threads are winding up with 0 for all the member variables of `d_sampler`.

I think what we could do is call `setSamplingInterval` unconditionally in `trackAllocationImpl` and `trackDeallocationImpl`, and update `setSamplingInterval` to short-circuit in the case where the old sampling interval and the new one are the same.

src/memray/_memray/tracking_api.cpp
```diff
@@ -441,16 +480,43 @@ Tracker::childFork()
...
             old_tracker->d_follow_fork,
-            old_tracker->d_trace_python_allocators));
+            old_tracker->d_trace_python_allocators,
+            old_tracker->d_sampling_interval));
     RecursionGuard::isActive = false;
 }
 
+// static std::unordered_set<void*> pointers;
+
+static constexpr char TRACE_MARK = 42;
```
godlygeek, line 498: I'd suggest that a better mark is a pointer to itself. That is:
```c
*(void*)(addr) = (void*)addr;
```

I think that's _extremely_ unlikely to occur by chance. It does require us to think about alignment, though.

src/memray/_memray/tracking_api.cpp
```diff
@@ -441,16 +480,43 @@ Tracker::childFork()
...
+
+static inline void
+markPtr(void* ptr, const hooks::Allocator& func)
+{
+    if (func == hooks::Allocator::MMAP || isPymalloc(func)) {
+        return;
+    }
+    size_t usable_size = malloc_usable_size(ptr);
```
godlygeek, line 506: In order for this to be safe, we need to start hooking `malloc_usable_size` and subtracting out our own overhead in the case where the mark is set, so that the user doesn't think it's safe to write into our extra buffer. Something like:
```cpp
size_t
malloc_usable_size(void* addr) noexcept
{
    assert(hooks::malloc_usable_size);
    size_t ret = hooks::malloc_usable_size(addr);
    if (ret > 0 && isPtrMarked(addr)) {
        ret -= MEMRAY_ALLOC_OVERHEAD;
    }
    return ret;
}
```
though that glosses over the fact that this stuff is spread across different translation units...

src/memray/_memray/tracking_api.cpp
```diff
@@ -248,18 +251,53 @@ std::atomic<bool> Tracker::d_active = false;
 std::unique_ptr<Tracker> Tracker::d_instance_owner;
 std::atomic<Tracker*> Tracker::d_instance = nullptr;
 MEMRAY_FAST_TLS thread_local size_t NativeTrace::MAX_SIZE{64};
+thread_local BernoulliSampler Tracker::d_sampler{};
```
godlygeek, line 254: Hm. There's only two piece of mutable state inside the `BernoulliSampler` - the `random_engine`, and `d_bytes_until_next_sample`. It looks like we can easily make the `random_engine` thread-local, since (at least on glibc) it has a trivial destructor, so we don't need to worry about using it while a thread is being torn down. If we make the `random_engine` thread-local, I think we can use a `std::atomic` for `d_bytes_until_next_sample`, to allow sharing a single sampler across all threads without needing any mutexes.

This implementation seems to pass the tests, after making `d_sampler` a regular instance variable instead of a `static thread_local` - can you poke any holes in this?

```cpp
class BernoulliSampler
{
  public:
    // Methods
    void setSamplingInterval(size_t sampling_interval);
    size_t calculateSampleSize(size_t size);

  private:
    // Methods
    size_t poissonStep();

    // Data members
    size_t d_sampling_interval_in_bytes;
    double d_sampling_probability;
    std::atomic<size_t> d_bytes_until_next_sample;
};

void
BernoulliSampler::setSamplingInterval(size_t sampling_interval)
{
    d_sampling_interval_in_bytes = sampling_interval;
    d_sampling_probability = 1.0 / d_sampling_interval_in_bytes;
    d_bytes_until_next_sample = poissonStep();
}

size_t
BernoulliSampler::poissonStep()
{
    static_assert(std::is_trivially_destructible<std::default_random_engine>::value);
    static thread_local std::default_random_engine s_random_engine;
    std::exponential_distribution<double> exponential_dist(d_sampling_probability);
    return static_cast<size_t>(exponential_dist(s_random_engine)) + 1;
}

size_t
BernoulliSampler::calculateSampleSize(size_t size)
{
    if (size >= d_sampling_interval_in_bytes) {
        return size;
    }

    size_t n_samples = 0;
    size_t next_step = 0;
    while (size) {
        size_t loaded_bytes_until_next_sample = d_bytes_until_next_sample.load();
        if (size >= loaded_bytes_until_next_sample) {
            // We're being sampled! We need to replace our atomic entirely.
            // Randomly decide when to stop next, if we haven't already.
            if (!next_step) {
                next_step = poissonStep();
            }

            if (!d_bytes_until_next_sample.compare_exchange_weak(
                        loaded_bytes_until_next_sample,
                        next_step)) {
                continue;  // Didn't update the atomic; loop and try again.
            }

            // Successfully updated the atomic; some bytes are accounted for.
            next_step = 0;  // Our random number has been used
            size -= loaded_bytes_until_next_sample;
            n_samples++;
        } else {
            // We're not being sampled; just decrement our atomic.
            if (!d_bytes_until_next_sample.compare_exchange_weak(
                        loaded_bytes_until_next_sample,
                        loaded_bytes_until_next_sample - size))
            {
                continue;  // Didn't update the atomic; loop and try again.
            }

            // Successfully updated the atomic; all bytes are accounted for.
            size = 0;
        }
    }

    return d_sampling_interval_in_bytes * n_samples;
}
```

godlygeek: I think it would be possible to use `memory_order_acq_rel` for the read/modify/write, and `memory_order_relaxed` for the loads - but I don't claim to understand the C++ memory model terribly well, and that may be premature optimization.

godlygeek: In fact, I _think_ it would be correct to use `memory_order_relaxed` for everything, since we're only looking for atomicity and don't care at all about any writes any other thread may have done to any address besides our atomic.

docs/sampling.rst
```diff
@@ -0,0 +1,133 @@
...
+accuracy of the results. For example, choosing a very small value for ``p``
+(which means we will wait a lot for taking samples) will give us a lot of speed
+(because we won't be sampling many allocations) but will cost us a lot in
+accuracy. On the other hand, setting ``p=1`` will give us perfect accuracy, as
+we will be sampling every single byte. We need to understand how ``p`` impacts
+our accuracy so we can choose the right value for every situation.
+
+In the following figure, you can see the probability of taking a sample
+for different sampling sizes, depending on the allocation size.  We can
+see from the chart below that if we 16X our sampling rate from 32B to 512B
```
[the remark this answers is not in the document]
pablogsal: `sample rate` = 1/`sample size`. The figure is increasing the sample size and therefore decreasing the rate.

docs/sampling.rst
```diff
@@ -0,0 +1,133 @@
...
+  this becuse the probability tht we are going to take at least one sample is very high and therefore is a good performance
+  optimization.
+
+* If the allocation size smaller than our sampling size, then we substract the size of the allocation from the ammount of bytes
+  we need to wait until we take a sample. If this number reaches zero, we take a sample and we draw another random number from
+  the exponential distribution. The amount of times we had to draw from the exponential distribution is the number of samples
+  the allocation should count as.
+
+* We multiply the number of samples we drew within the allocation by the sampling rate to get an estimate of the statistical
+  size of the allocation. We record this allocation and its stack as if the allocation had the statistical size that we obtained.
```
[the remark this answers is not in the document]
pablogsal: Not sure I completely follow, but if you are proposing with "ust track the exact number of bytes allocated since our last sample, and weight the new sample as that size" to just take a sample every nth byte, then the problem is that you are biasing towards the allocation pattern of the application. If the application has cyclic/periodic/regular allocation patterns there is a lot of them that you will never see, that's why you need an unbiased statistical approach.

docs/sampling.rst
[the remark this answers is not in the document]
godlygeek: Right, the figure shows increasing the sample size, but this sentence says "if we 16X our sampling rate" (which is 1/sample size), and then says "from 32B to 512B" (which reflects an increase of sample size).

Shouldn't the sentence say "if we raise our sample size from 32B to 512B (a factor of 16)" or something like that?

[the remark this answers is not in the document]
pablogsal: Ah I see what you mean. Yeah, apologies, you are absolutely right, this needs correcting.

docs/sampling.rst
```diff
@@ -0,0 +1,133 @@
...
+* If the allocation size is bigger than the sampling size, we sample the entire allocation with its original size. We do
+  this becuse the probability tht we are going to take at least one sample is very high and therefore is a good performance
+  optimization.
+
+* If the allocation size smaller than our sampling size, then we substract the size of the allocation from the ammount of bytes
+  we need to wait until we take a sample. If this number reaches zero, we take a sample and we draw another random number from
+  the exponential distribution. The amount of times we had to draw from the exponential distribution is the number of samples
+  the allocation should count as.
+
+* We multiply the number of samples we drew within the allocation by the sampling rate to get an estimate of the statistical
```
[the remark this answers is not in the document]
pablogsal: As we treat this with their exact size they just count as one sample. The key is that statistically we know that they are going to be sampled at least one so we just transform the allocation into a single sample of the true size. Doing this is still statistically significant and keeps the sampling unbiased.

docs/sampling.rst
```diff
@@ -0,0 +1,132 @@
...
+The exponential distribution tells us the amount of bytes we need to wait before
+taking the next sample. This is much better than throwing a random
+number per allocation, because this way we only roll one random number per
+sample, then keep substracting bytes from this number until it reaches zero - at
+which point we record a new sample, roll a new random number, and start over.
+This minimizes the amount of random numbers we need to generate.
+
+As is customary in this method, an optimization here is that if a given
+allocation is bigger than the average ammount of bytes we need to wait until the
+next sample, we take it directly as is very likely that we are going to sample
```
godlygeek, line 46:
```suggestion
next sample, we take it directly as it's very likely that we are going to sample
```

src/memray/_memray/tracking_api.cpp
godlygeek: Let's stick with just the 1 byte of overhead for now, since it's simpler and means we don't need to worry about alignment. We can do this in a follow up PR if we want to.

pablogsal:
Seems that we have a segfault in alpine:

```
  tests/integration/test_tracking.py::test_simple_allocation_tracking_sampling[realloc-AllocatorType.REALLOC] Fatal Python error: Segmentation fault
```

pablogsal:
Closing this in favour of #277

Issue #96, opened 2022-05-14
Aggregations in the high watermark are not taking into account the native frame


pablogsal:
We are collapsing together allocations with the same python stack trace but different native stack trace. This produces flame graphs correct in the Python part but incorrectly collapsed in the native part. The problem is here:

https://github.com/bloomberg/memray/blob/0c3874a95b1ea26868f3fa344595dfb78fd6e8fa/src/memray/_memray/snapshot.cpp#L88-L92

The map from location to allocation only takes into account the `frame_index` but not the `native_frame_id`.

PR #97, opened 2022-05-15
Do not aggregate native allocations located in different native lines

pablogsal:
Closes: #96

When we are calculating the high watermark allocation records, we
aggregate together allocations that happen in the same location.
Unfortunately, currently, we define the "location" only by taking into
account the Python stack. This means that two allocations that happen
under the same Python stack but with different native stacks underneath
will be incorrectly aggregated together leading to incorrect final
records and incorrect reports.

We must take into account both the C and the Python stack in the data
structures that are used in the snapshot to be able to produce correct
aggregated allocations.


godlygeek (approved):
I made some minor edits, after which this LGTM

pablogsal:
> I made some minor edits, after which this LGTM

Thanks! I am pushing a small fix for the pyi file

godlygeek (approved):
LGTM

PR #111, opened 2022-05-17
Add a progress bar indicator for record processing

pablogsal:
As record processing is an operation that can take an arbitrary amount
of time depending on the size of the file, some users can be frustrated
because they don't know how much time the processing is going to take.

To help with this, add a progress bar indicator that periodically
updates while we are processing records in the `FileReader`. The class is
updated strategically to avoid impacting the performance of the actual
processing.


gaborbernat (dismissed):
Needs a news entry 👍

src/memray/_memray/record_reader.h
```diff
@@ -83,6 +85,7 @@ class RecordReader
     std::unordered_map<thread_id_t, std::string> d_thread_names;
     Allocation d_latest_allocation;
     MemoryRecord d_latest_memory_record;
+    std::atomic<size_t> d_processed_allocations;
```
pablogsal, line 88:
```suggestion
    std::atomic<size_t> d_processed_allocations{0};
```

godlygeek:
I played with this for a while locally. I've pushed some fixup commits that manage to make it much faster, especially in the incomplete file / OOM killer case. Rather than letting Rich decide when to redraw the screen, my version calculates that ourselves, using calls to `clock_gettime` (which is in the VDSO) rather than making calls into Python code. This winds up being substantially faster, especially when a too-low value was chosen for `_update_interval`.

Incidentally, I also raised `_update_interval` from 200 to 100,000. On my WSL box with its relatively slow disk IO, we still process approximately 7,000,000 allocations per second, so at 10 refreshes per second we process ~700k allocations per refresh, which works out to 7 calls to `clock_gettime` per refresh, which seems like a fair enough price to pay for snappier updates of the progress bar.

src/memray/_memray.pyx
```diff
@@ -335,21 +341,106 @@ def start_thread_trace(frame, event, arg):
...
+            self._next_refresh = now
+            self._next_refresh.tv_nsec += self._ns_between_refreshes
+            while self._next_refresh.tv_nsec > 1_000_000_000:
+                self._next_refresh.tv_nsec -= 1_000_000_000
+                self._next_refresh.tv_sec += 1
+            return True
+
+        return False
+
+    cdef void update(self, size_t n_processed) except*:
```
godlygeek, line 414: Why make this `cdef void` with an `except*` instead of just making it
```suggestion
    cdef update(self, size_t n_processed):
```
? I'd expect `except *` to be slightly slower, since that requires an extra call to `PyErr_Occurred()` per call, instead of just being able to compare a pointer against `NULL`. Some rough profiling seems to bear out that it's very slightly slower...

pablogsal: I added that as a temporary fix to the exceptions being raised and I forgot to clean it up. Good catch!

godlygeek (approved):
LGTM

PR #125, opened 2022-05-26
Update stats in capture files as records are written

godlygeek:
This will allow us to have an accurate count of records even in the case where a process is killed early by a signal.


godlygeek:
Closing in favor of #128

Issue #126, opened 2022-05-27
Some tests for native tracking don't work in aarch64


pablogsal:
Seems that https://github.com/bloomberg/memray/pull/52 has broken the tests for native tracking on aarch64:

```
   def test_simple_call_chain_with_native_tracking(tmpdir, monkeypatch):
        # GIVEN
        output = Path(tmpdir) / "test.bin"
        extension_name = "multithreaded_extension"
        extension_path = tmpdir / extension_name
        shutil.copytree(TEST_NATIVE_EXTENSION, extension_path)
        subprocess.run(
            [sys.executable, str(extension_path / "setup.py"), "build_ext", "--inplace"],
            check=True,
            cwd=extension_path,
            capture_output=True,
        )

        # WHEN
        with monkeypatch.context() as ctx:
            ctx.setattr(sys, "path", [*sys.path, str(extension_path)])
            from native_ext import run_simple  # type: ignore

            with Tracker(output, native_traces=True):
                run_simple()

        # THEN
        records = list(FileReader(output).get_allocation_records())
        vallocs = [
            record
            for record in filter_relevant_allocations(records)
            if record.allocator == AllocatorType.VALLOC
        ]

        assert len(vallocs) == 1
        (valloc,) = vallocs

        (python_stack_trace,) = valloc.stack_trace()
        func, filename, line = python_stack_trace
        assert func == "test_simple_call_chain_with_native_tracking"
        assert filename.endswith(__file__)

        expected_symbols = ["baz", "bar", "foo"]
>       assert expected_symbols == [stack[0] for stack in valloc.native_stack_trace()[:3]]
E       AssertionError: assert ['baz', 'bar', 'foo'] == []
E         Left contains 3 more items, first extra item: 'baz'
E         Full diff:
E         - []
E         + ['baz', 'bar', 'foo']

tests/integration/test_native_tracking.py:112: AssertionError
```

Bisecting shows b6d865b7f481c4c67f6268483bc81f11a0833527 as the first bad commit:

```
b6d865b7f481c4c67f6268483bc81f11a0833527 is the first bad commit
commit b6d865b7f481c4c67f6268483bc81f11a0833527
Author: Matthias Diener <mdiener@illinois.edu>
Date:   Fri Apr 22 10:36:02 2022 -0700

    Avoid inlining `exact_unwind`

    On some architectures this can't be inlined. We don't actually care
    whether or not it is inlined, only that whether it is or isn't inlined
    is consistent regardless of build flags.

[...]

PR #52, opened 2022-04-22
Avoid inlining `exact_unwind`

matthiasdiener:
*Issue number of the reported bug or feature request: #51*

**Describe your changes**
Remove inline attribute from exact_unwind()

**Testing performed**
Compiled and ran a simple test on a ppc64le machine.

PR #127, opened 2022-05-27
Revert "Avoid inlining `exact_unwind`" to fix unwinding in aarch64

pablogsal:
Reverts bloomberg/memray#52

Seems that libunwind just chokes if the function that calls `unw_get_reg` is not inlined in aarch64. The original change was made as a best-effort to better support powerpc but given that powerpc doesn't even fully work but aarch64 does, we prefer to support aarch64 here.

Closes: #126

### DIFF
--- a/src/memray/_memray/tracking_api.h
+++ b/src/memray/_memray/tracking_api.h
@@ -110,7 +110,6 @@ class NativeTrace
         if (size == MAX_SIZE) {
             d_data.resize(0);
             size = exact_unwind();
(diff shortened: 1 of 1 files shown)


godlygeek:
Let's chat about this tomorrow. I'm OK with reverting it, though I wonder if there's a solution that would work for both aarch64 and ppc64le.

pablogsal:
Closing this in favour of https://github.com/bloomberg/memray/pull/141

PR #129, opened 2022-06-01
Recognize and handle truncated capture files

godlygeek:
Make the `RecordWriter` finish a capture file by writing a non-zero
trailer byte that the `RecordReader` will ignore. When the `FileSource`
reads a (non-compressed) capture file, have it ignore any trailing zero
bytes without returning them to the caller. If the caller attempts to
read into them, treat it as though EOF has been reached instead.

When a trailer byte is encountered by the reader, treat it as EOF,
rather than continuing to request bytes from the reader.

### DIFF
--- a/src/memray/_memray/record_reader.cpp
+++ b/src/memray/_memray/record_reader.cpp
@@ -430,8 +430,16 @@ RecordReader::nextRecord()
         }
 
         switch (record_type_and_flags.record_type) {
-            case RecordType::UNINITIALIZED: {
-                // Skip it. All remaining bytes should be 0.
+            case RecordType::OTHER: {
+                switch (static_cast<OtherRecordType>(record_type_and_flags.flags)) {
+                    case OtherRecordType::TRAILER: {
+                        return RecordResult::END_OF_FILE;
+                    } break;
+                    default: {
+                        if (d_input->is_open()) LOG(ERROR) << "Invalid record subtype";
+                        return RecordResult::ERROR;
+                    } break;
+                }
             } break;
             case RecordType::ALLOCATION: {
                 AllocationRecord record;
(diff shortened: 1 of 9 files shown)


src/memray/_memray/source.cpp
```diff
@@ -63,6 +75,21 @@ FileSource::close()
     _close();
 }
 
+void
+FileSource::findReadableSize()
+{
```
pablogsal, line 80: Can you add a comment here explaining what is the "readable size" for quick future reference?

src/memray/_memray/record_reader.cpp
```diff
@@ -430,8 +430,16 @@ RecordReader::nextRecord()
         }
 
         switch (record_type_and_flags.record_type) {
-            case RecordType::UNINITIALIZED: {
-                // Skip it. All remaining bytes should be 0.
+            case RecordType::OTHER: {
+                switch (static_cast<OtherRecordType>(record_type_and_flags.flags)) {
```
pablogsal, line 434: I know you don't want to, but I would pull this into another function to reduce the clutter, but I am fine either way. I was wanted to take the opportunity to complain

src/memray/_memray/source.cpp
```diff
@@ -63,6 +75,21 @@ FileSource::close()
...
+    d_raw_stream->seekg(-1, d_raw_stream->end);
+    while (*d_raw_stream) {
```
pablogsal, line 82: I know it can't happen, but we may want to check also that we don't go past the beginning of the file.

pablogsal (approved):
LGTM

Great work. I just left a collection of nits and other stuff, but just feel free to land when you are happy, don't wait for me to check back in :)

src/memray/_memray/source.cpp
godlygeek: Good idea, will do.

src/memray/_memray/source.cpp
godlygeek: I just checked the behavior - if we try, `*d_raw_stream` will be false, because `failbit` gets set, so we break out of the loop without changing `d_readable_size` from its default value of `0` - so all future `read()` calls will fail.

And, the only way this can happen is if someone overwrites the file with all zeroes while we were in the middle of reading it, since we've already read some non-zero bytes by this point. I don't think this is worth worrying about.

godlygeek: I'll add a comment explaining what will happen, though:
```c
        // If we're at BOF, this sets failbit and makes the loop break.
        d_raw_stream->seekg(-1, d_raw_stream->cur);
```

src/memray/_memray/record_reader.cpp
godlygeek: I don't want to - at least, not yet. It's a bit cluttered, but this way all of the record types are handled in one function, which I think is a feature. I don't want to handle the "overflow" ones specially just because they didn't fit in 4 bits. They're regular record types, except that we chose to encode them in a weird way.

pablogsal (approved):

Issue #85, opened 2022-05-07
Information about which allocator was used is lost in roll-up


godlygeek:
Consider this code:

```py
import mmap

for i in range(5):
    x = mmap.mmap(-1, 10000000000) if i == 0 else "x" * 1000000
```

This is what we currently show in the `stats` reporter:

```
📏 Total allocations:
        21

📦 Total memory allocated:
        9.314GB

📊 Histogram of allocation size:
        min: 528.000B
        ----------------------------------------
        < 2.754KB  : 8 ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇
        < 14.717KB : 2 ▇▇▇▇▇▇▇
        < 78.624KB : 0
        < 420.047KB: 0
        < 2.191MB  : 0
        < 11.708MB : 0
        < 62.549MB : 0
        < 334.166MB: 0
        < 1.743GB  : 0
        <=9.314GB  : 1 ▇▇▇▇
        ----------------------------------------
        max: 9.314GB

📂 Allocator type distribution:
         MALLOC: 11
```

The allocator type distribution is incorrect - it says there were no MMAP allocations and only MALLOC allocations. The fact that different types of allocations occurred on the same line is being lost.

godlygeek:
Also - the fact that "📂 Allocator type distribution:" shows 11 allocations and "📏 Total allocations:" shows 21 seems suspicious... is free() being counted as an allocation?

pablogsal:
> Also - the fact that "📂 Allocator type distribution:" shows 11 allocations and "📏 Total allocations:" shows 21 seems suspicious... is free() being counted as an allocation?

No, it is taking into account how many allocations have been aggregated into the same place IIRC.

pablogsal:
> The fact that different types of allocations occurred on the same line is being lost.

That happens because we aggregate all allocations that happen in the same place together and indeed if multiple allocators are used this information is lost.

PR #136, opened 2022-06-06
Fix many bugs in the stats reporter

godlygeek:
Remove the `--include-all-allocations` / `-a` option for the `stats`
reporter and have it process all allocations by default. Its default
behavior of only processing allocations that made up the high water mark
of the application's memory usage was confusing and misleading.

Previously when `-a` was provided, deallocations with `free` and
`munmap` were included in the reported count of "Total allocations".
This has been fixed.

Similarly, when `-a` was provided, memory freed by `munmap` (though not
`free`) was incorrectly being added to the count of total memory
allocated. This has been fixed.

The histogram has been fixed to work on the actual sizes of all
allocations. Previously it only saw the sizes after a rollup by stack
had already been performed, and it was therefore binning allocation
sizes that had already been summed.

Fix the two "largest allocating locations" sections to actually
aggregate by location. Previously they were aggregating by distinct
stack, so if the same line of code was reached by two different paths,
it would be counted separately instead of together.

Exclude `PYMALLOC_FREE` from the allocator type distribution (other
deallocators were already being ignored, but this recently added one was
missed).

Additionally, this fixes the bug introduced in #134 where `memray stats -a` was incorrectly handling any pointer that was reused after being reallocated. I haven't included a changelog entry for that one since it was never included in any released version.

Closes: #85 
Closes: #134

### DIFF
--- a/news/136.bugfix.3.rst
+++ b/news/136.bugfix.3.rst
@@ -1,0 +1,1 @@
+Fix the ``memray stats`` histogram to be based on the actual sizes of all allocations. Previously it only saw the sizes after a rollup by stack had already been performed, so it was binning allocation sizes that had already been summed.
--- a/news/136.bugfix.4.rst
+++ b/news/136.bugfix.4.rst
@@ -1,0 +1,1 @@
+Fix the two "largest allocating locations" sections in the ``memray stats`` report to actually aggregate by location. Previously they were aggregating by distinct stacks, so if two different paths hit the same line of code, it would be counted separately instead of together.
--- a/src/memray/_memray.pyx
+++ b/src/memray/_memray.pyx
@@ -1,5 +1,6 @@
 import collections
 import contextlib
+import heapq
 import os
 import pathlib
 import sys
@@ -16,6 +17,7 @@ from posix.time cimport CLOCK_MONOTONIC
 from posix.time cimport clock_gettime
 from posix.time cimport timespec
 
+from _memray.hooks cimport Allocator
 from _memray.hooks cimport isDeallocator
 from _memray.logging cimport setLogThreshold
 from _memray.record_reader cimport RecordReader
@@ -28,6 +30,7 @@ from _memray.sink cimport FileSink
 from _memray.sink cimport NullSink
 from _memray.sink cimport Sink
 from _memray.sink cimport SocketSink
+from _memray.snapshot cimport AllocationStatsAggregator
 from _memray.snapshot cimport HighWatermark
 from _memray.snapshot cimport HighWatermarkFinder
 from _memray.snapshot cimport Py_GetSnapshotAllocationRecords
@@ -39,6 +42,7 @@ from _memray.source cimport SocketSource
 from _memray.tracking_api cimport Tracker as NativeTracker
 from _memray.tracking_api cimport install_trace_function
 from cpython cimport PyErr_CheckSignals
+from libc.stdint cimport uint64_t
 from libcpp cimport bool
 from libcpp.limits cimport numeric_limits
 from libcpp.memory cimport make_shared
@@ -46,12 +50,14 @@ from libcpp.memory cimport make_unique
 from libcpp.memory cimport shared_ptr
 from libcpp.memory cimport unique_ptr
 from libcpp.string cimport string as cppstring
+from libcpp.unordered_map cimport unordered_map
 from libcpp.utility cimport move
 from libcpp.vector cimport vector
 
 from ._destination import FileDestination
 from ._destination import SocketDestination
 from ._metadata import Metadata
+from ._stats import Stats
 
 include "_memray_test_utils.pyx"
 
@@ -344,6 +350,32 @@ def start_thread_trace(frame, event, arg):
     return start_thread_trace
 
 
+cdef millis_to_dt(millis):
+    return datetime.fromtimestamp(millis // 1000).replace(
+        microsecond=millis % 1000 * 1000)
+
+
+cdef _create_metadata(header, peak_memory):
+    stats = header["stats"]
+    allocator_id_to_name = {
+        PythonAllocatorType.PYTHON_ALLOCATOR_PYMALLOC: "pymalloc",
+        PythonAllocatorType.PYTHON_ALLOCATOR_PYMALLOC_DEBUG: "pymalloc debug",
+        PythonAllocatorType.PYTHON_ALLOCATOR_MALLOC: "malloc",
+        PythonAllocatorType.PYTHON_ALLOCATOR_OTHER: "unknown",
+    }
+    return Metadata(
+        start_time=millis_to_dt(stats["start_time"]),
+        end_time=millis_to_dt(stats["end_time"]),
+        total_allocations=stats["n_allocations"],
+        total_frames=stats["n_frames"],
+        peak_memory=peak_memory,
+        command_line=header["command_line"],
+        pid=header["pid"],
+        python_allocator=allocator_id_to_name[header["python_allocator"]],
+        has_native_traces=header["native_traces"],
+    )
+
+
 cdef class ProgressIndicator:
     cdef bool _report_progress
     cdef object _indicator
@@ -519,12 +551,7 @@ cdef class FileReader:
     def __exit__(self, exc_type, exc_value, exc_traceback):
         self.close()
 
-    def _aggregate_allocations(
-        self,
-        size_t records_to_process,
-        bool merge_threads,
-        bool ignore_frees,
-    ):
+    def _aggregate_allocations(self, size_t records_to_process, bool merge_threads):
         cdef SnapshotAllocationAggregator aggregator
         cdef shared_ptr[RecordReader] reader_sp = make_shared[RecordReader](
             unique_ptr[FileSource](new FileSource(self._path))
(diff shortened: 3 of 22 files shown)


news/136.bugfix.3.rst
```diff
@@ -0,0 +1 @@
+Fix the ``memray stats`` histogram to be based on the actual sizes of all allocations. Previously it only saw the sizes after a rollup by stack had already been performed, and it was therefore binning allocation sizes that had already been summed.
```
pablogsal, line 1:
```suggestion
Fix the ``memray stats`` histogram to be based on the actual sizes of all allocations. Previously it only saw the sizes after a rollup by stack had already been performed, and it was, therefore, binning allocation sizes that had already been summed.
```

news/136.bugfix.4.rst
```diff
@@ -0,0 +1 @@
+Fix the two "largest allocating locations" sections in the ``memray stats`` report to actually aggregate by location. Previously they were aggregating by distinct stack, so if the same line of code was reached by two different paths, it would be counted separately instead of together.
```
pablogsal, line 1:
```suggestion
Fix the two "largest allocating locations" sections in the ``memray stats`` report to actually aggregate by location. Previously they were aggregating by distinct stacks, so if two different paths reached the same line of code, it would be counted separately instead of together.
```

src/memray/reporters/stats.py
```diff
@@ -202,21 +131,25 @@ def render(
...
+            yield f"{self._format_location(location)} -> {size_fmt(size)}"
+
+    def _get_top_allocations_by_count(self) -> Iterator[str]:
+        for location, count in self._stats.top_locations_by_count:
+            yield f"{self._format_location(location)} -> {count}"
+
+    def _get_allocator_type_distribution(self) -> Iterator[str]:
+        for allocator_name, count in sorted(
+            self._stats.allocation_count_by_allocator.items(),
+            key=lambda item: item[1],
```
pablogsal, line 152: Nit: you can use `operator.itemgetter` for better performance

src/memray/reporters/stats.py
```diff
@@ -1,98 +1,41 @@
...
 
     # Determine the upper bound in bytes for each bin
     steps = [int(math.exp(low + step * (i + 1))) for i in range(bins)]
-    dist = Counter(min((x - low) // step, bins - 1) for x in it)
+    dist: Dict[int, int] = Counter()
+    for size, count in data.items():
+        if size:
+            bucket = min(int((math.log(size) - low) // step), bins - 1)
+        else:
+            bucket = 0
```
pablogsal, line 31: Nit:

```suggestion
        bucket = min(int((math.log(size) - low) // step), bins - 1) if size else 0
```

pablogsal (changes requested):
We need to update the docs if we are removing the `-a` parameter and likely the text needs some tweaking as the semantics  are changing a bit.

src/memray/_memray.pyx
```diff
@@ -602,27 +614,104 @@ cdef class FileReader:
...
+    cdef ProgressIndicator progress_indicator = ProgressIndicator(
+        "Computing statistics",
+        total=total,
+        report_progress=report_progress,
+    )
+    with progress_indicator:
+        while True:
+            PyErr_CheckSignals()
+            ret = reader.nextRecord()
+            if ret == RecordResult.RecordResultAllocationRecord:
```
pablogsal, line 644: RecordResultRecordRecordRecordResultRecordEhIsARecordDontBeMistakenIsARecord

pablogsal: For the curious observer: I named this 😈

src/memray/_memray.pyx
```diff
@@ -602,27 +614,104 @@ cdef class FileReader:
...
-                        python_allocator=python_allocator,
-                        has_native_traces=self._header["native_traces"])
+    # Ignore the n_allocations in the header, use our observed value.
+    header["stats"]["n_allocations"] = progress_indicator.num_processed
+
+    # NOTE: We need to cast away the constness of some references below,
+    #       because Cython's iterator handling is not const-correct.
+
+    # Convert allocation counts by allocator/by size to Python dicts.
+    allocation_count_by_allocator = {
```
pablogsal, line 659: This is quite more expensive that it needs to be. This is creating an `AllocatorType` key and hashing it over for insertion. `AllocatorType` are enums in Python and those are quite expensive to build. It may be much faster to take the dictionary as it is (Cython automatically transforms C++ maps into dictionaries) and then create other where the keys are transformed.

src/memray/_memray.pyx
```diff
@@ -602,27 +614,104 @@ cdef class FileReader:
...
+        for id_and_count in <unordered_map[Allocator, uint64_t]&>(
+            aggregator.allocationCountByAllocator()
+        )
+    }
+    allocation_count_by_size = {
+        size_and_count.first: size_and_count.second
+        for size_and_count in <unordered_map[size_t, uint64_t]&>(
+            aggregator.allocationCountBySize()
+        )
+    }
```
pablogsal, line 670:
```suggestion
    allocation_count_by_size = aggregator.allocationCountBySize()
```

pablogsal: Same as before, Cython should be able to return this automatically, no?

src/memray/_memray.pyx
```diff
@@ -602,27 +614,104 @@ cdef class FileReader:
...
+        size_and_count = size_and_count_by_location[top_of_stack[0]]
+        size_and_count[0] += location_key_and_size_and_count.second.first
+        size_and_count[1] += location_key_and_size_and_count.second.second
+
+    # Now we've got sizes and counts by source location. Find the top locations.
+    top_locations_by_size = [
+        (loc, size_and_count[0])
+        for loc, size_and_count in size_and_count_by_location.items()
+    ]
+    top_locations_by_size.sort(key=lambda pair: pair[1], reverse=True)
```
pablogsal, line 693: Nit: Using `operator.itemgetter` may be better than the lambda

src/memray/_memray.pyx
```diff
@@ -602,27 +614,104 @@ cdef class FileReader:
...
+    del top_locations_by_size[num_largest:]
+
+    top_locations_by_count = [
+        (loc, size_and_count[1])
+        for loc, size_and_count in size_and_count_by_location.items()
+    ]
+    top_locations_by_count.sort(key=lambda pair: pair[1], reverse=True)
```
pablogsal, line 700: Nit: Using `operator.itemgetter` may be better than the lambda

src/memray/reporters/stats.py
godlygeek: This dict contains one entry per allocator, and we've got less than 16 of them. I'm gonna keep the more readable, less performant one, simply because this dict is too small to care.

src/memray/_memray.pyx
```diff
@@ -602,27 +614,104 @@ cdef class FileReader:
...
+        size_and_count[0] += location_key_and_size_and_count.second.first
+        size_and_count[1] += location_key_and_size_and_count.second.second
+
+    # Now we've got sizes and counts by source location. Find the top locations.
+    top_locations_by_size = [
+        (loc, size_and_count[0])
+        for loc, size_and_count in size_and_count_by_location.items()
+    ]
+    top_locations_by_size.sort(key=lambda pair: pair[1], reverse=True)
+    del top_locations_by_size[num_largest:]
```
pablogsal, line 694: We may be able to just use `heapq.nlargest` instead of the construction + sort, but I don't know how better or worse in perf would be.

src/memray/_memray.pyx
godlygeek: Discussed offline: this dict contains each allocator only once, so the cost of constructing this dict is absolutely trivial compared to all the work we just did reading the file.

The cost of hashing the `AllocatorType`, along with the cost of `+=` on a Python `int`, are the reason that I chose to build this up as a `unordered_map<int, size_t>` and only convert it to a Python `dict[AllocatorType, int]` at the very end.

src/memray/_memray.pyx
godlygeek: I couldn't find the right magical incantation to make that happen, but Pablo and I played with it offline and got this working.

src/memray/_memray.pyx
godlygeek: I rewrote this to use `heapq.nlargest` and avoided the need for `key` entirely.

src/memray/_memray.pyx
godlygeek: I rewrote this to use `heapq.nlargest` and avoided the need for `key` entirely.

src/memray/_memray.pyx
godlygeek: It seems to make very little performance difference one way or the other, but it may scale a bit better in extreme cases, so I've implemented this.

godlygeek:
I think I've addressed all of your comments except the note that the docs likely need to be reviewed. I'll check those out on Monday and see what changes are needed.

godlygeek:
I've fixed up the docs (only required removing a few sentences) and added a NEWS entry describing the speedups and the fact that it's now easier to interpret the results.

I profiled on a variety of files, and I found that this new implementation was at least 20% faster and up to 50% faster than the one on `main`, even though this implementation looks at all allocations and the one on `main` only looks at high water mark allocations.

godlygeek:
I pushed another commit that performs the aggregation in C++ by location rather than by stack, to avoid the need for a second aggregation pass in the Cython code (and to eliminate some ugly code that neither of us really liked).

It has the unfortunate effect of exposing frame IDs from the reader layer to the Cython code and aggregator, though (previously, only stack IDs were exposed).

I used an `unordered_map` for the aggregation - it seems to be around 5% faster to use a vector instead, but that introduces a dependency on even more implementation details (namely, that frame IDs are monotonically increasing numbers that start at 0). I've got the `vector` code stashed if you'd prefer to take the faster but less maintainable version over the slower but more maintainable one.

pablogsal:
@godlygeek Seems that you need to rebase :(

pablogsal (approved):
LGTM

PR #138, opened 2022-06-08
Support Python 3.11

godlygeek:


pablogsal:
We must update the most important element: https://camo.githubusercontent.com/778662d3cb46324d345de692250db09cdb613c4fca64f51cf75c839605fee837/68747470733a2f2f696d672e736869656c64732e696f2f707970692f707976657273696f6e732f6d656d726179

😆

godlygeek:
> We must update the most important element

I think that badge is based on the `"Programming Language :: Python :: 3.10",` etc classifiers on the latest version of the package published to PyPI, so I don't think we can change that without publishing a new sdist. Correct me if I'm wrong.

I did miss adding the classifier for 3.11, so I've added that, along with a changelog entry stating that we now support 3.11

src/memray/_memray.pyx
```diff
@@ -4,6 +4,14 @@ import os
 import pathlib
 import sys
 
+
+cdef extern from *:
```
pablogsal, line 8: Please, add a comment here about why are we doing this. Notice that this should go in the next beta release as I fixed this upstream.

src/memray/_memray/hooks.h
```diff
@@ -1,6 +1,12 @@
 #ifndef _MEMRAY_HOOKS_H
 #define _MEMRAY_HOOKS_H
 
+#define PY_SSIZE_T_CLEAN
+#include <Python.h>
+#ifdef static_assert
```
pablogsal, line 6: Same as https://github.com/bloomberg/memray/pull/138/files#r898939038

src/memray/_memray.pyx
godlygeek: > Please, add a comment here about why are we doing this.

Done

> Notice that this should go in the next beta release as I fixed this upstream.

I also added a check on `PY_VERSION_HEX` to capture the fact that this only applies to 3.11 <= version < 3.11.0b4.

pablogsal:
I'm planning to release Python 3.11.0b4 today or tomorrow so we can hold this so we can remove all the `static_assert` workarounds.

godlygeek:
Several tests passed on 3.11.0b3 but are failing on 3.11.0b4.

`tests/integration/test_api.py::test_file_destination` and `tests/integration/test_api.py::test_file_destination_str_path` passed on 3.11.0b3 but are failing on 3.11.0b4. They're failing because they see an extra allocation that they didn't expect, with this stack:
```py
[('_PyMem_RawMalloc', 'Objects/obmalloc.c', 101),
   ('PyMem_RawMalloc', 'Objects/obmalloc.c', 586),
   ('_PyObject_Malloc', 'Objects/obmalloc.c', 2003),
   ('_PyObject_Malloc', 'Objects/obmalloc.c', 1996),
   ('PyMem_Malloc', 'Objects/obmalloc.c', 623),
   ('_PyCode_CreateLineArray', 'Objects/codeobject.c', 724),
   ('_PyCode_InitLineArray', './Include/internal/pycore_code.h', 529),
   ('call_trace', 'Python/ceval.c', 6866),
   ('call_trace', 'Python/ceval.c', 6850),
   ('trace_call_function', 'Python/ceval.c', 7298),
   ('_PyEval_EvalFrameDefault', 'Python/ceval.c', 4771),
   ('_PyEval_EvalFrame', './Include/internal/pycore_ceval.h', 73),
```

Looks like that changed in https://github.com/python/cpython/commit/ab0e60101637b7cf47f3cc95813996791e7118c4

`tests/integration/test_tracing.py::test_traceback_for_high_watermark` also fails because of this, though for a slightly different reason. It doesn't fail just because it sees an extra allocation, but rather because the high watermark is now detected to be a point at which the only allocation is the one performed by `_PyCode_CreateLineArray`.

`tests/integration/test_tracing.py::test_large_number_of_frame_pops_between_subsequent_allocations` also fails because of this, for an even more frustrating reason: this malloc in `_PyCode_CreateLineArray` happens below our `mmap.mmap` call, which means it shares exactly the same Python stack. Our high water mark aggregator is grouping this malloc together with our mmap, and we're seeing a single combined allocation record with an allocation count of 2, and our test goes on to fail because this combined record doesn't have the size we expected (instead it's the sum of the size we expected and the size of the allocation performed by `_PyCode_CreateLineArray`).

godlygeek:
And, the current `manylinux2010` image doesn't include 3.11 anymore. See https://github.com/pypa/manylinux/commit/97c4ab0bc6f0d9a44f60e7938b025c69788a89b7 - it was dropped because `manylinux2010` will officially unsupported before Python 3.11 becomes ABI stable. Discussed a bit in https://github.com/pypa/cibuildwheel/issues/1187

godlygeek:
I've added workarounds for the 4 failing tests, and updated the cibuildwheel config to build manylinux2014 wheels for 3.11+, keeping manylinux2010 wheels for Python 3.10 and earlier. I couldn't figure out a way to make the manylinux version specified in the pyproject.toml make it into the name of the GitHub Actions job, so we'll need to update our branch protection rules to stop requiring `Wheel for cp37-*-x86_64-manylinux2010` and start requiring `Wheel for cp37-*-x86_64`, but that seems like no big deal.

src/memray/_memray/tracking_api.cpp
```diff
@@ -32,12 +33,38 @@ get_executable()
...
+#endif
+}
+
 static bool
 starts_with(const std::string& haystack, const std::string_view& needle)
 {
     return haystack.compare(0, needle.size(), needle) == 0;
 }
 
+PyCodeObject*
```
pablogsal, line 53:
```suggestion
static inline PyCodeObject*
```

src/memray/_memray/tracking_api.cpp
```diff
@@ -32,12 +33,38 @@ get_executable()
     return std::string(buff, len);
 }
 
+bool
```
pablogsal, line 36:
```suggestion
static inline bool
```

pablogsal (approved):
LGTM except for two small nits

Great job!

pablogsal:
I have updated the CI required targets to these PR names so we can land the PR and move forward with the new target names.

src/memray/_memray/tracking_api.cpp
godlygeek: OK, sure. I don't think this matters much, though - this is in an unnamed namespace, so `static` doesn't really have any effect, and given that we're using LTO, there's no chance this wasn't already being inlined.

Doesn't hurt to be explicit, though.

godlygeek:
Wait, I screwed that force push up. Do over...

pablogsal (approved):

gaborbernat:
When do we plan to release this?

godlygeek:
Soon-ish. The earliest we could have released wheels for this was last week (when 3.11.0rc1 was released and the ABI was frozen). Now we're unblocked on that front, but we're going to finish up the work for supporting Mac OS X before we cut our next release.

I expect to release this by the end of the month, though, if nothing surprising comes up.

PR #141, opened 2022-06-11
Always leverage fast unwinding to obtain the native stack

pablogsal:
Seems that libunwind just chokes if the function that calls unw_get_reg
is not inlined in aarch64. Additionally, there are some problems with
inlined functions that call libunwind's APIs in powerpc and other
architectures.

To avoid all this trouble, leverage just the fast unwinding APIs, which
will make the program counters more consistent and will eliminate the
inline shenanigans.

Closes: https://github.com/bloomberg/memray/issues/126

### DIFF
--- a/news/141.bugfix.rst
+++ b/news/141.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a bug where aggregating native call stacks could give misleading results on aarch64 under some circumstances.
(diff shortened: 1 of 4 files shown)


godlygeek (dismissed):
LGTM, but one suggestion. Feel free to ignore and land if you don't want to take the suggestion.

godlygeek:
Also - does this deserve a changelog entry? This fixes a bug where aggregating call stacks could give misleading results on aarch64 under some circumstances, right?

pablogsal:
> This fixes a bug where aggregating call stacks could give misleading results on aarch64 under some circumstances, right?

Not really, is not user-visible because the actual stack traces were the same, is just that the test suite fails because the native_frame_ids are different although they produce the same results.

pablogsal:
@godlygeek Can you please review again?

PR #142, opened 2022-06-11
Add plot with heap size to the flamegraph/table memory plot

pablogsal:
To allow users to compare resident size and allocated size over time,
add a new plot line to the graphs that we show in the different reporter
that shows also allocated memory size.

This commit also changes the memory records to include the heap size as
well as the resident size.


news/142.feature
```diff
@@ -0,0 +1 @@
+Include a line with the heap size along with the resident size in the memory plot that is shown in the html-based reporters.
```
godlygeek, line 1:
```suggestion
Add a line showing the heap size over time to the memory plot in the html-based reporters which already showed the resident size over time.
```

src/memray/_memray/records.h
```diff
@@ -93,6 +93,7 @@ struct MemoryRecord
 {
     unsigned long int ms_since_epoch;
     size_t rss;
+    size_t heap;
```
godlygeek, line 96: Hm, this is quite ugly. This same `MemoryRecord` struct is used on the tracking side and on the reader side. On the tracking side, the `heap` field is never set (or even initialized!), and it is ignored by the `RecordWriter`. On the reader side, the `RecordReader` returns a `MemoryRecord` whose `heap` field is uninitialized to the Cython layer, and the Cython layer then copies the two valid fields out of that `MemoryRecord` into a new `MemoryRecord`, ignoring the uninitialized and invalid `heap` value, and provides a valid value for `heap` that it computed independently.

I think we really want two different structures instead - I think `MemoryRecord` should not get a `heap` field added to it, and the `vector[_MemoryRecord]` in the Cython layer should become a vector of a new type that has the additional field.

tests/integration/test_tracking.py
```diff
@@ -1116,6 +1116,7 @@ def test_memory_records_are_written(self, tmp_path):
 
         assert memory_records
         assert all(record.rss > 0 for record in memory_records)
+        assert all(record.heap > 0 for record in memory_records)
```
godlygeek, line 1119: Isn't this racy? If a `MemoryRecord` gets written after the `allocator.free()` call and before the `Tracker.__exit__` call, won't it have `record.heap` set to 0? Likewise for one written after `Tracker.__enter__` and before `allocator.valloc()`

tests/integration/test_tracking.py
```diff
@@ -1136,6 +1137,7 @@ def test_memory_records_tick_interval(self, tmp_path):
 
         assert len(memory_records)
         assert all(record.rss > 0 for record in memory_records)
+        assert all(record.heap > 0 for record in memory_records)
```
godlygeek, line 1140: As above, seems racy

godlygeek (changes requested):

tests/integration/test_tracking.py
pablogsal: I changed this to check if a heap field equals the heap size at some point, which is less racy and probably more aligned to what we want to check here.

tests/integration/test_tracking.py
pablogsal: I changed this to check if a heap field equals the heap size at some point, which is less racy and probably more aligned to what we want to check here.

src/memray/_memray/records.h
pablogsal: Whops, sorry, this was actually done in another commit but I messed it up :(

I have pushed a fixed version that uses a `MemorySnapshot` to not interfere with the native side.

godlygeek (approved):
LGTM. `MemorySnapshot` is definitely a better name for the Python type that will one day be exposed as part of a public API, too.

PR #147, opened 2022-06-14
Fix several live mode bugs

### LINKED ISSUE #143: ZeroDivisionError on TUI
## Bug Report

**Current Behavior** TUI crashes with a `ZeroDivisionError` when pressing "right arrow" after program has completed.

**Input Code**

Run `memray run --live -c 'print(1)'`. It completes shortly, and the TUI opens saying "Thread 1 of 0". Pressing the right arrow key to cycle between threads crashes with:

```
  File "/.../venv/bin/memray", line 8, in <module>
    sys.exit(main())
  File "/.../venv/lib/python3.9/site-packages/memray/commands/__init__.py", line 124, in main
    arg_values.entrypoint(arg_values, parser)
  File "/.../venv/lib/python3.9/site-packages/memray/commands/run.py", line 308, in run
    _run_child_process_and_attach(args)
  File "/.../venv/lib/python3.9/site-packages/memray/commands/run.py", line 117, in _run_child_process_and_attach
    raise error from None
  File "/.../venv/lib/python3.9/site-packages/memray/commands/run.py", line 114, in _run_child_process_and_attach
    LiveCommand().start_live_interface(port)
  File "/.../venv/lib/python3.9/site-packages/memray/commands/live.py", line 110, in start_live_interface
    tui.previous_thread()
  File "/.../venv/lib/python3.9/site-packages/memray/reporters/tui.py", line 371, in previous_thread
    self._thread_idx = (self._thread_idx - 1) % len(self._threads)
ZeroDivisionError: integer division or modulo by zero
```

**Expected behavior/code** no crash

**Environment**

- Python(s): 3.9

**Possible Solution**

n/a

**Additional context/Screenshots** n/a

### LINKED ISSUE #145: Incorrect quoting of popen arguments in `memray run --live`
## Bug Report

This command works as expected:
```sh
memray run --live -c "print('hi')"
```

This command hangs forever:
```sh
memray run --live -c 'print("hi")'
```

`--live` mode tries to fork off a subprocess and run the command in that subprocess, and the way the arguments are being marshalled from the parent process to the child process fails if double quotes are used.

Plus, the parent process hangs forever if the child process fails to start up, so that's a separate issue.

### LINKED ISSUE #146: Delay before TUI appears in `memray run --live`
## Bug Report

For some reason
```sh
memray run --live -c "__import__('time').sleep(300)"
```
takes more than 5 seconds before it starts showing any output.

godlygeek:
Closes #143 
Closes #145
Closes #146

### DIFF
--- a/src/memray/commands/run.py
+++ b/src/memray/commands/run.py
@@ -95,13 +95,12 @@ def _run_child_process_and_attach(args: argparse.Namespace) -> None:
 
     arguments = (
         f"{port},{args.native},{args.run_as_module},{args.run_as_cmd},{args.quiet},"
-        f'"{args.script}",{args.script_args}'
+        f"{args.script!r},{args.script_args}"
     )
     tracked_app_cmd = [
         sys.executable,
         "-c",
-        f"from memray.commands.run import _child_process;"
-        f"_child_process({arguments})",
+        f"from memray.commands.run import _child_process;_child_process({arguments})",
     ]
     with contextlib.suppress(KeyboardInterrupt):
         with subprocess.Popen(
--- a/news/147.bugfix.1.rst
+++ b/news/147.bugfix.1.rst
@@ -1,0 +1,1 @@
+Fix the handling of the thread switch commands in the live mode TUI before the first allocation has been seen.
--- a/news/147.bugfix.2.rst
+++ b/news/147.bugfix.2.rst
@@ -1,0 +1,1 @@
+Fix a bug that made ``memray run --live -c`` fail if the command to run contained double quotes.
--- a/news/147.bugfix.3.rst
+++ b/news/147.bugfix.3.rst
@@ -1,0 +1,1 @@
+Ensure our TUI isn't displaying stale data by periodically flushing the latest available data from the tracker (rather than only flushing when a buffer fills up).
--- a/src/memray/_memray/record_writer.h
+++ b/src/memray/_memray/record_writer.h
@@ -162,7 +162,7 @@ bool inline RecordWriter::writeRecordUnsafe(const MemoryRecord& record)
 {
     RecordTypeAndFlags token{RecordType::MEMORY_RECORD, 0};
     return writeSimpleType(token) && writeVarint(record.rss)
-           && writeVarint(record.ms_since_epoch - d_stats.start_time);
+           && writeVarint(record.ms_since_epoch - d_stats.start_time) && d_sink->flush();
 }
 
 bool inline RecordWriter::writeRecordUnsafe(const ContextSwitch& record)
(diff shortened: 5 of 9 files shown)


src/memray/commands/run.py
```diff
@@ -66,24 +66,17 @@ def _run_tracker(
...
+def _child_process() -> None:
+    params = json.loads(sys.stdin.read())
     args = argparse.Namespace(
-        native=native,
-        run_as_module=run_as_module,
-        run_as_cmd=run_as_cmd,
-        quiet=quiet,
-        script=script,
-        script_args=script_args,
+        native=params["native"],
```
pablogsal, line 72: Ugh, this is a bit ugly, tbh. I would prefer to only scape the stuff that we need to scape instead of affecting all parameters. I don't feel very strongly, though, in case you do

godlygeek: I'm not sure I follow what you mean. If you think that the previous approach - formatting user-provided strings into a string of Python source code - was cleaner than the approach of serializing parameters and sending them to the process over stdin, I'm shocked.

I also don't want some hybrid of the two approaches where we handle fields differently depending on whether or not they contain strings.

But if you do prefer the previous approach, it should be possible to make it work correctly, instead - the Python injection approach isn't intrinsically wrong, it just requires correct escaping...

pablogsal: > I'm not sure I follow what you mean. If you think that the previous approach - formatting user-provided strings into a string of Python source code - was cleaner than the approach of serializing parameters and sending them to the process over stdin, I'm shocked.

No, I am saying that we are jsonifiying all arguments here where a bunch of them are just `bools`. If in the future we need to pass some Python object that cannot be serialized this may be annoying so I am advocating to only serialize the arguments that need to be sanitized.

pablogsal: Yeah, I find it a bit cleaner than adding tons of dict access, but I am happy to go with the other if you prefer it

pablogsal: > I don't like that this leaves `ps` looking like:

If you are referring to the `_child_process(49241,False,False,True,False`, that's a good point, but one that's fixable with keyword arguments :)

godlygeek: I'm more talking about the fact that it embeds the entire state of an `argparse.parse` call into command line arguments that show up in `ps`. It feels like it's leaking implementation details. I also don't like that it takes up the entire width of my screen, it's weird.

But that's mostly personal preference, I guess, and I'm happy to call it out of scope for the bug fix.

godlygeek: > Do you prefer https://github.com/bloomberg/memray/commit/c923a7541f1b7b5b8148dd29759604cffa06c5ae ?

Make that https://github.com/bloomberg/memray/pull/147/commits/bacbb94ae4da66f660dfff5ef338f0d5547dc1db (amended because I missed the news entry for the bugfix).

PR #150, opened 2022-07-05
Ensure we can read a file with no allocations

pablogsal:
Due to how we handle trailing 0 in files, if a file has no allocations
we will fail correctly reading the Python allocator as currently is 4
bytes, but the last 3 are always 0. If there aren't any other records
after it we will incorrectly invalidate all trailing zeroes, which will
fail reading the Python allocator.

This technically requires a more complex fix, but for now we can make
a simple change and make the Python allocator 1 byte, which will not
suffer from this problem.

### DIFF
--- a/src/memray/_memray/records.h
+++ b/src/memray/_memray/records.h
@@ -14,7 +14,7 @@
 namespace memray::tracking_api {
 
 const char MAGIC[] = "memray";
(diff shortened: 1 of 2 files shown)


godlygeek (dismissed):
LGTM

godlygeek (approved):
LGTM

PR #152, opened 2022-07-08
Capture running threads' Python stacks

### LINKED ISSUE #130: Use api that "stack trace unavailable" in sub-thread
I had a situation that main thread has a function and its memory grows up slowly, at the same time I open a sub-thread to monitor this

```
import time
from threading import Thread
import memray

def hello():
    a = 'h' * 10240
    count = 0
    while count<15:
        a += a
        count += 1

def m():
    with memray.Tracker("test.bin"):
        time.sleep(2)

t = Thread(target=m, args=())
t.start()
hello()
t.join()
```

```
root@b55328e153b0:/tmp# memray stats test.bin
📏 Total allocations:
	2

📦 Total memory allocated:
	480.000MB

📊 Histogram of allocation size:
	min: 480.000MB
	----------------------------------------
	< 225.000PB: 1 ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇
	< 105.5YB  : 0
	< 53084201343.8YB: 0
	< 26718158566205665280.0YB: 0
	< 13447692140004132921498664960.0YB: 0
	< 6768446389904240393238904681458040832.0YB: 0
	< 3406671275343084001598671104068678912850264064.0YB: 0
	< 1714634128676590725640136232502358247467182517297086464.0YB: 0
	< 863003782167520542775850501504234338770282486944023855555936256.0YB: 0
	<=434364110441617396824120646697741245745787014207085330508234065870585856.0YB: 0
	----------------------------------------
	max: 480.000MB

📂 Allocator type distribution:
	 MALLOC: 1

🥇 Top 5 largest allocating locations (by size):
	- <stack trace unavailable> -> 480.000MB

🥇 Top 5 largest allocating locations (by number of allocations):
	- <stack trace unavailable> -> 2
```

in the end, "stack trace unavailable" is printed on the screen , I can't get the detail which line consume memory

godlygeek:
When starting tracking from one thread, previously we were not able to
capture any information about the Python stack in some other
already-running thread, though we would capture stacks in threads
spawned later. To improve this situation, capture the Python stack of
every existing thread when tracing starts, and make it available to the
allocation recording machinery in those already-running threads.

Closes #130


godlygeek:
This needs some new tests to be added to the test suite before I'll be convinced that I didn't miss anything. I'm just putting the PR in now to let CI run and make sure I didn't screw anything up regarding portability across Python versions.

godlygeek:
I've added some tests, and proved that this fixes both the issue reported in #130 as well as a bug that we had previously not noticed: if a background thread continued running across two different tracking sessions, it would have incorrect stacks reported for allocations captured during the second session if its Python stack changed between when the first tracking session ended and the second started.

I'm marking this as ready for review.

godlygeek:
Actually, there is another thing that we'll need to discuss before we decide whether/how to proceed with this. This will leave us with two methods of collecting stacks - incrementally by monitoring frame pushes/pops using our profile function, and all at once by grabbing a Python frame and walking the stack from it. These two approaches will give different results for Cython frames when Cython profiling is enabled, because Cython calls the profile function as profiled Cython function calls are pushed/popped, but these frames are not visible as part of the execution stack visible to the interpreter.

src/memray/_memray/tracking_api.cpp
```diff
@@ -197,6 +252,7 @@ PythonStackTracker::emitPendingPushes()
 inline int
 PythonStackTracker::getCurrentPythonLineNumber()
 {
+    reloadStackIfTrackerChanged();
```
pablogsal, line 255: This is called  twice in the code path from `pushPythonFrame` and maybe others. Maybe we should move this to the callers (`pushPythonFrame` and `trackAllocationImpl`). Seems that that way we may also cover other stuff because `trackAllocationImpl` also calls `emitPendingPops` that also calls `reloadStackIfTrackerChanged` so we can eliminate that call as well.

src/memray/_memray/tracking_api.cpp
```diff
@@ -219,9 +275,47 @@ PythonStackTracker::setMostRecentFrameLineNumber(int lineno)
     }
 }
 
+void
+PythonStackTracker::reloadStackIfTrackerChanged()
```
pablogsal, line 279: I am a bit concerned with this function. Is very easy to miss some situation when this needs to be called or to also overdo it. What about some alternative design: we add a getter/setter around the `d_stack` vector for the frames and we check the generation number then and act appropriately. This may have some downsides but I think that will be always correct, no?

pablogsal (changes requested):

src/memray/_memray/tracking_api.cpp
```diff
@@ -336,11 +535,11 @@ Tracker::~Tracker()
     tracking_api::Tracker::deactivate();
     PythonStackTracker::s_native_tracking_enabled = false;
     d_background_thread->stop();
-    t_python_stack_tracker.reset(nullptr);
     d_patcher.restore_symbols();
     if (d_trace_python_allocators) {
         unregisterPymallocHooks();
     }
+    PythonStackTracker::stopTracking();
```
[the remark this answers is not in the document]
godlygeek: Yeah, the whole name "PythonStackTracker" is maybe not great and needs a rethink, honestly.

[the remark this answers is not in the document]
pablogsal: May be also time to move that to another file as well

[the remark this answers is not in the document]
godlygeek: I'll rename `startTracking` and `stopTracking` to `installProfileHooks` and `removeProfileHooks`.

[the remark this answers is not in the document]
godlygeek: > May be also time to move that to another file as well

Let's defer that for now. When we initially profiled this we found that having all the Python stack TLS variables in the same translation unit as the allocation tracking made things measurably faster. I'm reluctant to move this without spending some effort on profiling to ensure it doesn't cause performance to regress.

src/memray/_memray/tracking_api.cpp
```diff
@@ -50,6 +50,23 @@ starts_with(const std::string& haystack, const std::string_view& needle)
     return haystack.compare(0, needle.size(), needle) == 0;
 }
 
+static inline PyFrameObject*
```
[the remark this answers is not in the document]
godlygeek: Some of these are already there on `main`. I agree with you, but let's tackle it in another PR.

src/memray/_memray/tracking_api.cpp
godlygeek: We discussed a few options offline for improving this, but we couldn't come up with anything better than adding a getter that handles reloading the stack when fetching the stack tracker. I've implemented that.

src/memray/_memray/tracking_api.cpp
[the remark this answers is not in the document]
godlygeek: Eh, nevermind. I just did it with another commit to this PR, even though it refactors a little bit of code unrelated to this change.

src/memray/_memray/compat.cpp
```diff
@@ -0,0 +1,40 @@
+#include "compat.h"
```
pablogsal, line 1: spy-x-family-spy-family

src/memray/_memray/tracking_api.cpp
```diff
@@ -284,6 +321,101 @@ std::unique_ptr<Tracker> Tracker::d_instance_owner;
...
+         tstate != nullptr;
+         tstate = PyThreadState_Next(tstate))
+    {
+        PyFrameObject* frame = compat::threadStateGetFrame(tstate);
+        if (frame) {
+            stack_by_thread[tstate] = pythonFrameToStack(frame);
+            if (PyErr_Occurred()) {
+                throw std::runtime_error("Failed to capture a thread's Python stack");
+            }
+        }
```
pablogsal, line 370: Sorry, I am allergic to indentation

```suggestion
        if (!frame) {
            continue;
        }            
        stack_by_thread[tstate] = pythonFrameToStack(frame);
        if (PyErr_Occurred()) {
            throw std::runtime_error("Failed to capture a thread's Python stack");
        }
```

pablogsal (approved):
I have tested this extensively in some weird threading configurations and could not make it crash or report nonsensical stuff, so I think we are good to go.

LGTM

tests/integration/test_tracing.py
```diff
@@ -535,6 +596,228 @@ def tracking_function():
...
+    alloc2_funcs = [frame[0] for frame in second.stack_trace()]
+
+    # Cython frames called before tracking started aren't in the Python stack.
+    assert alloc1_funcs[:3] == ["valloc", "func1", "thread_body"]
+
+    # But we do get pop events for them if tracking was enabled when they were
+    # pushed. Ensure we handle these unexpected pops.
+    assert alloc2_funcs[:2] == ["valloc", "thread_body"]
+
+
```
pablogsal, line 820: Would be great if we could add a test that's doing some allocations in native code before the tracker gets installed

godlygeek:
> I have tested this extensively in some weird threading configurations and could not make it crash or report nonsensical stuff

Are any of those weird configurations weird enough to justify an addition to the test suite?

tests/integration/test_tracing.py
godlygeek: I don't follow. If it's doing allocations before the tracker gets installed, we won't see those allocations, because our malloc hooks aren't installed yet. What would the test be looking for?

pablogsal:
> Are any of those weird configurations weird enough to justify an addition to the test suite?

No, is mostly hundreds of threads getting created, and destroyed while the GIL is dropped and the tracker is randomly created in some of these threads. I don't think we want this on the test suite :)

src/memray/_memray/compat.cpp
```diff
@@ -0,0 +1,40 @@
...
+namespace memray::compat {
+
+void
+setprofileAllThreads(Py_tracefunc func, PyObject* arg)
+{
+    assert(PyGILState_Check());
+
+    PyThreadState* this_tstate = PyThreadState_Get();
+    PyInterpreterState* interp = threadStateGetInterpreter(this_tstate);
+    for (PyThreadState* tstate = PyInterpreterState_ThreadHead(interp); tstate != nullptr;
```
pablogsal, line 12: I think we may have a problem here. Is possible that calling `_PyEval_SetProfile` drops the GIL and that can do random stuff with threads? What's worse, `PyThreadState_Next` can be called at the same time a native extension destroys a thread state (this is a known CPython problem). 

I don't think we can do anything here until we have the new API I am working on

pablogsal:
@godlygeek I think you need to rebase the fixups to make the commit test pass :)

src/memray/_memray/compat.cpp
godlygeek: > It's possible that calling `_PyEval_SetProfile` drops the GIL and that can do random stuff with threads, like creating new ones or destroying them (this may not be a problem, though).

`_PyEval_SetProfile` can only drop the GIL if either a) the `__del__` of the previous profile arg fires, or b) an audit hook is installed for `sys.setprofile`, right? We work around the first one with a call to `setprofileAllThreads(nullptr, nullptr)` before we try to do anything interesting (albeit it's possible that causes a user supplied `__del__` to run and call `sys.setprofile` itself, at least before 3.11, in which case we're screwed).

There's not much that we can do about the audit function running and dropping the GIL, though. Interestingly, that can't happen in 3.7 or 3.8 thanks to my backport of `_PyEval_SetProfile`...

Hm.

> What's worse, `PyThreadState_Next` can be called at the same time a native extension destroys a thread state

Ah. Yep. That's bad.

pablogsal (approved):

tests/integration/test_tracing.py
godlygeek: Pablo and I chatted offline, and he explained his goal in a bit more detail. He wants a test that sets up a Python call stack and then drops the GIL in C code. While the GIL is dropped, another thread installs a tracker. After the tracker is installed, the first thread performs an allocation, without reacquiring the GIL. We expect the Python stack associated with that allocation to be correct, but no test currently covers that case. I'll add one.

src/memray/_memray/compat.cpp
godlygeek: But there's nothing we can do about it. Sounds like this will eventually be improved in CPython, and in the meantime we believe that this is unlikely to cause problems.

PR #165, opened 2022-07-25
Add Tracker.{start,stop}

jonashaag:
*Issue number of the reported bug or feature request: None*

**Describe your changes**
Add `Tracker.{start,stop}` methods that are equivalent to `with Tracker(...):`. This helps profiling code that cannot be enclosed in a `with` statement.

**Testing performed**
Added whitebox unit test.

**Additional context**
-

### DIFF
--- a/src/memray/_memray.pyx
+++ b/src/memray/_memray.pyx
@@ -238,6 +238,16 @@ cdef class Tracker:
     re-enable tracking after the ``with`` block ends, you will need to create
     a fresh `Tracker` instance.
 
+    You can also explictly call the ``start`` and ``stop`` methods instead of
+    using the context manager::
+
+        tracker = memray.Tracker(...)
+        tracker.start()
+        try:
+            ...
+        finally:
+            tracker.stop()
+
     Args:
         file_name (str or pathlib.Path): The name of the file to write the
             captured allocations into. This is the only argument that can be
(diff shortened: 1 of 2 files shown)


godlygeek:
I'm not sure how I feel about this - outside of pretty advanced use cases I wouldn't want people doing this, and for those advanced use cases people could always call `tracker.__enter__()` and `tracker.__exit__(None, None, None)` directly.

If we do decide to move forward with this, it needs quite a few more tests, since this introduces many more failure modes to the public API. Off the top of my head:
1. Calling `start()` and never calling `stop()`
2. Calling `start()` after `__enter__` (e.g. `with memray.Tracker("some_output_file.bin") as tracker: tracker.start()`) (or, equivalently, calling `start()` twice)
3. Calling `stop()` after `__exit__` (or, equivalently, calling `stop()` twice)
4. Calling `stop()` on a never-started `Tracker`

@pablogsal What are your thoughts?

godlygeek:
Pablo and I discussed this offline. At this point, we're not interested in exposing an API like this, because we worry that it will be too prone to misuse, and that it adds a lot of testing complexity.

Someone who knows what they're doing can always directly call `tracker.__enter__()` and `tracker.__exit__(None, None, None)` directly. That both gives them the feature that is being requested here, _and_ makes it clear that they're taking correctness into their own hands and are responsible for ensuring that `__exit__` eventually gets called.

If we find more demand for this feature request in the future we might revisit it, but at this point, the costs seem to outweigh the benefits.

PR #174, opened 2022-08-10
Add macOS support

pablogsal:


pablogsal:
We probably should:

* Warn for now if you use memray on macOS (call it "experimental") ?
* Definitely hard fail if someone passes --native on macOS

src/memray/_memray/macho_utils.h
```diff
@@ -0,0 +1,109 @@
...
+#    define SEG_AUTH_CONST "__AUTH_CONST"
+#endif
+
+struct DynamicInfoTable
+{
+    using segment_filter_t = std::function<bool(const char* segname)>;
+
+    const nlist_t* symbol_table;
+    const char* string_table;
+    const uint32_t* dynsym_table;
```
godlygeek, line 48: These need to be initialized to `nullptr`. As it is, the constructor can return without ever setting them, in which case calling `operator bool` or `getSymbol` would be UB.

src/memray/_memray/macho_shenanigans.cpp
```diff
@@ -0,0 +1,145 @@
...
+        vm_size_t section_size,
+        bool restore_original)
+{
+    vm_size_t trunc_address = trunc_page((vm_size_t)addr);
+    vm_size_t trunc_size = (vm_size_t)addr - trunc_address;
+
+    kern_return_t err = vm_protect(
+            mach_task_self(),
+            (uintptr_t)trunc_address,
+            section_size + trunc_size,
```
godlygeek, line 31: Do we need to round down to the page boundary? https://web.mit.edu/darwin/src/modules/xnu/osfmk/man/vm_protect.html says

> The region starts at the beginning of the virtual page containing address; it ends at the end of the virtual page containing address + size - 1.

If that's right, it seems like just using `addr` as the address and `section_size` as the size ought to work.

src/memray/_memray/macho_shenanigans.cpp
```diff
@@ -0,0 +1,145 @@
...
+            0,
+            VM_PROT_READ | VM_PROT_WRITE | VM_PROT_COPY);
+    if (err == KERN_SUCCESS) {
+        auto typedAddr = reinterpret_cast<typename Hook::signature_t*>(addr);
+        *typedAddr = restore_original ? hook.d_original : intercept;
+        LOG_IF_DEBUG(symname << " intercepted!");
+    } else {
+        LOG(ERROR) << "Failed to patch " << symname;
```
godlygeek, line 39: Should we do anything else here besides log an error?

src/memray/_memray/macho_shenanigans.cpp
```diff
@@ -0,0 +1,145 @@
...
+patch_symbols_in_section(
+        const section_t* section,
+        uintptr_t slide,
+        const DynamicInfoTable& dyninfo_table,
+        bool restore_original)
+{
+    auto symbol_addr_table = reinterpret_cast<void**>(slide + section->addr);
+    for (unsigned int i = 0; i < section->size / sizeof(void*); i++) {
+        const char* symbol_name = dyninfo_table.getSymbol(section->reserved1, i);
+        if (!symbol_name || !symbol_name[0] || !symbol_name[1]) {
```
godlygeek, line 53: Why the `|| !symbol_name[1]` - seems to be ignoring symbols with single-character names, but why?

Oh wait, I think I figured it out - the first character is some dummy letter like `.` or `_` and that's why our strcmp below is ignoring it, right? If that's the case, should we be checking that it's exactly that character that we expect, instead of just anything but `'\0'`?

src/memray/_memray/macho_shenanigans.cpp
```diff
@@ -0,0 +1,145 @@
...
+patch_symbols_in_shared_object(const struct mach_header* header, intptr_t slide, bool restore_original)
+{
+    Dl_info info;
+    if (dladdr(header, &info) == 0) {
+        return;
+    }
+
+    const char* image_name = info.dli_fname;
+    if (strstr(image_name, "memray.cpython") || strstr(image_name, "/dyld")
+        || strstr(image_name, "dyld.dylib"))
```
godlygeek, line 82: How stable are these names? It looks like you're matching "dyld" in two different ways, and I'm surprised to see ".cpython" in the name of the "memray" one - do we have to worry about these changing over time?

src/memray/_memray/macho_shenanigans.cpp
```diff
@@ -0,0 +1,145 @@
...
+    LOG_IF_DEBUG("Patching symbols for " << image_name);
+
+    auto segment_filter = [](const char* seg_name) {
+        return strcmp(seg_name, SEG_DATA) == 0 || strcmp(seg_name, SEG_DATA_CONST) == 0
+               || strcmp(seg_name, SEG_AUTH_CONST) == 0;
+    };
+
+    DynamicInfoTable dyninfo_table(header, slide, segment_filter);
+    if (!dyninfo_table) {
+        LOG_IF_DEBUG("Could not construct dynamic information table" << image_name);
```
godlygeek, line 96: Should this be an error, instead? Is this a condition that we expect to happen with some dynamic libraries, or would every occurrence of this indicate a problem with our MachO parsing?

If the latter, I think we'd want to complain loudly, so that users report problems they trigger to us.

src/memray/_memray/macho_shenanigans.cpp
```diff
@@ -0,0 +1,145 @@
...
+        return;
+    }
+
+    LOG_IF_DEBUG("Found " << dyninfo_table.segments.size() << " data segments");
+
+    for (const auto& segment_cmd : dyninfo_table.segments) {
+        const auto current_seg = reinterpret_cast<uintptr_t>(segment_cmd);
+        auto section_head = reinterpret_cast<const section_t*>(current_seg + sizeof(segment_command_t));
+        LOG_IF_DEBUG("Considering segment " << segment_cmd->segname);
```
godlygeek, line 105: Any risk of getting a `segname` that's not null-terminated, or anything like that?

src/memray/_memray/macho_shenanigans.cpp
```diff
@@ -0,0 +1,145 @@
...
+{
+    patch_symbols_in_shared_object(header, slide, true);
+}
+
+/* Public API functions */
+
+void
+SymbolPatcher::overwrite_symbols() noexcept
+{
+    _dyld_register_func_for_add_image(_patch_symbols_in_shared_objects);
```
godlygeek, line 136: This seems a bit weird - there's a register, but no unregister. Is it possible to register more than one function, or does calling register a second time unregister the first thing and register the second?

src/memray/_memray/hooks.h
```diff
@@ -4,16 +4,21 @@
...
-#include <sys/mman.h>
-#include <sys/prctl.h>
+#ifdef __linux__
+#    include "elf_utils.h"
+#    include <malloc.h>
+#    include <sys/prctl.h>
+#endif
+
+#ifdef __APPLE__
+#endif
```
godlygeek, line 21: Unnecessary

src/memray/_memray/hooks.h
```diff
@@ -26,22 +31,39 @@
...
-    FOR_EACH_HOOKED_FUNCTION(valloc)                                                                    \
-    FOR_EACH_HOOKED_FUNCTION(dlopen)                                                                    \
-    FOR_EACH_HOOKED_FUNCTION(dlclose)                                                                   \
-    FOR_EACH_HOOKED_FUNCTION(mmap)                                                                      \
-    FOR_EACH_HOOKED_FUNCTION(munmap)                                                                    \
-    FOR_EACH_HOOKED_FUNCTION(prctl)                                                                     \
-    FOR_EACH_HOOKED_FUNCTION(PyGILState_Ensure)                                                         \
-    MEMRAY_PLATFORM_HOOKED_FUNCTIONS
+#if defined(__linux__)
+#    define MEMRAY_HOOKED_FUNCTIONS                                                                     \
```
godlygeek, line 35: The platform-specific differences are supposed to be captured by `MEMRAY_PLATFORM_HOOKED_FUNCTIONS` above, rather than making `MEMRAY_HOOKED_FUNCTIONS` platform specific and listing all the common ones twice. We used to only have 2 cases for `MEMRAY_PLATFORM_HOOKED_FUNCTIONS` (glibc vs musl), now we're adding a 3rd (Apple).

Looks like it should be:
```cpp
#if defined(__APPLE__)
#    define MEMRAY_PLATFORM_HOOKED_FUNCTIONS
#elif defined(__GLIBC__)
#    define MEMRAY_PLATFORM_HOOKED_FUNCTIONS                                                            \
        FOR_EACH_HOOKED_FUNCTION(memalign)                                                              \
        FOR_EACH_HOOKED_FUNCTION(prctl)                                                                 \
        FOR_EACH_HOOKED_FUNCTION(pvalloc)                                                               \
        FOR_EACH_HOOKED_FUNCTION(mmap64)
#else
#    define MEMRAY_PLATFORM_HOOKED_FUNCTIONS                                                            \
        FOR_EACH_HOOKED_FUNCTION(memalign)                                                              \
        FOR_EACH_HOOKED_FUNCTION(prctl)
#endif
```

If we want to split "platform" into 2 separate dimensions (OS vs libc) we can do that as well.

src/memray/_memray/hooks.h
```diff
@@ -4,16 +4,21 @@
...
 
 #include <cstdlib>
 #include <iostream>
-#include <malloc.h>
 
-#include "elf_utils.h"
-#include <dlfcn.h>
-#include <sys/mman.h>
-#include <sys/prctl.h>
+#ifdef __linux__
```
godlygeek, line 14: Hm - can we encapsulate the Linux version all into a separate header that's conditionally included, rather than littering `#ifdef __linux` all over this file (can be saved for a later cleanup PR, if you prefer)

src/memray/_memray/hooks.h
```diff
@@ -70,6 +94,7 @@ struct SymbolHook
 
     void ensureValidOriginalSymbol()
     {
+#if defined(__linux__)
```
godlygeek, line 97: So on Linux, we assume that our original function pointer can be wrong, and we attempt to correct it by resolving the symbol ourselves. Remind me why we need that on Linux, and why we don't on Mac?

src/memray/_memray/hooks.cpp
```diff
@@ -162,34 +164,6 @@ pymalloc_free(void* ctx, void* ptr) noexcept
     }
 }
 
-void*
-mmap(void* addr, size_t length, int prot, int flags, int fd, off_t offset) noexcept
```
godlygeek, line 166: Some of the changes in this file appear unnecessary. Looks like you moved the `mmap` and `mmap64` and `munmap` functions around in the file, but they're not guarded by any additional macros or anything... This makes the diff harder to review than necessary

godlygeek (changes requested):

src/memray/_memray/macho_utils.h
```diff
@@ -0,0 +1,117 @@
...
+typedef struct mach_header mach_header_t;
+typedef struct segment_command segment_command_t;
+typedef struct section section_t;
+typedef struct nlist nlist_t;
+#    define ARCH_LC_SEGMENT LC_SEGMENT
+#endif
+
+#ifndef SEG_DATA_CONST
+#    define SEG_DATA_CONST "__DATA_CONST"
+#endif
```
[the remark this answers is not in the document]
pablogsal: That's it, I think MacOS >=11.0 define it and previous versions doesn't

src/memray/_memray/macho_shenanigans.cpp
pablogsal: > and I'm surprised to see ".cpython" in the name of the "memray" one
That's to avoid matching something that just has "memray" on the full path (like a `memray` folder).

> How stable are these names? It looks like you're matching "dyld" 

They are stable across different versions for what I have seen (the CI checks it). This basically avoids patching the linker. One is the linker itself (`/dyld`) and the other is its runtime library (`dyld.dylib`). I didn't want to just add `dyld` to avoid unwanted substring matching

src/memray/_memray/hooks.h
pablogsal: >  Remind me why we need that on Linux, and why we don't on Mac?

Because this only happens on the weird Debian systems where they build with some odd options that cause the GNU linker to give us the PLT address for just "&malloc". This doesn't happen in Mac because dyld doesn't not get confused by this (I checked).

src/memray/_memray/logging.h
```diff
@@ -21,6 +21,12 @@ logToStderr(const std::string& message, int level);
 void
 setLogThreshold(int threshold);
 
+logLevel
+getLogThreshold();
+
+#define LOG_IF_DEBUG(msg)                                                                               \
+    if (getLogThreshold() == DEBUG) LOG(DEBUG) << msg
```
godlygeek, line 28: This should be `>=` technically, though it probably doesn't matter much.

Why not just make `Log::operator<<` handle this? Couldn't we just make it a no-op if the log level is less than the log threshold?

src/memray/_memray/hooks.h
pablogsal: Let's do this in a cleanup PR if you don't mind, just to avoid having to restructure a lot and go into merge conflicts :S

src/memray/_memray/macho_shenanigans.cpp
godlygeek: Oh, for some reason I was misreading this as though the part after the `.` was some sort of namespace, but of course it's not, and that's just part of our filename. Nevermind this, you're right, this should be stable if it's just based on file names.

src/memray/_memray/logging.h
pablogsal: Because that still would create the string in the caller, no?

src/memray/_memray/hooks.h
godlygeek: Mm. For what it's worth, we may also need `ensureValidOriginalSymbol` (or something like it) if we ever want to patch a symbol that isn't available from our DSO dependencies. We've suggested patching things like `__tls_get_addr` or `PyInit_foomodule` in the past, and for those things, we can't just take `&` and use that as the original address, we need to look it up dynamically.

Something to keep in mind, at least, though we're currently not doing that anywhere.

src/memray/_memray/macho_shenanigans.cpp
pablogsal: Apparently it adds it a some kind of list:

https://github.com/apple-opensource/dyld/blob/1128192c016372ae94793d88530bc5978c1fce93/dyld3/AllImages.cpp#L1321-L1339

src/memray/_memray/macho_shenanigans.cpp
pablogsal: > s this a condition that we expect to happen with some dynamic libraries

Yeah, some system libraries do not have one of the 3 tables. This means that these libraries don't have PLT and we cannot patch them.

src/memray/_memray/sink.cpp
```diff
@@ -24,6 +24,19 @@ using namespace memray::exception;
...
+#ifdef __APPLE__
+static int
+posix_fallocate(int fd, off_t offset, off_t len)
+{
+    fstore_t store = {F_ALLOCATEALL, F_PEOFPOSMODE, 0, len, 0};
+    int res = fcntl(fd, F_PREALLOCATE, &store);
+    if (res != 0) {
+        return -1;
+    }
+    return ftruncate(fd, offset + len);
```
godlygeek, line 36: The real `posix_fallocate` returns an error number, instead of setting errno, and our code that calls it checks that. I think this needs to be:
```cpp
    if (res != 0) {
        return errno;
    }
    res = ftruncate(fd, offset + len);
    if (res != 0) {
        return errno;
    }
    return 0;
```

But also, we probably need to retry on `EINTR` here. Our code that's calling `posix_fallocate` already loops if it gets `EINTR` returned to it, but that's not enough here - if `ftruncate` is repeatedly interrupted but `fcntl` succeeds on every iteration, we'll wind up preallocating far more space than we need.

src/memray/_memray/macho_shenanigans.cpp
pablogsal: No, we do the same for linux IIRC. There is nothing else we can do if the page cannot be unprotected.

src/memray/_memray/macho_shenanigans.cpp
pablogsal: No, is supposed to be a 16-character max null-terminated string always.

src/memray/_memray/hooks.cpp
pablogsal: Sorry, splitting this into commits was not easy :( I can try to restore the functions if you want

src/memray/_memray.pyx
```diff
@@ -493,7 +492,8 @@ cdef class FileReader:
         except OSError as exc:
             raise OSError(f"Could not open file {file_name}: {exc.strerror}") from None
 
-        self._path = "/proc/self/fd/" + str(self._file.fileno())
+        # self._path = "/proc/self/fd/" + str(self._file.fileno())
```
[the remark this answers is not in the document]
godlygeek: We open the file twice. This is an attempt to make sure that we get the same file both times, by opening it once and then re-opening a file handle to it later. We can try refactoring around this - ultimately the problem is that the `source` code is expecting a string for the file to open, and we want to pass it an already-opened FD, so that when we do our two passes over the file, they're both guaranteed to be over the _same_ file.

OTOH that was just me being paranoid, and perhaps that's not even a scenario we should bother worrying about - if the user replaces the file we're reading in the middle of a call to a reporter, maybe weird behavior is acceptable.

tests/integration/test_main.py
```diff
@@ -77,6 +77,9 @@ def track_and_wait(output_dir, sleep_after=100):
 
 
 def _wait_until_process_blocks(pid: int) -> None:
+    if "linux" not in sys.platform:
```
godlygeek, line 80: The commit message for this commit has a typo ("testig" -> "testing")

tests/integration/test_native_tracking.py
```diff
@@ -16,6 +16,9 @@
 TEST_MULTITHREADED_EXTENSION = HERE / "multithreaded_extension"
 TEST_NATIVE_EXTENSION = HERE / "native_extension"
 
+if "linux" not in sys.platform:
+    pytest.skip("tests only works on Linux", allow_module_level=True)
```
godlygeek, line 20: nit: The message should either be "tests that only work on Linux" or "tests only working on Linux"

tests/integration/test_processes.py
```diff
@@ -13,6 +14,14 @@
 from tests.utils import filter_relevant_pymalloc_allocations
 
 
+@pytest.fixture(scope="module", autouse=True)
+def set_multiprocessing_to_fork():
+    current_method = multiprocessing.get_start_method()
+    multiprocessing.set_start_method("fork", force=True)
+    yield
+    multiprocessing.set_start_method(current_method, force=True)
```
godlygeek, line 22: `force=True` is undocumented - any risk to this breaking, or going away in the future?

setup.py
```diff
@@ -165,8 +168,9 @@ def build_js_files(self):
         "c_string_type": "unicode",
         "c_string_encoding": "utf8",
     }
-    EXTRA_COMPILE_ARGS = ["-D_GLIBCXX_DEBUG", "-D_LIBCPP_DEBUG"]
-    UNDEF_MACROS = ["NDEBUG"]
+    if IS_LINUX:
+        EXTRA_COMPILE_ARGS = ["-D_GLIBCXX_DEBUG", "-D_LIBCPP_DEBUG"]
+        UNDEF_MACROS = ["NDEBUG"]
```
godlygeek, line 173: We probably want to unset `NDEBUG` even on Darwin, I would think?

setup.py
```diff
@@ -213,20 +220,38 @@ def build_js_files(self):
...
+    language="c++",
+    extra_compile_args=["-std=c++17", "-Wall", *EXTRA_COMPILE_ARGS],
+    extra_link_args=["-std=c++17", *EXTRA_LINK_ARGS],
+    define_macros=DEFINE_MACROS,
+    undef_macros=UNDEF_MACROS,
+)
+
+
+if not (IS_LINUX or IS_MAC):
+    raise RuntimeError(f"memray does not supports this platform ({platform})")
```
godlygeek, line 254: `s/supports/support`

.github/workflows/build.yml
```diff
@@ -59,6 +59,8 @@ jobs:
         python3 -m pip install tox tox-gh-actions
     - name: Run tox
       run: tox . -vvv
+    - name: Setup upterm session
+      uses: lhotari/action-upterm@v1
```
godlygeek, line 63: Let's not forget to remove this before we land the PR.

src/memray/_memray/hooks.h
pablogsal: Done

.github/workflows/build_wheels.yml
```diff
@@ -17,7 +17,7 @@ concurrency:
 
 jobs:
   build_wheels:
-    name: Wheel for ${{ matrix.cibw_python }}-${{ matrix.cibw_arch }}
+    name: Wheel for Linux-${{ matrix.cibw_python }}-${{ matrix.cibw_arch }}
```
godlygeek, line 20: nit: Extra whitespace at EOL

.github/workflows/build_wheels.yml
```diff
@@ -37,6 +37,31 @@ jobs:
...
+          path: ./wheelhouse/*.whl
+
+  build_wheels_macos:
+    name: Wheel for MacOS-${{ matrix.cibw_python }}-${{ matrix.cibw_arch }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-11]
+        cibw_python: ["cp38-*", "cp39-*", "cp310-*", "cp311-*"]
```
godlygeek, line 51: Why no 3.7 wheel for OS X, when we have one for Linux?

src/memray/_memray/hooks.h
pablogsal: Unfortunately adding support for that is quite a lot of extra work, because of how macOS does it symbol exposure, so I would prefer to not do it for now.

src/memray/_memray/macho_shenanigans.cpp
```diff
@@ -70,14 +70,21 @@ patch_symbols_in_section(
...
+        const char* image_name,
+        bool restore_original,
+        std::set<std::string>& patched)
 {
-    Dl_info info;
-    if (dladdr(header, &info) == 0) {
-        return;
+    if (!restore_original) {
+        if (patched.find(image_name) != patched.end()) {
+            // We already patched this libraru
```
godlygeek, line 82: typo, "libraru" -> "library"

tests/integration/test_processes.py
pablogsal: > force=True is undocumented - any risk to this breaking, or going away in the future?

Very low risk. If it breaks, we will set it globally in the test suite. This is just a cleaner way.

setup.py
pablogsal: 👍

.github/workflows/build_wheels.yml
pablogsal: They don't work for macOS:

https://cibuildwheel.readthedocs.io/en/stable/

src/memray/_memray/sink.cpp
pablogsal: Done in a fixup

src/memray/_memray/logging.h
godlygeek: No, the signature is:
```cpp
    template<typename T>
    LOG& operator<<(const T& msg)
```
and we're calling it like:
```cpp
LOG_IF_DEBUG(symname << " intercepted!");
```

So `operator<<` is being called twice, each time with a `const char*`. The only place where a dynamic memory allocation is occurring is when `operator<<` does `buffer << msg` because that's appending to a stringstream, which may need to grow the string. If we made `operator<<` return instead of calling `buffer << msg`, no allocation would happen, and we wouldn't need `LOG_IF_DEBUG` instead of our normal `LOG_DEBUG`.

src/memray/_memray/macho_shenanigans.cpp
godlygeek: We switched to iterating over the loaded libraries.

src/memray/_memray/hooks.cpp
godlygeek: No, as long as you didn't make any changes to the implementation of the functions, I'm happy. If you did make changes to the implementations, I didn't notice them through the diff noise.

pablogsal: I did not, so I am leaving as this. Apologies for the noise.

src/memray/_memray/logging.h
pablogsal: Ok, I will implement this in the `operator<<`

src/memray/_memray/macho_shenanigans.cpp
pablogsal: Ok, done in the last rebase.

src/memray/_memray.pyx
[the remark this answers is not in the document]
pablogsal: Ok, I have preserved the Linux behaviour and leave it like `self._path = str(file_name)` for macos.

docs/index.rst
```diff
@@ -28,7 +28,7 @@ Memray can help with the following problems:
    :align: center
 
 .. note::
-    Memray only works on Linux and cannot be installed on other platforms.
+    Memray only works on Linux and MacOS and cannot be installed on other platforms.
```
pradyunsg, line 31:
```suggestion
    Memray only works on Linux and MacOS, and cannot be installed on other platforms.
```

godlygeek (approved):
LGTM

PR #175, opened 2022-08-11
Simplify the CI

pablogsal:
We are testing multiple Python versions twice: once when building wheels
and another in a target in the main CI. The only thing this target
can give us more than the tests that are being executed as part of the
wheels is to run the test suite under "pycoverage".

To simplify the setup, just run coverage under the latest Python version
in the main CI, as there is not (currently) anything version-specific.


godlygeek (approved):
LGTM

PR #176, opened 2022-08-11
Handle our profile function getting uninstalled

godlygeek:
Previously, we would retain references to frames that were on the stack
in a given thread, but without our profile function we wouldn't know
when those frames were popped and it was no longer safe to reference
them.

After this change, we are able to detect when our profile function is
deregistered (either because a new one has been installed or because the
thread state the profile function was registered with is dying). This
gives us a hook to invalidate our cached frames to prevent this
use-after-free bug.

### DIFF
--- a/src/memray/_memray/tracking_api.cpp
+++ b/src/memray/_memray/tracking_api.cpp
@@ -110,6 +110,8 @@ class PythonStackTracker
     static void installProfileHooks();
     static void removeProfileHooks();
 
+    void clear();
+
     static PythonStackTracker& get();
     void emitPendingPops();
     void emitPendingPushes();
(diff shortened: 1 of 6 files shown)


src/memray/_memray/tracking_api.cpp
```diff
@@ -913,6 +890,35 @@ Tracker::unregisterPymallocHooks() const noexcept
 
 // Trace Function interface
 
+PyObject*
+create_profile_arg()
+{
+    // Borrowed reference
+    PyObject* memray_ext = PyDict_GetItemString(PyImport_GetModuleDict(), "memray._memray");
```
pablogsal, line 897: `PyImport_GetModuleDict` cannot return `NULL` IIRC but it would be good to add an assert that's not NULL just in case

pablogsal (approved):
LGTM

src/memray/_memray/tracking_api.cpp
godlygeek: It cannot, it raises a fatal error in that case instead: https://github.com/python/cpython/blob/6fd4c8ec7740523bb81191c013118d9d6959bc9d/Python/import.c#L253-L261

I believe that case could happen if something tries to spawn a new thread late in finalization while Memray tracking is still somehow enabled - but a fatal error from the interpreter is a nice way to handle it, anyway.

pablogsal (approved):

PR #177, opened 2022-08-11
Complete MacOS support with some warnings and NEWS entry

pablogsal:
- Add NEWS entry for MacOS support
- Show a warning message when creating Tracker objects on MacOS

### DIFF
--- a/src/memray/_memray.pyx
+++ b/src/memray/_memray.pyx
@@ -9,6 +9,7 @@ cimport cython
 import threading
 from datetime import datetime
 
+from rich import print as pprint
 from rich.progress import Progress
 from rich.progress import SpinnerColumn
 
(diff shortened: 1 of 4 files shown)


src/memray/_memray.pyx
```diff
@@ -302,6 +303,17 @@ cdef class Tracker:
         else:
             raise TypeError("destination must be a SocketDestination or FileDestination")
 
+    def __init__(self, object file_name=None, *, object destination=None,
+                  bool native_traces=False, unsigned int memory_interval_ms = 10,
+                  bool follow_fork=False, bool trace_python_allocators=False):
+        IF UNAME_SYSNAME == "Darwin":
```
godlygeek, line 309: Do we really need to use conditional compilation for this? Seems overly complicated, I'd rather just a regular `if platform.system() ...` check...

Also, why can't we do this in `__cinit__`?  It's not a big deal to have a `__init__` as well, but it'll be slightly annoying to update the args in both spots.

pablogsal: > Also, why can't we do this in `__cinit__`? It's not a big deal to have a `__init__` as well, but it'll be slightly annoying to update the args in both spots.

I prefer to leave the `__cinit__` for purely C-level initialization and do any python-only stuff in __init__. 

> Do we really need to use conditional compilation for this? Seems overly complicated, I'd rather just a regular if platform.system() ... check...

Is literally almost the same code and we skip one `platform.system()` call when creating trackers in Linux :shrug:  If you feel strongly, I can make it a regular conditional

godlygeek: OK, I don't have super strong feelings about either of these things, I can live with both.

godlygeek (approved):
LGTM

lkollar (approved):

pradyunsg (approved):

Issue #166, opened 2022-07-28
Segfault with Sqlalchemy


auxsvr:
## Bug Report

Memray  triggers segfault running Sqlalchemy code using greenlets.
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  PyCode_Addr2Line (addrq=0, co=0x66) at Objects/codeobject.c:1248
1248        if (addrq < 0) {
[Current thread is 1 (Thread 0x7fef9e344200 (LWP 23383))]
#0  PyCode_Addr2Line (addrq=0, co=0x66) at Objects/codeobject.c:1248
#1  PyFrame_GetLineNumber (f=<unknown at remote 0x5594cd476a00>) at Objects/frameobject.c:49
#2  0x00007fef9dc2a287 in memray::tracking_api::PythonStackTracker::pushPythonFrame(_frame*) [clone .constprop.52] ()
  from /home/petros/.local/lib/python3.10/site-packages/memray/_memray.cpython-310-x86_64-linux-gnu.so
#3  0x00007fef9dc2b5b8 in memray::tracking_api::PyTraceFunction(_object*, _frame*, int, _object*) ()
  from /home/petros/.local/lib/python3.10/site-packages/memray/_memray.cpython-310-x86_64-linux-gnu.so
#4  0x00007fef9e870e55 in call_trace (func=0x7fef9dc2b4c0 <memray::tracking_api::PyTraceFunction(_object*, _frame*, int,  _object*)>, obj=123, tstate=0x5594cc6614e0, 
[...]

PR #185, opened 2022-08-17
Detect the greenlet module and try to handle it

godlygeek:
Attempt to detect when the `greenlet` module is used (whether it has
already been imported at the point when tracking starts, or it gets
imported after tracking has started). If we discover that the `greenlet`
module is in use, register a callback to discover greenlet switch events
and react to them.  Unfortunately, this callback needs to be registered
once per thread, and the only place where we can reasonably register it
in a background thread is when our profile function is called, so we may
miss greenlet switches if they occur before any Python calls.

If we detect Greenlet, print a warning that this support is experimental.

Closes: #166

### DIFF
--- a/src/memray/_memray/tracking_api.cpp
+++ b/src/memray/_memray/tracking_api.cpp
@@ -358,6 +358,18 @@ PythonStackTracker::installGreenletTraceFunctionIfNeeded()
 
     // Note: guarded by the GIL
     d_greenlet_hooks_installed = true;
+
+    static bool warned = false;
+    if (!warned) {
+        warned = true;
+
+        PyObject* ret = PyObject_CallMethod(_memray, "print_greenlet_warning", nullptr);
+        Py_XDECREF(ret);
+        if (!ret) {
+            PyErr_Print();
+            _exit(1);
+        }
+    }
 }
 
 void
--- a/src/memray/_memray/tracking_api.cpp
+++ b/src/memray/_memray/tracking_api.cpp
@@ -12,6 +12,7 @@
 #    include <mach/task.h>
 #endif
 
+#include <algorithm>
 #include <mutex>
 #include <type_traits>
 #include <unistd.h>
@@ -94,6 +95,7 @@ class PythonStackTracker
     };
 
   public:
+    static bool s_greenlet_tracking_enabled;
     static bool s_native_tracking_enabled;
 
     static void installProfileHooks();
@@ -106,6 +108,9 @@ class PythonStackTracker
     void invalidateMostRecentFrameLineNumber();
     int pushPythonFrame(PyFrameObject* frame);
     void popPythonFrame(PyFrameObject* frame);
+
+    void installGreenletTraceFunctionIfNeeded();
+    void handleGreenletSwitch(PyObject* from, PyObject* to);
 
   private:
     // Fetch the thread-local stack tracker without checking if its stack needs to be reloaded.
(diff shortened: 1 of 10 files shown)


src/memray/_memray/tracking_api.cpp
```diff
@@ -300,6 +310,97 @@ PythonStackTracker::popPythonFrame(PyFrameObject* frame)
...
+    PyObject* ret = PyObject_CallMethod(
+            _greenlet,
+            "settrace",
+            "N",
+            PyObject_GetAttrString(_memray, "greenlet_trace_function"));
+    Py_XDECREF(ret);
+
+    if (!ret) {
+        // This might be hit from PyGILState_Ensure when a new thread state is
+        // created on a C thread, so we can't reasonably raise the exception.
```
pablogsal, line 353: I mean, we *could* raise an exception because at that time we have the GIL, no? Is just that the exception will be ignored

src/memray/_memray/tracking_api.cpp
```diff
@@ -300,6 +310,97 @@ PythonStackTracker::popPythonFrame(PyFrameObject* frame)
...
+}
+
+void
+PythonStackTracker::handleGreenletSwitch(PyObject*, PyObject*)
+{
+    RecursionGuard guard;
+
+    // Clear any old TLS stack, emitting pops for frames that had been pushed.
+    if (d_stack) {
+        for (auto& frame : *d_stack) {
```
pablogsal, line 381: Nit, you can use `std::count_if(d_stack->begin(), d_stack->end(), [](const auto& frame){return frame.state != FrameState::NOT_EMITTED;});`

src/memray/_memray/tracking_api.cpp
```diff
@@ -300,6 +310,97 @@ PythonStackTracker::popPythonFrame(PyFrameObject* frame)
...
+            if (frame.state != FrameState::NOT_EMITTED) {
+                ++d_num_pending_pops;
+            }
+        }
+        d_stack->clear();
+        emitPendingPushesAndPops();
+    }
+
+    // Re-create our TLS stack from our Python frames, most recent last.
+    PyFrameObject* frame = PyEval_GetFrame();
```
pablogsal, line 391: This is correct because if this is NULL for whatever reason we don't do anything, but it may be good to add an assert or early return to make it clearer.

src/memray/_memray/tracking_api.cpp
```diff
@@ -300,6 +310,97 @@ PythonStackTracker::popPythonFrame(PyFrameObject* frame)
...
+
+    std::vector<PyFrameObject*> stack;
+    while (frame) {
+        stack.push_back(frame);
+        frame = compat::frameGetBack(frame);
+    }
+
+    for (auto frame_it = stack.rbegin(); frame_it != stack.rend(); ++frame_it) {
+        pushPythonFrame(*frame_it);
+    }
```
pablogsal, line 401:
```suggestion
    std::transform(stack.rbegin(), bar.rend(), [](auto& frame){pushPythonFrame(frame);});
```

src/memray/_memray/tracking_api.cpp
godlygeek: The exception will be ignored, because `PyGILState_Ensure` isn't expected to return with the exception indicator set, and so the caller won't check for it. The caller will go on to make Python C API calls with the exception indicator already set, in violation of the C API's contract.

src/memray/_memray/tracking_api.cpp
godlygeek: It's written to allow NULL, as you say. I'd rather add a comment than more code, I'll do that.

src/memray/_memray/tracking_api.cpp
godlygeek: Since `pushPythonFrame` is a member function, the lambda needs to capture `this`. But OK.

godlygeek: Hah, oops: and we want `for_each`, not `transform`.

pablogsal (approved):

PR #186, opened 2022-08-18
docs: Document supported environments

godlygeek:
Also document any known issues that can lead to incorrect, misleading,
or confusing stack traces.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -56,5 +56,6 @@ Memray can help with the following problems:
    :hidden:
    :caption: Project Information
 
+   supported_environments
    changelog
    licenses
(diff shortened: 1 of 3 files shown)


godlygeek:
@pablogsal feel free to force-push any edits. I'll take another pass at the grammar afterwards for a native speaker polish if you do.

You suggested a table, but I don't think it works well - we have too many dimensions (Python version, OS, libc, CPU architecture).

godlygeek:
Also, I put the new page under "Project information", which is a bit low down the side bar. We can move it if you think it'd be better higher up.

pablogsal (approved):

Issue #195, opened 2022-08-30
Segfault when trying to flush pending frames to disk in forked processes


pablogsal:
If we try to flush pending frames to disk in a forked process that has been created without `follow_fork=True` we will miserably fail when trying to get hold of the tracker and call `Tracker::getTracker()->popFrame` or any other member function. We are supposed to protect against this here:

https://github.com/bloomberg/memray/blob/acd738eeb353c4a38988b77909710545c87263b6/src/memray/_memray/tracking_api.cpp#L1086-L1088

but unfortunately, we don't deactivate the tracker when forking without `follow_fork=True`:

https://github.com/bloomberg/memray/blob/acd738eeb353c4a38988b77909710545c87263b6/src/memray/_memray/tracking_api.cpp#L713-L722

As a consequence, we enter https://github.com/bloomberg/memray/blob/acd738eeb353c4a38988b77909710545c87263b6/src/memray/_memray/tracking_api.cpp#L513 and hell goes loose. This is enough to reproduce in any system:

```python
import multiprocessing
import sys
import memray

multiprocessing.set_start_method("fork")

def bar(n):
    sys.setprofile(None)

def foo():
    with multiprocessing.Pool() as pool:
        result = pool.map_async(bar, [1024])
        result.get(timeout=1)

def main():
    with memray.Tracker("/dev/null") as t:
        foo()

if __name__ == "__main__":
    main()
```

PR #196, opened 2022-08-30
Ensure the Tracker is deactivated after a fork without "follow_fork"

pablogsal:
If we try to flush pending frames to disk in a forked process that has
been created without `follow_fork=True` we will miserably fail when
trying to get hold of the tracker and call
`Tracker::getTracker()->popFrame` or any other member function. We are
supposed to protect against this in the `forget_python_stack` function
but that's down by checking if the Tracker is deactivated or not by
calling `Tracker::isActive()`. Unfortunately, in `Tracker::childFork()`
we are missing a call to `Tracker::deactivate` if ``follow_fork`` has
not been passed.

As a consequence, we enter `emitPendingPushesAndPops` and hell goes
loose.

Closes: #195

### DIFF
--- a/src/memray/_memray/tracking_api.cpp
+++ b/src/memray/_memray/tracking_api.cpp
@@ -716,6 +716,7 @@ Tracker::childFork()
         // Note that the old tracker's hooks may still be installed. This is
         // OK, as long as they always check the (static) isActive() flag before
         // calling any methods on the now null tracker singleton.
+        Tracker::deactivate();
         d_instance = nullptr;
         RecursionGuard::isActive = false;
         return;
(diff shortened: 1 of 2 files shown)


src/memray/_memray/tracking_api.cpp
```diff
@@ -306,6 +306,7 @@ PythonStackTracker::popPythonFrame(PyFrameObject* frame)
     if (d_stack->back().state != FrameState::NOT_EMITTED) {
         d_num_pending_pops += 1;
         assert(d_num_pending_pops != 0);  // Ensure we didn't overflow.
+    } else {
```
pradyunsg, line 309: This is a no-op, and can be dropped?

pablogsal: Yeah, this should not have been there

lkollar (approved):

PR #200, opened 2022-09-02
Add a new `transform` subcomand

pablogsal:
Add a new `transform` subcomand that allows to translate memray result files
into output files compatible with other tools. As a first output format
we add also a `gprof2dot` format that allows to produce graph-like
reports when combined with `graphviz`.

**Describe your changes**

**Testing performed**

**Additional context**


godlygeek:
"convert" or "transform" seem like better names for this than "dump"

pradyunsg:
`export` is another option.

tests/unit/test_transform_reporter.py
```diff
@@ -0,0 +1,180 @@
...
+        # WHEN
+        reporter.render_as_gprof2dot(output)
+        output.seek(0)
+
+        # THEN
+        output_data = json.loads(output.read())
+        assert output_data == {
+            "costs": [{"description": "Memory", "unit": "bytes"}],
+            "events": [],
+            "functions": [],
```
godlygeek, line 178: Hm - we don't want to report this case as `<unknown>` like we do elsewhere?

src/memray/commands/transform.py
```diff
@@ -0,0 +1,94 @@
...
+        assert self.output_file is not None
+        try:
+            import gprof2dot  # noqa
+        except ImportError:
+            pprint(
+                ":exclamation: [red]gprof2dot doesn't seem to be installed. "
+                "Please install it to be able to process the transform.[/red]"
+                "\n:point_right: Check out https://github.com/jrfonseca/gprof2dot for "
+                "installation instructions"
+            )
```
godlygeek, line 80: This assumes that `gprof2dot` is gonna be pip installed into the same virtualenv as Memray is running in, but it can also be installed from a distro package manager, in which case this results in a false positive. Maybe we should also try running `shutil.which("gprof2dot")`? Then if `which("gprof2dot")` works, we'd use just `gprof2dot -f json` instead of `python -m gprof2dot -f json` in the the message below.

This isn't just theoretical, when I tried running this before reading the code, I got this warning message, and then I ran `apt-get install gprof2dot`, and I was confused when that didn't make the warning go away.

godlygeek (commented):
A few comments, plus I made some fixup commits for some things.

tests/unit/test_transform_reporter.py
pablogsal: Hummmm, that's an interesting point. Let me see what I can do.

godlygeek: Hm. OK.

godlygeek (approved):
LGTM

PR #201, opened 2022-09-03
Add an option to get temporary allocations

pablogsal:
Temporary allocations are defined as allocations that happen
immediately followed by their deallocation. Tracking these is useful
because they can point to potential optimizations. For instance, these
can appear if someone pushes to a big vector without preallocating
first, leading to a lot of reallocations instead of having a big
allocation at the start.

We add a new `--temporary-allocations` option to the flame-graph, table
and tree reporters that allow to show the temporary allocations instead
of the high watermark ones.


src/memray/commands/summary.py
```diff
@@ -26,6 +27,30 @@ def prepare_parser(self, parser: argparse.ArgumentParser) -> None:
...
+            high watermark allocations.  An allocation is considered temporary
+            if at most N other allocations occur before it is deallocated.  With
+            N=0, an allocation is temporary only if it is immediately
+            deallocated before any other allocation occurs.
+            """
+            ),
+            action="store",
+            dest="temporary_allocations",
+            type=int,
+            default=0,
```
pablogsal, line 45: @godlygeek I recalled you wanted this to be n-1 so the normal version will run with --temporary-allocation-threshold=1, no? What value should we use to flag that's not activated?

godlygeek: > What value should we use to flag that's not activated?

I'll go with -1 for now, but if you don't like that we should be able to use `None` instead.

godlygeek:
Well, I asked you to change it so that we could pass the threshold as a separate argument `--temporary-allocation-threshold=N`, which seemed like a good idea since `--temporary-allocations[=N]` seemed pretty unclear about what the `N` means.

But, I didn't realize that we'd wind up with `--temporary-allocations` and `--temporary-allocation-threshold=N` being mutually exclusive - it's pretty terrible to not be able to pass
```
--temporary-allocations --temporary-allocation-threshold=2
```
since the latter seems like an option impacting the mode that the former put us in, and it's weird to not be able to explicitly turn on the mode and also provide the option affecting that mode.

And we can't just make them not mutually exclusive, since we need them to both be mutually exclusive with `--leaks`, so that both `--leaks --temporary-allocations` and `--leaks --temporary-allocation-threshold=2` fail, since either of them can be used to turn on the mode.

Ugh. I think I dislike this more than I disliked the `--temporary-allocations` parameter that could either be called with or without an argument. That version was unclear and surprising in one way, but this version is unclear and surprising in a different way that users are more likely to hit, and this version is more work and more code.

Sorry for the wasted work, but I'm gonna put it back to how you originally had it. My fault for not seeing the consequences of my request...

src/memray/_memray/snapshot.cpp
```diff
@@ -122,6 +122,77 @@ SnapshotAllocationAggregator::getSnapshotAllocations(bool merge_threads)
...
+            case hooks::AllocatorKind::RANGED_DEALLOCATOR: {
+                if (it == d_current_allocations.end()) {
+                    break;
+                }
+
+                auto alloc_it = std::find_if(
+                        it->second.begin(),
+                        it->second.end(),
+                        [&](auto& current_allocation) {
+                            bool match = (current_allocation.address == allocation.address);
```
godlygeek, line 157: I'm not sure there's anything we can reasonably do about it, but this could have false positives if pointers are passed between threads and freed by some other thread, due to pointers being pooled and reused.

Imagine pointer 0x1234 gets allocated in thread A, and passed to thread B on a `queue.Queue`. Thread B does some processing of that work item and then destroys it. Thread B then creates a new object and gets pointer 0x1234 again, reusing it for the new object. Thread B passes that object back to thread A as a response. Thread A logs the response and destroys it.

We'll think that deallocation of 0x1234 is for the first object, which was allocated on thread A, and we'll incorrectly decide that the first allocation was temporary.

I'm not sure there's any good solution for that, though, without ignoring thread IDs entirely and mashing all allocations from all threads together.

godlygeek:
> Sorry for the wasted work, but I'm gonna put it back to how you originally had it.

No, I did that and it's even worse. That allows
```
memray summary test.bin --temporary-allocations
```
but does not allow
```
memray summary --temporary-allocations test.bin 
```
because it thinks that `test.bin` is the optional argument for `--temporary-allocations`. If we only use one option for both things, it can't come right before the first non-option argument, thanks to the `nargs="?"`, which will surely confuse users - it will be the only option for any of our commands that can't be placed in any position on the command line.

I think that's worse than not being able to supply both `--temporary-allocations` and `--temporary-allocation-threshold` on the same command line.

I'm going back to 2 arguments after all.

src/memray/_memray_test_utils.pyx
```diff
@@ -269,3 +269,12 @@ def allocate_cpp_vector(size_t size):
     v.reserve(nelems)
     return v.size()
 
+
+@cython.profile(True)
+def fill_cpp_vector(size_t size):
+    cdef vector[int] v;
+    cdef size_t nelems = <size_t>(size / sizeof(int))
```
godlygeek, line 276: Is there some reason for this `<size_t>` cast that I'm not seeing?

The tests pass even if it's removed, and it shouldn't be necessary because we're already dividing one `size_t` by another, which always returns a `size_t`.

tests/integration/test_tracking.py
```diff
@@ -1062,6 +1064,311 @@ def allocating_function(allocator, amount, stop_flag):
...
+        and free.
+
+        Note that there is not an easy way to synchronize this test because
+        using condition variables may allocate, disrupting the test as there
+        will be other spurious allocations between our call to posix_memalign
+        and the call to free.
+
+        The best effort is that both threads do a bunch of calls to
+        posix_memalign and free in a loop and hopefully some of them will be
+        interleaved.
```
godlygeek, line 1306: How about we use `time.sleep()` instead? At the cost of making the test take 750ms to run, we could make
```
Thread 1    Thread 2
------ time=0ms -------
allocate    sleep(0.25)
sleep(0.5)
------ time=250ms -----
            allocate
            sleep(0.5)
------ time=500ms -----
    free
------ time=750ms -----
            free
```

Sure, that's racy - but so is this approach, so I'm not sure it's any worse.

I bet we could enforce that order by spinlocking on a Python value, actually... What if we did:

```py
thread1_allocated = False
thread1_should_free = False

def thread1_body(allocator):
    global thread_1_allocated
    allocator.posix_memalign(amount)
    thread_1_allocated = True
    while not thread_1_should_free:
        pass
    allocator.free()

def thread2_body(allocator):
    global thread1_should_free
    while not thread1_allocated:
        pass
    allocator.alloc()
    allocator.free()
    thread1_should_free = True
```

I think that's not racy (thanks to the GIL) and I think no unexpected allocations can happen in there.

tests/integration/test_tracking.py
```diff
@@ -1200,3 +1508,48 @@ def test_memory_snapshots_tick_interval(self, tmp_path):
...
+
+        # THEN
+        reader = FileReader(output)
+        temporary_allocations = list(
+            reader.get_temporary_allocation_records(temporary_buffer_size=2)
+        )
+        assert elements == 512
+        assert len(temporary_allocations) == 1
+        (record,) = temporary_allocations
+        assert record.n_allocations == 10
```
godlygeek, line 1530: Hm. This is an implementation detail that depends on the initial size and may not be consistent across different C++ STL implementations, right?

godlygeek:
OK - I've pushed a bunch of fixups of all the (mostly small) things I noticed while reading through this. Please check over them and see if there's anything you disagree with.

tests/integration/test_tracking.py
pablogsal: Yeah, but most are going to match it. We can go with this and change if we encounter something failing

src/memray/_memray/snapshot.cpp
pablogsal: Yeah, I though about this but there isn't anything that we can do here. The other alternative is flatten all threads but that not only is more racy but will cause more problems than what you mention as many threads racing allocations will interleave the allocs/dealloc so we won't detect them correctly.

tests/integration/test_tracking.py
pablogsal: Ok, let's go with the second version spinlocking on a python value

src/memray/_memray_test_utils.pyx
pablogsal: I am getting this:

```
src/memray/_memray_test_utils.pyx:276:30: Cannot assign type 'double' to 'size_t'
```

godlygeek: Looks like that most likely should have been `//` instead of `/` so that we get integer division - but whatever, casting to double and then back to size_t isn't wrong, just a bit weird.

godlygeek (approved):
LGTM

PR #202, opened 2022-09-08
Build lz4 ourselves when building wheels

pablogsal:
- Revert "Don't build wheels on arm64 macOS for now"
- Build lz4 ourselves when building wheels

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -53,6 +53,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+      - name: Sets env vars for compilation
+        if: matrix.cibw_arch == 'arm64'
+        run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/build_wheels.yml
```diff
@@ -61,6 +66,7 @@ jobs:
           CIBW_PRERELEASE_PYTHONS: True
           CIBW_TEST_EXTRAS: test
           CIBW_TEST_COMMAND: pytest {package}/tests
+          CIBW_BUILD_VERBOSITY: 1
```
pablogsal, line 69: I'm leaving this on so we can quickly check that there are no linkage warnings in the amd64 wheels

godlygeek (changes requested):
2 minor nits:

godlygeek (approved):
LGTM

PR #206, opened 2022-09-13
Drop special support for tracing Cython functions

godlygeek:
Up until now, if the program being profiled included a Cython module built with profiling support enabled, those Cython functions would show up in our Python call stacks. This was rarely useful in practice, as most Cython libraries aren't distributed with profiling support enabled. Supporting this had significant maintenance costs that outweigh the benefits, so we are dropping that special support for Cython.


godlygeek:
This isn't really a _good_ way of filtering out the artificial Cython frames, but I haven't found a good way. I'd be very happy to switch to a cheaper method, if there is one.

pablogsal (approved):
LGTM

PR #209, opened 2022-09-14
Fix hybrid stacks on Python 3.11

### LINKED ISSUE #205: Incorrect hybrid stacks on Python 3.11
Because `_PyEval_EvalFrameDefault` can allocate memory in Python 3.11 before it calls our profile function, we can fail to correctly pair up the Python stack with the native stack.

Moderately horrific reproducer:
```py
import functools
import memray


def foo():
    list(map(bar, [None]))


def bar(_):
    x = 0
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    x += 1
    print(x)


with memray.Tracker(
    destination=memray.FileDestination("output.bin", overwrite=True),
    native_traces=True,
):
    functools.partial(foo)()

with memray.FileReader("output.bin") as f:
    for i, allocation in enumerate(f.get_high_watermark_allocation_records(), 1):
        print(f"\nAllocation {i}:")
        for frame in allocation.hybrid_stack_trace():
            print(frame[0])
```

The printed hybrid stack is:
```
_PyMem_RawMalloc
PyMem_RawMalloc
_PyObject_Malloc
_PyObject_Malloc
PyMem_Malloc
_PyCode_CreateLineArray
_PyCode_InitLineArray
call_trace
call_trace_protected
trace_function_entry
foo
_PyEval_EvalFrame
_PyEval_Vector
_PyFunction_Vectorcall
_PyObject_VectorcallTstate
map_next
list_extend
list___init___impl
list_vectorcall
_PyObject_VectorcallTstate
PyObject_Vectorcall
trace_call_function
<module>
```
which shows `map_next` calling into `foo`, rather than the other way around.

godlygeek:
We've assumed that memory can never be allocated by a thread in between the Python frame evaluation function, `_PyEval_EvalFrameDefault`, being called and it calling the function we've installed with `sys.setprofile` to let us know that a new Python frame has been entered.

Python 3.11 breaks this assumption, because `_PyEval_EvalFrameDefault` can now allocate memory before calling the trace function in order to create a table used to quickly find the line numbers corresponding to bytecode instructions in the called function.

Work around this change, and add a test that should begin to fail if our workaround ever stops working.

Closes #205


src/memray/_memray/record_writer.cpp
```diff
@@ -45,10 +45,19 @@ RecordWriter::RecordWriter(
...
             command_line,
             ::getpid(),
+            0,
+            0,
             getPythonAllocator()};
     strncpy(d_header.magic, MAGIC, sizeof(d_header.magic));
 }
 
+void
+RecordWriter::setMainTidSkip(thread_id_t main_tid, size_t skipped_frames_on_main_tid)
```
pablogsal, line 55: Sorry for the nit but i found the name a bit odd

```suggestion
RecordWriter::setMainTidAndSkipFrames(thread_id_t main_tid, size_t skipped_frames_on_main_tid)
```

src/memray/_memray/tracking_api.cpp
```diff
@@ -388,6 +388,25 @@ PythonStackTracker::handleGreenletSwitch(PyObject*, PyObject*)
         emitPendingPushesAndPops();
     }
 
+    // Save current TID on old greenlet. Print errors but otherwise ignore them.
+    PyObject* tid = PyLong_FromUnsignedLong(t_tid);
+    if (!tid || 0 != PyObject_SetAttrString(from, "_memray_tid", tid)) {
+        PyErr_Print();
```
pablogsal, line 394: Don't we want to return/abort in this case? Is technically not a catastrophic failure, but technically will produce incorrect results when switching back to, no?

src/memray/_memray.pyx
```diff
@@ -170,12 +170,21 @@ cdef class AllocationRecord:
...
                 raise NotImplementedError("Stack traces for deallocations aren't captured.")
 
             if max_stacks is None:
                 self._stack_trace = self._reader.get().Py_GetStackFrame(self._tuple[4])
+                if self._tuple[0] == self._reader.get().getMainThreadTid():
+                    to_skip = self._reader.get().getSkippedFramesOnMainThread()
+                    if len(self._stack_trace) > to_skip:
+                        to_keep = len(self._stack_trace) - to_skip
+                    else:
+                        to_keep = 0
```
pablogsal, line 186:
```suggestion
                    to_keep = max(len(self._track_trace) - to_skip, 0)
```

pablogsal: Although this may convert back and forth to Python stuff depending on what Cython decides to do, so I am not sure is a clear win.

src/memray/_memray.pyx
```diff
@@ -199,37 +208,66 @@ cdef class AllocationRecord:
...
+        )
+        native_stack = self.native_stack_trace()
 
-        if not python_stack:
-            yield from native_stack
-            return
+        cdef ssize_t num_non_entry_frames = 0
+        for val in is_entry_frame:
+            if not val:
+                num_non_entry_frames += 1
```
pablogsal, line 220:
```suggestion
        cdef ssize_t num_non_entry_frames = sum(int(not val) for val in is_entry_frame)
```

Same concern as before regarding going back and forth to python and C

src/memray/_memray.pyx
```diff
@@ -199,37 +208,66 @@ cdef class AllocationRecord:
...
+                        first_kept_python_frame = False
+
+                    hybrid_stack[hidx] = python_stack[pidx]
+                    hidx -= 1
+                    pidx -= 1
+
+                    if to_skip > 0:
+                        to_skip -= 1
+                        if to_skip == 0:
+                            first_kept_python_frame = True
```
pablogsal, line 257: This code is quite difficult to follow. I think we should extract some of these pieces into functions at the very least.

src/memray/_memray.pyx
```diff
@@ -199,37 +208,66 @@ cdef class AllocationRecord:
...
-                    yield python_frame
-                    if is_entry_frame[python_frame_index - 1]:
-                        break
+            if pidx >= 0 and self._is_eval_frame(symbol):
+                first = True
+                while pidx >= 0 and (first or not is_entry_frame[pidx]):
+                    first = False
+                    assert hidx >= 0
+                    if first_kept_python_frame:
+                        del hybrid_stack[hidx + 1:]
```
pablogsal, line 247: This is just deleting the extra `None` at the end np?

src/memray/_memray.pyx
godlygeek: Ah, I originally had `to_skip` and `to_keep` as `size_t` instead of `ssize_t`, and needed to avoid integer underflow - that's why it was slightly awkward. I'll check the generated code and make sure Cython does something reasonable with this `max` call, and use it if so.

src/memray/_memray.pyx
godlygeek: This generates some pretty not-great code - not just a conversion to Python, but a whole generator expression closure.

Really, since `is_entry_frame` is a vector, what we want is just `std::count(is_entry_frame.begin(), is_entry_frame.end(), 0)` - but Cython doesn't include a declaration for `std::count`, and I was being lazy. I'll just add one - it's more code, but more elegant.

PR #213, opened 2022-09-16
Use a custom frame eval function

godlygeek:
This incidentally solves #205, at least in most cases, as well as giving us about a 13% speedup.

It turned out to be surprisingly complex to implement, though.

The last commit adds a new test case for something I couldn't figure out how to fix.


godlygeek:
@pablogsal and I have decided to abandon this (again, after the first abandoned attempt in #207).

This does have some great advantages:

- It makes tracking in non-native mode a bit over 10% faster, which is a pretty considerable performance gain.
- It also addresses #205, by making it so that we see every frame before `_PyEval_EvalFrameDefault` can do any allocations for it, instead of after.
- It also injects a frame into the C stack for every Python call (at least, every Python call after tracking starts) that comes from Memray itself, which we know is built with debug symbols. We could look for that symbol when pairing up the native stack with the Python stack to produce the hybrid stack - we may have been able to improve our capabilities on platforms where the Python interpreter doesn't have DWARF debug information by leveraging that frame.

It also has some pretty serious disadvantages:

- Because the Memray API can be used to start tracking inside a running application at any time, some frames on the Python stack may be using the default eval function, and some may be using our custom eval function, which makes reasoning about the state of things harder, as well as making it harder to write any logic that depends on the existence of our injected frame (since it won't always be there).
- We thought we'd be able to throw away our original tracking approach (leveraging a profile function), but it turns out we still need to keep that: when there are frames on the stack above where tracking started that use the default frame eval function, we need some way to know when those frames are popped off the stack - a profile function can tell us that, but a custom eval function can't.
- Since we need to keep the profile function, we unset it below our custom eval function, and reset it when returning from our custom eval function. That leads to very complex interactions between the frame eval function and the profile function:
    1. Sometimes our custom frame eval function will fire before our profile function has ever run for a thread, but sometimes after, and it can see a different state depending on this (it can even see a different profile function installed, since we install a trampoline first)
    2. Previously we could detect when someone sets a different profile function underneath us, but when our custom eval function has temporarily uninstalled our profile function we lose the ability to detect that case.
    3. If tracking stops underneath a still running thread, and then later restarts, _we_ will reinstall our own profile function below our eval function. When popping frames, we will later try to restore our original profile function, only to discover that our own profile function is already installed, and our check for whether someone overwrote our profile function is now working against us since we overwrote our own.
- The method for registering custom frame eval functions is different across Python versions that we support - Python 3.7 and 3.8 require the eval function be registered by directly manipulating the `PyInterpreterState`, and Python 3.9 and later have a setter method that you need to call instead.
- The interface for custom frame eval functions is different across Python versions that we support - 3.7 and 3.8 use one signature, 3.9 and 3.10 use another, and 3.11 uses still another.
- 3.11 passes a `_PyInterpreterFrame*` to the eval function, rather than a `PyFrameObject*`. Handling this case requires us to make some of our lower level tracking code generic, working on different types of frame objects, and adding additional complexity.
- Custom frame eval functions have an uncertain future - they didn't prove to be very useful for their intended use case, and several of the CPython core devs dislike them and hope they get removed, as they hinder optimizations in the eval loop.
- And the nail in the coffin: using a custom frame eval function breaks frame inlining in 3.11, which results in many more native frames on the C stack, and therefore much slower native-mode tracking.

We have done more work for smaller performance gains in the past. But we originally hoped that this idea would make things simpler, and instead it makes things considerably more complex. There were a lot of gotchas, and there are likely still edge cases that we missed even after getting the entire test suite passing (and thinking up some new tests along the way).

PR #223, opened 2022-10-04
ci: Split docs building from linting

godlygeek:
This lays the groundwork for automatically uploading documentation when a release is cut.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -101,37 +101,21 @@ jobs:
         run: |
           make dev-install pycoverage
 
-  lint_and_docs:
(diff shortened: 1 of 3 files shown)


godlygeek:
This gets us 80% of the way there. On top of this, we should have an action that fires only when releases are published that publishes the `github-pages` artifact to the `gh-pages` branch using `actions/pages-deploy` - but I'm not sure how to test that right now, given that what's on main contains some staged documentation changes for the next release that we're not yet ready to publish.

pablogsal:
@godlygeek do you want to merge this PR as it is and work on the automation as a separate one?

PR #235, opened 2022-10-19
Make some source cleanup

pablogsal:
Update the source files to use C++ idioms, remove unused headers, remove
 shadowing variables, remove empty declarations, make member functions
 static when possible and fix small bugs due to calling virtual functions
in destructors.

### DIFF
--- a/src/memray/_memray/hooks.cpp
+++ b/src/memray/_memray/hooks.cpp
@@ -107,8 +107,8 @@ namespace memray::intercept {
 void*
 pymalloc_malloc(void* ctx, size_t size) noexcept
 {
-    PyMemAllocatorEx* alloc = (PyMemAllocatorEx*)ctx;
-    void* ptr = nullptr;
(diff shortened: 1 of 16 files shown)


src/memray/_memray/hooks.cpp
```diff
@@ -279,7 +276,7 @@ dlopen(const char* filename, int flag) noexcept
     void* ret = hooks::dlopen(filename, flag);
     if (ret) {
         tracking_api::Tracker::invalidate_module_cache();
-        if (filename && 0 != strstr(filename, "/_greenlet.")) {
+        if (filename && nullptr != strstr(filename, "/_greenlet.")) {
```
lkollar, line 279: Is this change necessary? My understanding is that `strstr` returns a null pointer if there is no match.

pablogsal: Yeah, that's what I am changing :)

lkollar (approved):

Issue #187, opened 2022-08-18
stats with --json output


kapilt:
i'm trying to measure the effect of a code change / branch to some application code. the app code uses multiprocessing pools to perform work. using memray with --follow-fork generates per pid files with parentage, but getting that to aggregate requires a custom script. unfortunately memray only does pretty output instead of json output afaics, so the aggregated stat view here can't easily be scripted  or parsed because its all terminal escape codes and formatted text. 

so the request would be to have a --json parameter  or --format=json parameter on some of the reporting commands (stats would be relevant for this use case).

godlygeek:
We wouldn't be willing to support this for all reporters, but if we scope it to only the `stats` reporter in particular, this seems like a reasonable request. It looks like the changes required to support that would be pretty localized, to only https://github.com/bloomberg/memray/blob/253b15bdfbd26ccff1779b1b5401cc09f78fb58d/src/memray/reporters/stats.py#L92 and the functions it calls. Plus some changes in https://github.com/bloomberg/memray/blob/253b15bdfbd26ccff1779b1b5401cc09f78fb58d/src/memray/commands/stats.py#L13 to accept the new `--json` argument and https://github.com/bloomberg/memray/blob/253b15bdfbd26ccff1779b1b5401cc09f78fb58d/src/memray/commands/stats.py#L36 to detect that `--json` was used and to pass the flag to `render`.

I've labelled this as "good first issue", since I think it makes a pretty easy place for someone to make a contribution to Memray. If anyone wants to pick this up and work on it, the first step would be deciding on a schema for the JSON this will return - what fields will be included, and what format the values will be in. We'd need to decide if it's more valuable to include the information displayed in the histogram in the JSON, or to include the raw information that's used to compute the histogram, for instance.

vmesel:
Hey guys, came across your repo and I'm willing to help on this issue:

What do you think about this JSON output?

```
{
    'stats': {
        'total_num_allocations': 0000,
        'total_memory_allocated': '128M',
        'allocator_types': [
            "type_1",
            "type_2",
        ],
        'top_allocations_by_size': [
            {
                "location": "xpto.py",
                "size": 13
            },
            {
                "location": "xpto.py",
                "size": 28
            },
            {
                "location": "xpto.py",
                "size": 120
            },
        ],
        'top_allocations_by_count': [
            {
                "location": "xpto.py",
                "count": 120
            },
        ],
    }
}
```

godlygeek:
[...]

PR #236, opened 2022-10-20
Add CLI arg to allow exporting stats to json file

meersuri:
Started work on adding a --json arg to save stats to a JSON file (#187)

Hoping to get early feedback on the direction. 
Things I'm not sure about - 
- If the aim is to output a JSON file or just print it to the console? If it's a file, how should I name it? pass the name of the results file to render and base it on that?
- what all to include in the JSON


godlygeek:
Hi @meersuri, thanks for picking this up! Sorry for the slow response to your questions.

> If the aim is to output a JSON file or just print it to the console?

I think we want it as a file, rather than force the user to pipe or redirect console output.

> If it's a file, how should I name it? pass the name of the results file to render and base it on that?

Yes, that's what we do for other reporters that generate files, like the `flamegraph` reporter or the `transform` reporter. The `flamegraph` reporter generates `memray-flamegraph-foo.html` for a `foo.bin` dump file, so I think we'd want something similar here: `memray-stats-foo.json` seems reasonable to me.

> what all to include in the JSON

I think https://github.com/bloomberg/memray/issues/187#issuecomment-1224885056 covers my thoughts on this.

godlygeek:
Thanks for the initial implementation, @meersuri! We're going to carry this over the finish line in #377, so I'm closing this PR, but thanks very much for getting this started. ❤️

Issue #215, opened 2022-09-24
IPython notebook (Jupyter) extension


ma-sadeghi:
### Is there an existing proposal for this?

- [X] I have searched the existing proposals

### Is your feature request related to a problem?

It'd be nice if we could use `memray` inside a Jupyter notebook (like `memory_profiler`'s `%memit`).

### Describe the solution you'd like

An extension that can be loaded into Jupyter environment so we can memory profile functions, cells, etc.

### Alternatives you considered

`memory_profiler`'s `%memit` magic command.

pablogsal:
Hi @ma-sadeghi and thanks for the proposal :heart: 

We will try to make a prototype and if is not too complex we will likely introduce it in one of the next releases!

ma-sadeghi:
Thank you so much! Looking forward to it!

PR #237, opened 2022-10-20
Allow memray to be executed as an IPython extension

pablogsal:
Allowing memray to profile code in Jupyter Notebooks will help a lot of
users such as data scientist when using memray as this will shorten the
gap between their workflow and the way the run the profiler.

Closes: #215

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -41,6 +41,7 @@ Memray can help with the following problems:
    native_mode
    examples/README
    api
+   jupyter_magic
 
 .. toctree::
    :hidden:
--- a/docs/jupyter_magic.rst
+++ b/docs/jupyter_magic.rst
@@ -1,0 +1,38 @@
+Jupyter Integration
+===================
+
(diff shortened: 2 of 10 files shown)


src/memray/_ipython.py
```diff
@@ -0,0 +1,169 @@
...
+                follow_fork=options.follow_fork,
+                code=indent(cell, " " * 4),
+            )
+            self.shell.run_cell(code)
+
+            merge_threads = not options.split_threads
+
+            reporter = None
+
+            with FileReader(dump_file, report_progress=True) as reader:
```
godlygeek, line 127: I am _extremely_ surprised to see that `report_progress=True` works in Jupyter notebooks. It's not perfect - it leaves extra blank lines between the cell's code and the output flamegraph where the progress bars were. But, a bit of extra wasted space is a small price to pay for the progress bar, so I think I'm happy to ship this despite the wasted space.

godlygeek (approved):
LGTM

Issue #239, opened 2022-10-26
Converting the output of summary reporter to csv/table/dataframe


sandeepkadapa:
### Is there an existing proposal for this?

- [x] I have searched the existing proposals

### Is your feature request related to a problem?

When there are a lot of allocations in the code, due to certain libraries like tensorflow or sklearn or pandas imported inside the code. The number of allocations goes to several thousand easily and the summary reporter isn't helpful in situations like this as the number of rows to display will be really high and is not easy to analyze.

### Describe the solution you'd like

So, if we convert summary reporter or any other suitable reporter output to a csv/table/dataframe that can be analyzed later with the help of pandas or any data manipulation libraries will be helpful to search and get useful insights. It can also help to profile modules within a big project without trying to profile each and every module individually.

pablogsal:
Thanks a lot @sandeepkadapa for the proposal ❤️ 

In #241 I have implemented a first version that allows the `transform` reporter to dump a CSV file for later analysis. I will try to get this into the next release :)

PR #241, opened 2022-10-31
Add 'csv' format to the 'transform' reporter

pablogsal:
When there are a lot of different allocations, some of the reporters that
we currently offer do not allow for a very granular inspection of the
data. This can happen for example when certain libraries like TensorFlow
or sklearn or pandas are imported as the number of allocations
goes easily to several thousand. In these situations, reporters like the
summary reporter are not helpful as the number of rows to display needs
to be really high and is not easy to analyze.

To allow other more specialized data analysis tools to work with the
data, add a new 'csv' output format to the 'transform' reporter that
will dump a comma-separated-value file with every allocation in the high
watermark as a row with columns representing different properties
of every allocation. The stack trace will be represented as a string
joined with characters that are illegal in function names.

Closes: #239

### DIFF
--- a/docs/transform.rst
+++ b/docs/transform.rst
@@ -72,6 +72,32 @@ allocated do not appear in the graph by default.
 Check the `gprof2dot documentation <https://github.com/jrfonseca/gprof2dot>`_ for more
 information on how to use the tool.
 
+csv
+~~~
+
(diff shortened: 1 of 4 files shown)


godlygeek (approved):
LGTM

PR #250, opened 2022-11-11
Consider macOS to be fully supported

### LINKED ISSUE #194: Allow disabling experimental support warning via env-var
For example, https://github.com/bloomberg/memray/blob/ed876621846e21e22df2ffa203058b3850065787/src/memray/_memray.pyx#L325. Printing out error on the stderr makes test that checks the stderr break even if memray works correctly.

Also ideally we should only print it once, but now it prints it for all instance creation (in case of pytest-memray) this is the number of tests we have (perhaps guard it with some kind of global variable?).

godlygeek:
Previously this warning would be printed each time a `Tracker` was constructed, which caused a lot of noise when `pytest-memray` was being used, since it creates one `Tracker` per pytest test. Now it will only be printed at most once per Python interpreter run.

Closes #194

### DIFF
--- a/docs/supported_environments.rst
+++ b/docs/supported_environments.rst
@@ -18,7 +18,7 @@ Supported operating systems
 
 You will have the best Memray experience on Linux.
 
-We have experimental support for macOS 11 or newer. We cannot support older
+We also support macOS 11 or newer. We cannot support older
 macOS versions, as they don't provide a C++17 compatible runtime. Although all
 features work on macOS, the way that macOS applications and Python libraries
 are typically distributed often results in subpar native stacks on Mac. See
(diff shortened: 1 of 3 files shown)


pablogsal:
Wouldn't this mean that even if someone is not using `--memray` will see the warning because the plugin loads?

godlygeek:
> Wouldn't this mean that even if someone is not using `--memray` will see the warning because the plugin loads?

Looks like yes - though `pytest-memray` could switch to lazily importing `memray` only when `--memray` is supplied...

godlygeek:
After some offline discussions, we've decided to drop the warning entirely, and to consider macOS to be fully supported rather than experimental.

pablogsal (approved):

PR #261, opened 2022-12-01
Add a checkbox that allows to hide frames from the import system

pablogsal:
The import system can introduce quite a lot of noise in flamegraphs
that can guide users towards the wrong place when checking for memory
peaks and memory leaks. As this information can still be interesting
we should not just hide it. To get the best of both worlds we add a new
checkbox to handle this situation similar to the "non interesting frames"
functionality.


src/memray/reporters/flamegraph.py
```diff
@@ -1,22 +1,35 @@
...
 
 MAX_STACKS = int(sys.getrecursionlimit() // 2.5)
 
+T = TypeVar("T")
+
+
+def pairwise(iterable: Iterator[T]) -> Iterable[Tuple[T, T]]:
+    a, b = tee(iterable)
+    next(b, None)
+    return zip_longest(a, b)
```
godlygeek, line 31: Let's not call this `pairwise`, since it's slightly different from `itertools.pairwise` (`zip_longest` vs `zip`).

I almost don't care what it's called, as long as it's anything but that. I originally thought this was just a backport of `itertools.pairwise`, but it's not, and that's gonna confuse future us.

news/261.feature
```diff
@@ -0,0 +1 @@
+Add a checkbox to flamegraphs that allow to hide frames from the import system
```
godlygeek, line 1:
```suggestion
Add a checkbox to flamegraphs that allows hiding frames from the import system
```

src/memray/reporters/flamegraph.py
```diff
@@ -91,12 +108,23 @@ def from_snapshot(
...
+            for index, (stack_frame, next_frame) in enumerate(
+                pairwise(reversed(stack))
+            ):
                 if is_cpython_internal(stack_frame):
                     num_skipped_frames += 1
                     continue
+                # Check if the next frame is from the import system. We check
+                # the next frame because the "import ..." code will be the parent
+                # of the first frame to enter the import system and we want to hide
+                # that one as well.
```
godlygeek, line 121: Whoa, do we? You and I were picturing two different things here, then. I thought we were just talking about hiding the frames inside `importlib`, not also hiding the actual call to `import` made from the Python code. The checkbox is labeled "Hide Import System Frames", and I wouldn't consider an `import numpy` statement in `__main__` to be an "import system frame"...

I thought we were just trying to make `import` less distracting and easier to follow by eliding the machinery, not eliding all of the imports entirely.

I wonder if that should be two different check boxes (hide imports vs hide importlib), or if that's overkill...

src/memray/reporters/templates/flamegraph.html
```diff
@@ -16,6 +16,11 @@
           title="Hide CPython eval frames and other, memray-related frames" checked>
     <label class="form-check-label text-white bg-dark">Hide Non-Relevant Frames</label>
 </div>
+<div class="form-check mr-3">
+  <input class="form-check-input" type="checkbox" data-toggle="tooltip" id="hideImportSystem"
+          title="Hide frames related to the Python import system" >
+    <label class="form-check-label text-white bg-dark">Hide import system Frames</label>
```
godlygeek, line 22: The other checkbox is "Hide Non-Relevant Frames", so this should be
```suggestion
    <label class="form-check-label text-white bg-dark">Hide Import System Frames</label>
```

Either both should be title case or neither should.

godlygeek (changes requested):

godlygeek (approved):
This updated version LGTM. Per an offline conversation with Pablo, we've agreed to defer the changes to hide `importlib` frames as non-relevant by default to a follow up PR.

PR #264, opened 2022-12-06
attach: Return an int from the trampoline function

godlygeek:
Previously we were returning a `const char*` pointer to a string literal, which we would then print in the debugger and check for in the `memray attach` code to ensure that it was printed. But, under some mysterious circumstances `gdb` could fail to print out the string that we've returned a pointer to, with "Address 0x7a1e7b90 out of bounds".

I don't really understand the root cause, but we can work around the issue by returning something simpler instead. We'll return just an int rather than a pointer to memory in the process we've attached to, and we'll make the debugger sessions either exit with an error code (gdb) or print a sentinel string we can recognize (lldb) if an error occurred.

### DIFF
--- a/src/memray/_memray/inject.cpp
+++ b/src/memray/_memray/inject.cpp
@@ -223,7 +223,7 @@ thread_body(void* arg)
 }  // unnamed namespace
 }  // namespace memray
 
-extern "C" __attribute__((visibility("default"))) const char*
+extern "C" __attribute__((visibility("default"))) int
 memray_spawn_client(int port)
 {
     // Running Python code directly in the point of attaching can lead to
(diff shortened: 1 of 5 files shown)


godlygeek:
k, I'll cherry pick that over once I see everything else passed.

pablogsal (approved):

godlygeek:
Re-requesting your review @pablogsal since there was a subtle issue with the original approach. It turns out we can't just trust the gdb return code, because old versions of gdb will exit with a zero return code even if they fail to execute some commands in the gdb script.

I've switched things up so that both the gdb and lldb scripts print an expression that evaluates to `"SUCCESS"` or `"FAILURE"`, and I now treat it as a successful run only if the debugger exited with a zero return code _and_ we see that an expression was printed with the value `"SUCCESS"`.

Yuck.

PR #266, opened 2022-12-07
docs: Document the attach subcommand

godlygeek:


godlygeek:
Rendered docs can be seen at https://godlygeek.github.io/memray/attach.html

docs/attach.rst
```diff
@@ -0,0 +1,90 @@
...
+container, you can use ``--cap-add=SYS_PTRACE`` to allow attaching to processes
+within the container.
+
+.. warning::
+
+   Allowing arbitrary processes to be traced `is insecure
+   <https://www.kernel.org/doc/html/latest/admin-guide/LSM/Yama.html>`_, as it
+   provides an easy vector for privilege escalation within a remote code
+   execution attack. Be sure to consider the security implications before you
+   choose to grant regular users the ability to attach to processes.
```
pablogsal, line 57: I would suggest maybe adding an example command to how to launch a docker container with `CAP_SYS_PTRACE` as this is going to be a very common situation.

docs/attach.rst
```diff
@@ -0,0 +1,90 @@
+Attaching to a running process
+==============================
+
+Memray allows you to attach to a running process and to observe the allocations
+it performs once you've attached. This doesn't allow you to see where memory
```
pablogsal, line 5: I propose to put the 

> This doesn't allow you to see where memory was allocated before you attached to the process

in an admonition or in bold

docs/attach.rst
```diff
@@ -0,0 +1,90 @@
...
+
+The general form of the ``attach`` subcommand is:
+
+.. code:: shell
+
+    memray attach [options] <pid>
+
+The only required argument for ``memray attach`` is a process ID to attempt to
+attach to. You must be :ref:`able to attach a debugger <ptrace privs>` to that
+process, and that process must be a Python process capable of running ``import
```
pablogsal, line 24: I would suggest to put the 

> and that process must be a Python process capable of running...

in bold or in an admonition

docs/attach.rst
```diff
@@ -0,0 +1,90 @@
...
+memray`` (that is, the Memray package must be installed in the environment that
+the process is running in).
+
+By default this will open a :doc:`live mode TUI <live>` showing the allocations
+that the process performs after you've attached, but you can instead provide
+the name of a capture file to write to with the ``-o`` option, allowing you to
+analyze the captured allocations with any Memray reporter you'd like. You can
+also provide most of the options that ``memray run`` accepts. See :ref:`the
+CLI reference <memray attach CLI reference>` for details.
```
pablogsal, line 33: Maybe we should clarify the status of the process after detaching because is not the same in `live` mode and in non `live` mode.

This makes me wonder if we should offer a "stop memray" to stop tracking a previously attached process in non live mode. WDYT?

docs/attach.rst
```diff
@@ -0,0 +1,90 @@
...
+Memray leverages a debugger for attaching to the process. It is compatible with
+both gdb and lldb, but one or the other must be installed in order for ``memray
+attach`` to work. Only a super user (either root, or a user with the
+``CAP_SYS_PTRACE`` capability) can attach to processes run by another user.
+Further, security settings on modern Linux systems typically prevent a regular
+user from attaching even to their own processes. You can loosen that
+restriction by writing ``0`` to ``/proc/sys/kernel/yama/ptrace_scope`` as root,
+allowing any user to attach to their own processes. When running a Docker
+container, you can use ``--cap-add=SYS_PTRACE`` to allow attaching to processes
+within the container.
```
pablogsal, line 49: Let's add that in macOS the debugger may ask for a password or authentication before proceeding (this happens in my machine when I try to debug a process I own but that is not the child of the ataching process).

pablogsal (approved):

PR #268, opened 2022-12-09
Mark frames from the import system as irrelevant

pablogsal:
To reduce the noise of the import system frames, mark them as irrelevant
so they are hidden by default unless the checkbox to hide irrelevant
frames is not marked. Additionally, rename the checkbox from
"hide non-relevant" to "hide irrelevant" for clarity.


godlygeek (approved):
LGTM - love it!

Issue #68, opened 2022-04-29
Set logging interval for run command


wigging:
As I mentioned in the discussion linked below, it would be useful to set the logging interval for the run command. For example, to log memory usage every 5 seconds something like the following option could be implemented:

```
memray run --interval 5 file.py
```

### Discussed in https://github.com/bloomberg/memray/discussions/66

<div type='discussions-op-text'>

<sup>Originally posted by **wigging** April 29, 2022</sup>
I'm using Memray to profile the memory usage of a long running Python program. The bin file is several gigabytes and I'm starting to run out of storage space. I think if I set the logging interval to a longer duration, it would decrease the size of the bin file. But the run command does not have any options for setting the logging interval. How can set the logging interval and/or decrease the size of the bin file?</div>

godlygeek:
> it would be useful to set the logging interval for the run command. For example, to log memory usage every 5 seconds

This doesn't seem possible to me, on a very fundamental level. The way Memray works is by capturing all allocations and deallocations that are performed - it doesn't have a sampling interval like you're picturing, it's a continuous process.

Memray is a tracing profiler, not a sampling profiler.

godlygeek:
I suppose it would be possible to aggregate records in memory, though - if you're OK with losing some information about the total number of allocations and the total bytes allocated, we could set some interval (like 5s) and, each time the timer expires, output only the allocation records that were either

a) allocated during that 5s window and not freed during that 5s window, or
b) freed during that 5s window and not allocated during that 5s window.

That would be possible, with an ensuing loss of fidelity.

godlygeek:
@pablogsal Is that what you were picturing?

pablogsal:
> @pablogsal Is that what you were picturing?

No, I think @wigging was referring to the interval at which we log the RSS. 🤔

wigging:
The [memory_profiler](https://github.com/pythonprofilers/memory_profiler) tool allows the user to set the interval (sampling period in seconds). For example, if I set the interval for memory_profiler to 5 seconds, it will log the memory usage of the python program every 5 seconds. This helps reduce the size of the log file for long running programs. I would like a similar feature for Memray while still being able to visualize how the memory is being allocated in the program.

godlygeek:
`memory_profiler` is a sampling profiler, which is why that option makes sense for it.

Memray is a tracing profiler, which doesn't have a sampling interval - it tracks each allocation as it occurs, instead of periodically inspecting the state of the heap as a sampling profiler would.

[...]

PR #277, opened 2022-12-23
Allow aggregating allocations in memory in the tracked process

godlygeek:
Relates-to: #68


src/memray/_memray/snapshot.h
```diff
@@ -226,6 +224,88 @@ class HighWatermarkFinder
...
+    // allocations or deallocations since the current peak.
+    struct UsageHistory
+    {
+        uint64_t last_known_peak{};
+        size_t allocations_contributed_to_last_known_peak{};
+        size_t bytes_contributed_to_last_known_peak{};
+        // NOTE: Signed, as we could have more deallocations than allocations
+        // since the last peak, or more bytes deallocated than allocated.
+        int64_t count_since_last_peak{};
+        int64_t bytes_since_last_peak{};
```
godlygeek, line 289: It would use extra space, but I'm tempted to simplify by making this
```cpp
size_t count_deallocated_since_last_peak{};
size_t bytes_deallocated_since_last_peak{};
size_t count_allocated_since_last_peak{};
size_t bytes_allocated_since_last_peak{};
```
Especially because my first attempt at writing this had a bug that only manifested on 32-bit because of arithmetic conversion rules (passing `-some_size_t` to a function accepting `int64_t` wound up passing a negative number on 64-bit, as I expected, but a positive number on 32-bit, and I needed to correct it to `-static_cast<int64_t>(some_size_t)`

pablogsal:
Hummmm, I am getting a lot of errors compiling this in my Linux box:

```
gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -D_FILE_OFFSET_BITS=64 -DUSE_MEMRAY_TLS_MODEL=1 -I/home/pablogsal/.pyenv/versions/memray/lib/python3.10/site-packages/memray/_memray -Isrc/memray/_memray -Isrc -I/home/pablogsal/github/memray/src/vendor/libbacktrace/install/include -I/home/pablogsal/.pyenv/versions/memray/include -I/home/pablogsal/.pyenv/versions/3.10.1/include/python3.10 -c src/memray/_memray.cpp -o build/temp.linux-x86_64-3.10/src/memray/_memray.o -std=c++17 -Wall -flto -fvisibility=hidden
src/memray/_memray.cpp:1123:68: error: ‘memray::tracking_api::FileFormat’ has not been declared
 1123 |   __pyx_e_6memray_7_memray_ALL_ALLOCATIONS = memray::tracking_api::FileFormat::ALL_ALLOCATIONS,
      |                                                                    ^~~~~~~~~~
src/memray/_memray.cpp:1124:75: error: ‘memray::tracking_api::FileFormat’ has not been declared
 1124 |   __pyx_e_6memray_7_memray_AGGREGATED_ALLOCATIONS = memray::tracking_api::FileFormat::AGGREGATED_ALLOCATIONS
      |                                                                           ^~~~~~~~~~
src/memray/_memray.cpp:1257:16: error: ‘AggregatedCaptureReaggregator’ in namespace ‘memray::api’ does not name a type
 1257 |   memray::api::AggregatedCaptureReaggregator __pyx_v_aggregator;
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/memray/_memray.cpp:1265:25: error: ‘AggregatedAllocation’ in namespace ‘memray::tracking_api’ does not name a type
 1265 |   memray::tracking_api::AggregatedAllocation __pyx_v_record;
      |                         ^~~~~~~~~~~~~~~~~~~~
src/memray/_memray.cpp:2374:105: error: ‘FileFormat’ does not name an enumeration in ‘memray::tracking_api’
 2374 | static CYTHON_INLINE PyObject* __Pyx_PyInt_From_enum__memray_3a__3a_tracking_api_3a__3a_FileFormat(enum memray::tracking_api::FileFormat value);
      |                                                                                                         ^~~~~~
src/memray/_memray.cpp:2374:127: error: ‘FileFormat’ in namespace ‘memray::tracking_api’ does not name a type
 2374 | static CYTHON_INLINE PyObject* __Pyx_PyInt_From_enum__memray_3a__3a_tracking_api_3a__3a_FileFormat(enum memray::tracking_api::FileFormat value);
```

pablogsal:
Some profiling data of hotspots:

overhead1
overhead2
overhead3
overhead4

pablogsal:
Seems that locking continues to be a pain (this is **self** time):

overhead5

src/memray/_memray.pyx
```diff
@@ -713,6 +811,11 @@ cdef class FileReader:
 
     def get_allocation_records(self):
         self._ensure_not_closed()
+        if self._header["file_format"] == FileFormat.AGGREGATED_ALLOCATIONS:
+            raise NotImplementedError(
+                "Can't get all allocations from a pre-aggregated capture file."
```
pablogsal, line 816: I think we are missing a test that exercises this code path

src/memray/_memray/record_writer.cpp
```diff
@@ -31,17 +39,141 @@ getPythonAllocator()
...
+    virtual bool
+    writeThreadSpecificRecord(thread_id_t tid, const NativeAllocationRecord& record) override;
+    virtual bool writeThreadSpecificRecord(thread_id_t tid, const ThreadRecord& record) override;
+
+    virtual bool writeHeader(bool seek_to_start) override;
+    virtual bool writeTrailer() override;
+
+    virtual void
+    setMainTidAndSkippedFrames(thread_id_t main_tid, size_t skipped_frames_on_main_tid) override;
+    virtual std::unique_ptr<RecordWriter> cloneInChildProcess() override;
```
pablogsal, line 79: You can drop `virtual` here in all of these as you are already using `override`

src/memray/_memray/record_writer.cpp
```diff
@@ -31,17 +39,141 @@ getPythonAllocator()
...
+}
+
+class StreamingRecordWriter : public RecordWriter
+{
+  public:
+    explicit StreamingRecordWriter(
+            std::unique_ptr<memray::io::Sink> sink,
+            const std::string& command_line,
+            bool native_traces);
+    virtual ~StreamingRecordWriter() = default;
```
pablogsal, line 54:
```suggestion
    ~StreamingRecordWriter() override = default;
```

src/memray/_memray/record_writer.cpp
```diff
@@ -31,17 +39,141 @@ getPythonAllocator()
...
+    void operator=(const AggregatingRecordWriter&) = delete;
+    void operator=(AggregatingRecordWriter&&) = delete;
+
+    virtual bool writeRecord(const MemoryRecord& record) override;
+    virtual bool writeRecord(const pyrawframe_map_val_t& item) override;
+    virtual bool writeRecord(const UnresolvedNativeFrame& record) override;
+
+    virtual bool writeMappings(const std::vector<ImageSegments>& mappings) override;
+
+    virtual bool writeThreadSpecificRecord(thread_id_t tid, const FramePop& record) override;
```
pablogsal, line 112: Same as before regarding `virtual` and `override`

src/memray/_memray/record_writer.cpp
```diff
@@ -31,17 +39,141 @@ getPythonAllocator()
...
+};
+
+class AggregatingRecordWriter : public RecordWriter
+{
+  public:
+    explicit AggregatingRecordWriter(
+            std::unique_ptr<memray::io::Sink> sink,
+            const std::string& command_line,
+            bool native_traces);
+    virtual ~AggregatingRecordWriter() = default;
```
pablogsal, line 99:
```suggestion
    ~AggregatingRecordWriter() override = default;
```

src/memray/_memray/tracking_api.cpp
```diff
@@ -885,14 +872,11 @@ Tracker::updateModuleCacheImpl()
     if (!d_unwind_native_frames) {
         return;
     }
-    auto writer_lock = d_writer->acquireLock();
-    if (!d_writer->writeRecordUnsafe(MemoryMapStart{})) {
-        std::cerr << "memray: Failed to write output, deactivating tracking" << std::endl;
-        deactivate();
-    }
+
+    std::vector<ImageSegments> mappings;
```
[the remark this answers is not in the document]
pablogsal: We could use the value of `d_mappings_by_generation[-1].size()`

src/memray/_memray/record_writer.cpp
```diff
@@ -93,23 +392,266 @@ RecordWriter::writeTrailer()
...
+    for (const auto& memory_snapshot : d_memory_snapshots) {
+        if (!writeSimpleType(AggregatedRecordType::MEMORY_SNAPSHOT) || !writeSimpleType(memory_snapshot))
+        {
+            return false;
+        }
+    }
+
+    for (const auto& [tid, thread_name] : d_thread_name_by_tid) {
+        if (!writeSimpleType(AggregatedRecordType::CONTEXT_SWITCH)
+            || !writeSimpleType(ContextSwitch{tid})
```
pablogsal, line 462: Why do we need to simulate a context switch here?

src/memray/_memray/record_writer.cpp
```diff
@@ -93,23 +392,266 @@ RecordWriter::writeTrailer()
...
 }
 
+bool
+AggregatingRecordWriter::writeRecord(const MemoryRecord& record)
+{
+    std::lock_guard<std::mutex> lock(d_mutex);
+    MemorySnapshot snapshot;
+    snapshot.ms_since_epoch = record.ms_since_epoch;
+    snapshot.rss = record.rss;
+    snapshot.heap = d_high_water_mark_aggregator.getCurrentHeapSize();
```
pablogsal, line 546: This is a bit verbose. Any reason why we are not just doing:

```suggestion
    MemorySnapshot snapshot{record.ms_since_epoch, record.rss, d_high_water_mark_aggregator.getCurrentHeapSize());
```

src/memray/_memray/record_writer.cpp
```diff
@@ -93,23 +392,266 @@ RecordWriter::writeTrailer()
...
+bool
+AggregatingRecordWriter::writeThreadSpecificRecord(thread_id_t tid, const AllocationRecord& record)
+{
+    std::lock_guard<std::mutex> lock(d_mutex);
+    Allocation allocation;
+    allocation.tid = tid;
+    allocation.address = record.address;
+    allocation.size = record.size;
+    allocation.allocator = record.allocator;
+    allocation.native_frame_id = 0;
```
pablogsal, line 618: Any reason why we are not just doing

```suggestion
    Allocation allocation{
    tid,
    record.address,
    record.size,
    record.allocator,
    }
```

?

src/memray/_memray/record_writer.cpp
```diff
@@ -93,23 +392,266 @@ RecordWriter::writeTrailer()
...
+    allocation.tid = tid;
+    allocation.address = record.address;
+    allocation.size = record.size;
+    allocation.allocator = record.allocator;
+    allocation.native_frame_id = record.native_frame_id;
+    auto& stack = d_python_stack_ids_by_thread[tid];
+    allocation.frame_index = stack.empty() ? 0 : stack.back();
+    allocation.native_segment_generation = d_mappings_by_generation.size();
+    allocation.n_allocations = 1;
+    d_high_water_mark_aggregator.addAllocation(allocation);
```
pablogsal, line 645: Same as before, any reason why we are not just doing

```suggestion
    std::lock_guard<std::mutex> lock(d_mutex);
    auto& stack = d_python_stack_ids_by_thread[tid];
    Allocation allocation{
     tid,
     record.address,
     record.size,
    record.allocator,
    record.native_frame_id,
    stack.empty() ? 0 : stack.back(),
    d_mappings_by_generation.size(),
    };
    d_high_water_mark_aggregator.addAllocation(allocation);
```

src/memray/_memray/snapshot.h
godlygeek: Narrowing is a problem when considering the integer types in the abstract, but I don't think it's possible to hit the implementation-defined behavior given the values we'll have in our domain. The count can never be greater than 1 or less than `-(SIZE_MAX / PAGE_SIZE)` (we never record more than one allocation at a time, and the largest number of deallocations we can record at one time is 1 per mmap'd region, which are a minimum of 1 page).

It's theoretically possible that the size could be large enough to trigger implementation defined behavior, but that would only come up if someone allocated more than 9 exabytes of memory in a single allocation, which isn't realistic on modern hardware, and probably won't be for a very long time.

godlygeek: That said - I do think the code would be easier to read and understand if we stuck with unsigned values for everything, at the cost of trading off some space. If you're happy with the extra memory usage, that works for me.

pablogsal:
You need to fix the conflicts in `src/memray/commands/run.py` from your other PR :)

tests/integration/test_main.py
```diff
@@ -510,6 +514,65 @@ def test_successful_parse(self, tmp_path):
         for _, count in record_count_by_type.items():
             assert count > 0
 
+    def test_successful_parse_of_aggregated_capture_file(self, tmp_path):
```
pablogsal, line 517: I would like a test that checks that the **aggregated** data at the high watermark using both reporters is the same for some bunch of code examples.

src/memray/_memray/record_writer.cpp
godlygeek: Just to allow us to use the same format for `THREAD_RECORD` between the aggregated capture file format and the all-allocations capture file format. The old file format didn't include the tid in the thread record, and instead the thread record was implicitly related to the last thread switched to. I matched that here only to allow reusing code in the reader.

src/memray/_memray/record_writer.cpp
godlygeek: I prefer not to use positional struct initialization for structures with more than 3 or 4 fields, as we've seen that it's a frequent source of bugs where a value gets assigned to the wrong field accidentally. Your proposal is the same number of lines of code, and with your version someone reading the code can't be sure whether the correct parameter was assigned to each field without opening the header and cross-referencing the initialization against the order of fields in the `struct` declaration. Same reason as I try not to pass more than 3 or 4 positional parameters in Python code.

src/memray/_memray/record_writer.h
```diff
@@ -1,71 +1,69 @@
...
-            bool native_traces);
-    void setMainTidAndSkippedFrames(thread_id_t main_tid, size_t skipped_frames_on_main_tid);
+    virtual ~RecordWriter() = default;
 
     RecordWriter(RecordWriter& other) = delete;
     RecordWriter(RecordWriter&& other) = delete;
     void operator=(const RecordWriter&) = delete;
     void operator=(RecordWriter&&) = delete;
 
+    virtual bool writeRecord(const MemoryRecord& record) = 0;
```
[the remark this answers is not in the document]
godlygeek: Discussed offline: this doesn't really work nicely because of this:
https://github.com/bloomberg/memray/blob/1e5ef53c2f6d068e14e8c19aedb22d92b77a0b4c/src/memray/_memray/tracking_api.cpp#L426

We treat `Tracker` as a singleton, and we store a pointer to the created singleton instance to be referenced later. For that to work, we need to know the concrete type of the tracker, which means that templatizing it based on the type of the writer doesn't really work nicely.

I suspect that the overhead of the `virtual` calls is fairly low, and architecturally they don't seem very easy to avoid, so I'd rather keep it the way I've written it unless profiling shows it to be a significant bottleneck.

src/memray/_memray/record_writer.h
```diff
@@ -1,71 +1,69 @@
...
+
+    virtual bool writeHeader(bool seek_to_start) = 0;
+    virtual bool writeTrailer() = 0;
+
+    virtual void setMainTidAndSkippedFrames(thread_id_t main_tid, size_t skipped_frames_on_main_tid) = 0;
+    virtual std::unique_ptr<RecordWriter> cloneInChildProcess() = 0;
+
+  protected:
+    // Expose the sink for use by the following helper functions.
+    RecordWriter(std::unique_ptr<memray::io::Sink> sink);
```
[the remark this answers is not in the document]
godlygeek: Is there any way that a protected constructor of an abstract class could ever be called implicitly? I can't think of one...

Regardless, it's an easy change to make, so I'll make it.

src/memray/_memray/hooks.cpp
```diff
@@ -205,7 +205,11 @@ calloc(size_t num, size_t size) noexcept
 {
     assert(hooks::calloc);
 
-    void* ret = hooks::calloc(num, size);
+    void* ret;
```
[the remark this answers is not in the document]
godlygeek: Yeah, the commit message for that one is:
```
Ignore allocations under calloc and posix_memalign

Apparently Alpine implements `calloc` in terms of `malloc`, and
`posix_memalign` in terms of `aligned_alloc`. Ensure that we don't
record duplicate allocations on Alpine.

Our `ALL_ALLOCATIONS` capture file format incorrectly recorded both the
`calloc` and the `malloc` it is implemented in terms of, but because the
`malloc` happens first and returns the same pointer as the `calloc`
later returns, the aggregator wound up overwriting the allocation record
for the `malloc` with the one for the `calloc`, silently dropping it.
```

I should maybe have made this a separate PR. It's an existing bug that our existing tests didn't catch, but tests introduced on this PR did.

src/memray/_memray/snapshot.cpp
```diff
@@ -192,6 +200,215 @@ TemporaryAllocationsAggregator::getSnapshotAllocations(bool merge_threads)
...
+        // We're currently at a new peak that we haven't yet fallen from.
+        final_peak_count++;
+    }
+
+    for (const auto& [loc, usage] : d_usage_history_by_location) {
+        size_t water_mark_allocations = 0;
+        size_t water_mark_bytes = 0;
+        size_t leaked_allocations = 0;
+        size_t leaked_bytes = 0;
+
```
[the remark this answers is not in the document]
godlygeek: I think you've misread. They're used below, when constructing the `AggregatedAllocation` object:
```cpp
        AggregatedAllocation alloc{
                loc.thread_id,
                loc.allocator,
                loc.native_frame_id,
                loc.python_frame_id,
                loc.native_segment_generation,
                water_mark_allocations,
                leaked_allocations,
                water_mark_bytes,
                leaked_bytes,
        };
```

src/memray/_memray/snapshot.cpp
```diff
@@ -192,6 +200,215 @@ TemporaryAllocationsAggregator::getSnapshotAllocations(bool merge_threads)
...
+            auto removal_stats = d_mmap_intervals.removeInterval(allocation.address, allocation.size);
+            for (const auto& [interval, allocation] : removal_stats.freed_allocations) {
+                recordUsageDelta(allocation, -1, -static_cast<int64_t>(interval.size()));
+            }
+            for (const auto& [interval, allocation] : removal_stats.shrunk_allocations) {
+                recordUsageDelta(allocation, 0, -static_cast<int64_t>(interval.size()));
+            }
+            for (const auto& [interval, allocation] : removal_stats.split_allocations) {
+                recordUsageDelta(allocation, 1, -static_cast<int64_t>(interval.size()));
+            }
```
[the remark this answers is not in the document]
godlygeek: I know, I just couldn't think of a better name ¯\\\_(ツ)\_/¯

OK. This code is a bit tricky to read, since the function parameter is called `allocation`, but might actually represent a deallocation, and if it does we need to look up the location(s) and size(s) of the allocation(s) being freed. I'm gonna handle this by renaming the function parameter to `const Allocation& allocation_or_deallocation`, and making every branch create a reference called `allocation` that always refers to the allocation being made or freed, and never refers to the deallocation record (which doesn't contain a location or size).

src/memray/_memray/snapshot.h
godlygeek: Actually - no, this proposed approach has a bigger problem than the narrowing. If I store the allocations and deallocations separately as unsigned integers, they might overflow. As things are written now, if you do
```cpp
free(malloc(2)); // High water mark
while (true) {
    free(malloc(1));
}
```
then each loop iteration switches `count_since_last_peak` and `bytes_since_last_peak` from 0 to 1 and back.

If we changed this to 4 separate counters, then each loop iteration would instead increment all 4 counters by `1`, and on a 32-bit system we'd overflow on the 4294967296th loop iteration.

godlygeek: ... OK, hear me out. We _could_ use an unsigned integer instead of a signed integer, but otherwise treat it exactly as we treat it today, allowing it to go "negative" (wrapping around to a large positive number), and count on the fact that `size_t(3) + size_t(-2) == size_t(1)`, even though `size_t(-2)` is a large positive number.

That fixes the narrowing, at the cost of making it much harder to read the structure in a debugger...

src/memray/_memray/tracking_api.cpp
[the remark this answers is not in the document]
godlygeek: Sure, we could, but I mean, 🤷 - this isn't a hot path.

But OK, I'll add a `.reserve()`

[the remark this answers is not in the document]
godlygeek: ... but I can't base it on `d_mappings_by_generation[-1].size()`, because that's an attribute of the `AggregatingRecordWriter`, and we're in `tracking_api.cpp`, which doesn't even know which concrete subclass of `RecordWriter` we've got.

I'll add a `static` variable to the function to track the last size.

src/memray/_memray.pyx
```diff
@@ -369,7 +382,8 @@ cdef class Tracker:
 
     def __cinit__(self, object file_name=None, *, object destination=None,
                   bool native_traces=False, unsigned int memory_interval_ms = 10,
-                  bool follow_fork=False, bool trace_python_allocators=False):
+                  bool follow_fork=False, bool trace_python_allocators=False,
+                  FileFormat file_format=FileFormat.ALL_ALLOCATIONS):
```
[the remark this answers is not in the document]
godlygeek: I'll add a `raise`, we can always remove it later if we find a useful case for aggregating to a socket.

src/memray/_memray/snapshot.pxd
```diff
@@ -38,6 +43,11 @@ cdef extern from "snapshot.h" namespace "memray::api":
     cdef cppclass index_thread_pair_hash:
         pass
 
+    cdef cppclass HighWaterMarkAggregator:
+        void addAllocation(const Allocation& allocation) except+
+        size_t getCurrentHeapSize() except+
```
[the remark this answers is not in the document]
godlygeek: It is in fact declared `noexcept`, I'll remove the `except+`.

src/memray/_memray/frame_tree.h
```diff
@@ -9,12 +9,23 @@ namespace memray::tracking_api {
 class FrameTree
 {
   public:
-    using index_t = uint32_t;
+    using index_t = uint32_t;  // TODO: Shouldn't this be size_t?
```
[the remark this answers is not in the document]
godlygeek: OK. That's unrelated to this PR, just something I noticed while going through this one. Rather than try to address that here, I'll remove the `TODO` comment, and add an issue for fixing this - https://github.com/bloomberg/memray/issues/285

PR #281, opened 2023-01-03
Ensure the tests can be ran independently with Python 3.11

pablogsal:
As Python 3.11 allocates memory when a function is first traced, some of
our tests results are depending on the fact that the allocators don't
trigger when a function is traced for the first time (what happens is
that other tests warm the pymalloc allocator).

To avoid this, add a new context manager that allows to run a piece of
code under tracing to warmup the caches, including the calling function.
For the calling function to be included the context manager needs to be
implemented in C.

**Describe your changes**

**Testing performed**

**Additional context**


pablogsal:
I also tested that after this PR we can run every test individually in isolation by running:

```
python -m pytest tests  --co | grep test_ | grep Function | sed 's/<Function \(.*\)>/\1/' | xargs -n 1  pytest tests -v -k
```

godlygeek (approved):
LGTM

PR #286, opened 2023-01-09
Add a new section to the docs about memray performance

pablogsal:
- Add benchmarks from pyperformance
- Add a new section to the docs about memray performance


pablogsal:
Instructions for review: don't review anything in the `benchmarks/benchmarking/cases` folder as these are the pyperformance cases taken directly from the repo and modified to be able to run with different configurations.

benchmarks/benchmarks.py
```diff
@@ -4,10 +4,21 @@
 
 from memray import AllocatorType
 from memray import FileReader
-from memray import MemoryAllocator
+
+try:
+    from memray._test import MemoryAllocator
+except ImportError:
+    from memray import MemoryAllocator
```
godlygeek, line 11: Hm? Is this for compatibility across versions of Memray? Any reason to still support the old one?

benchmarks/benchmarks.py
```diff
@@ -105,9 +116,8 @@ def setup(self):
         os.unlink(self.tempfile.name)
         self.tracker = Tracker(self.tempfile.name)
         with self.tracker:
-            for _ in range(MAX_ITERS):
-                self.allocator.valloc(1234)
-                self.allocator.free()
+            self.allocator.valloc(1234)
+            self.allocator.free()
```
godlygeek, line 120: Looks like the loop here was accidentally dropped.
```suggestion
            for _ in range(LOOPS):
                self.allocator.valloc(1234)
                self.allocator.free()
```

godlygeek (commented):
I'm a bit surprised that we're committing the JSON results files, rather than just artifacts generated from them. What's the rationale for why those are useful to include in the repo?

Also, I realize this is a stupid nit, but the filename `docs/_static/images/plot_Python_allocators_Native.png` reads like a SpOnGeBob mEme - it'd be nicer if the file names were all lower case, or if they consistently used title case.

docs/performance.rst
```diff
@@ -0,0 +1,108 @@
+Performance
+===========
+
+This section contains an analysis of the overhead of running ``memray`` over different
```
godlygeek, line 4: Typographic nit: you've consistently used \`\`memray\`\` here to refer to the name of the tool. I've tried to consistently use Memray as the name of the tool in all of our other docs pages, and I use \`\`memray\`\` only when I'm referring to the name of the importable module or the entry point script.

docs/performance.rst
```diff
@@ -0,0 +1,108 @@
...
+
+Source of the overhead
+----------------------
+
+The overhead of running ``memray`` comes from two places:
+
+- A `Python profile function <https://docs.python.org/3/library/sys.html#sys.setprofile>`_ implemented in
+  C++. This function is called every time a Python function is called or returns
+  and is used to register the function calls so ``memray`` can reconstruct the
+  stack trace of every allocations. Although the overhead is very very small, it
```
godlygeek, line 55:
```suggestion
  stack trace of every allocation. Although the overhead is very, very small, it
```

docs/performance.rst
```diff
@@ -0,0 +1,108 @@
...
+  is not negligible. This means that the fastest your applications calls Python
```
godlygeek, line 56:
```suggestion
  adds up. This means that the more your application calls Python
```

docs/performance.rst
```diff
@@ -0,0 +1,108 @@
...
+  functions the bigger the overhead will be.
+- The allocation registering code. This is the main source of the overhead.
+  Every time your application makes a memory allocation or deallocation,
+  ``memray`` needs to register it. This means that the faster your application
```
godlygeek, line 60:
```suggestion
  ``memray`` needs to register it. This means that the more frequently your application
```

docs/performance.rst
```diff
@@ -0,0 +1,108 @@
...
+  allocates memory, the bigger the overhead will be. Tracking Python allocators
+  will greatly increase this overhead because the Python allocator are used for
```
godlygeek, line 62:
```suggestion
  will greatly increase this overhead because the Python allocator is used for
```

docs/performance.rst
```diff
@@ -0,0 +1,108 @@
...
+  small, short lived objects which means that is generally called with very high
```
godlygeek, line 63:
```suggestion
  small, short lived objects which means that it's generally called with very high
```

docs/performance.rst
```diff
@@ -0,0 +1,108 @@
...
+Results
+-------
+
+The following plots show the overhead of running ``memray`` over different test
+cases using violin plots. Each row corresponds to the result of a given test.
+For a given row, the continuos vertical line in the middle corresponds to the
+**mean** while the two dashed vertical vars at the left and right of the mean
+correspond to the **10% and 90% quantiles**.  The shaded area represents the
+distribution of the data, with thicker areas indicating a higher density of data
+points. The **ALL** row corresponds to the distribution of all the results of all the
```
godlygeek, line 75:
```suggestion
points. The **ALL** row corresponds to the distribution of all the results across all the
```

godlygeek (changes requested):

benchmarks/benchmarks.py
pablogsal: Yep, `asv` us used to benchmark across all it history so it must be able to run reasonably well in all commits. (is ok if **some** tests don't run, `asv` handles that) .

benchmarks/benchmarking/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+            type_results.append(data)
+        results.append(BenchmarkResult(name=name, data=type_results))
+    return results
+
+
+if __name__ == "__main__":
+    if RESULTS_DIR.exists():
+        raise RuntimeError(f"Results directory {RESULTS_DIR} already exists")
+
+    RESULTS_DIR.mkdir(exist_ok=True)
```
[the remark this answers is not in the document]
pablogsal: Ah, indeed. This is because previously I was removing the directory but I figured out that was a bad idea in the final version.

pablogsal:
> I'm a bit surprised that we're committing the JSON results files, rather than just artifacts generated from them. What's the rationale for why those are useful to include in the repo?

That is an error. I will remove them

benchmarks/benchmarking/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+        self,
+        run_name: str,
+        template_file: pathlib.Path,
+        tracker_options: str,
+        results_file: str,
+    ) -> None:
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            case_file = pathlib.Path(tmpdirname) / f"{self.file_name}.py"
+            code = template_file.read_text().replace(
+                "MEMRAY_TRACKER_CODE_HERE", tracker_options
```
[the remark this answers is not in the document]
pablogsal: > I guess this replace() is fine - we shouldn't need to change it much - but it seems unnecessarily hacky.

Yeah, given that this code is not even part of any test, I prioritized having something working over having something specifically elegant.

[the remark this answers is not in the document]
pablogsal: OK, I have switched to use a temporary helper file as you suggested

pablogsal:
> Also, I realize this is a stupid nit, but the filename `docs/_static/images/plot_Python_allocators_Native.png` reads like a SpOnGeBob mEme - it'd be nicer if the file names were all lower case, or if they consistently used title case.

76ygkf

lkollar (approved):

Issue #288, opened 2023-01-09
Receiving segmentation error


sethmlarson:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Running pytest-memray with coverage in urllib3, occasionally seeing segmentation errors ([ref](https://github.com/urllib3/urllib3/actions/runs/3877342346/jobs/6612256458)):

```
Fatal Python error: Segmentation fault

Thread 0x00007fe6c2ffe700 (most recent call first):
  File "/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/selectors.py", line 468 in select
  File "/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/asyncio/base_events.py", line 1823 in _run_once
  File "/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/asyncio/base_events.py", line 570 in run_forever
  File "/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/asyncio/base_events.py", line 603 in run_until_complete
  File "/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/asyncio/runners.py", line 44 in run
  File "/home/runner/work/urllib3/urllib3/dummyserver/server.py", line 241 in _run_and_close_tornado
  File "/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/concurrent/futures/thread.py", line 57 in run
  File "/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/concurrent/futures/thread.py", line 80 in _worker
  File "/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/threading.py", line 870 in run
  File "/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/threading.py", line 932 in _bootstrap_inner
  File "/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/threading.py", line 890 in _bootstrap

Current thread 0x00007fe6c95e7700 (most recent call first):
  File "/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/ssl.py", line 1101 in read
  File "/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/ssl.py", line 1226 in recv
  File "/home/runner/work/urllib3/urllib3/dummyserver/testcase.py", line 31 in consume_socket
[...]

PR #289, opened 2023-01-11
Revamp tracker locking

godlygeek:
In order to guarantee mutual exclusion between the tracker singleton being destroyed and allocation tracking that may be happening in other threads, switch from a fine grained locking model for the tracker to a coarse grained one where the same lock is used for all critical sections.

Closes: #288

### DIFF
--- a/src/memray/_memray/tracking_api.cpp
+++ b/src/memray/_memray/tracking_api.cpp
@@ -1119,7 +1119,7 @@ PyTraceFunction(
 }
 
 void
-forget_python_stack()
+Tracker::forgetPythonStack()
 {
     if (!Tracker::isActive()) {
         return;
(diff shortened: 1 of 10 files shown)


src/memray/_memray/tracking_api.cpp
```diff
@@ -782,21 +799,23 @@ Tracker::computeMainTidSkip()
     return num_frames - 1;
 }
 
-void
-Tracker::trackAllocationImpl(void* ptr, size_t size, hooks::Allocator func)
+void __attribute__((noinline)) Tracker::fillNativeTraceIfNeeded(NativeTrace& trace)
 {
-    if (RecursionGuard::isActive || !Tracker::isActive()) {
-        return;
+    if (PythonStackTracker::s_native_tracking_enabled) {
```
pablogsal, line 804: We should bubble up this conditional to make the call go away

pablogsal: I have modify this by adding a static function to the tracker that acts as a proxy to `PythonStackTracker::s_native_tracking_enabled` so we can move the conditional up. That will also help with #294 

The main difference is that the compiler can indeed inline that call now

pablogsal (approved):
LGTM

PR #290, opened 2023-01-12
Manage the Tracker's lifetime with `shared_ptr`

godlygeek:
Whenever any hook needs to make use of the Tracker singleton, have it atomically grab a reference to a `shared_ptr` singleton. When stopping tracking, unset that singleton, and then wait for all of the remaining references to it to be destroyed before really destroying the Tracker.

### DIFF
--- a/src/memray/_memray/tracking_api.cpp
+++ b/src/memray/_memray/tracking_api.cpp
@@ -197,18 +197,21 @@ PythonStackTracker::emitPendingPushesAndPops()
     }
     auto first_to_emit = it.base();
 
-    // Emit pending pops
-    if (d_num_pending_pops) {
-        Tracker::getTracker()->popFrames(d_num_pending_pops);
-        d_num_pending_pops = 0;
-    }
-
-    // Emit pending pushes
-    for (auto to_emit = first_to_emit; to_emit != d_stack->end(); ++to_emit) {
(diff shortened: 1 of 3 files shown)


pablogsal:
I did some benchmarking and this solution is 60% slower than main :(:

Main:
```
hyperfine 'memray run --trace-python-allocators -o /dev/null -m test test_list test_list test_list test_list'
Benchmark 1: memray run --trace-python-allocators -o /dev/null -m test test_list test_list test_list test_list
  Time (mean ± σ):      1.657 s ±  0.021 s    [User: 1.598 s, System: 0.062 s]
  Range (min … max):    1.640 s …  1.709 s    10 runs
```	

This PR:
```
hyperfine 'memray run --trace-python-allocators -o /dev/null -m test test_list test_list test_list test_list'
Benchmark 1: memray run --trace-python-allocators -o /dev/null -m test test_list test_list test_list test_list
  Time (mean ± σ):      2.748 s ±  0.029 s    [User: 2.682 s, System: 0.072 s]
  Range (min … max):    2.715 s …  2.806 s    10 runs
```

godlygeek:
Well, that's very sad. This looks very elegant, but obviously we can't accept that kind of overhead. It looks like it's coming from locking in the `shared_ptr` implementation. That's supposed to be cheap, but when you do something on every malloc or free, the cost adds up.

Oh well.

PR #291, opened 2023-01-12
Synchronize destruction using spinlock

pablogsal:
Do the thing with spinning things


pablogsal:
This has no impact on performance that I can measure.

godlygeek (commented):
I started to review this, and force-pushed some improvements based on some comments that we had offline (which I'll submit here for posterity). As we kept playing with this, though, we found that it's about 8% slower than what's on `main` for `test_list` from the Python test suite, so we're leaning towards abandoning this in favor of #289, which solves the same problem in a different way and is faster than what's on `main`.

pablogsal:
We are closing this in favour of #289 as this introduces a 8% slowdown on some benchmarks.

PR #294, opened 2023-01-13
Stop allocating a new vector for each unwind call

godlygeek:
Up until now, each time we would unwind the native stack we would start with an empty vector, and then reserve enough space for it to hold the largest stack we had seen so far on the current thread. Instead of starting from scratch each time, save this vector in a thread-local variable so that we can reuse it from one unwind call to the next.

Note that our `RecursionGuard` prevents us from ever trying to unwind a second time on a thread before we've finished recording the first unwound stack. Note also that we don't record stack traces for deallocations, only allocations, so it won't be a problem that our deallocation hook will be called for this thread-local vector once the thread dies.

### DIFF
--- a/src/memray/_memray/tracking_api.h
+++ b/src/memray/_memray/tracking_api.h
@@ -9,6 +9,7 @@
 #include <fstream>
 #include <iterator>
 #include <memory>
+#include <optional>
 #include <string>
 #include <thread>
 #include <unordered_set>
@@ -98,6 +99,11 @@ class NativeTrace
   public:
     using ip_t = frame_id_t;
 
+    NativeTrace(std::vector<ip_t>& data)
+    : d_data(data)
+    {
+    }
+
     auto begin() const
     {
         return std::reverse_iterator(d_data.begin() + d_skip + d_size);
@@ -116,23 +122,20 @@ class NativeTrace
     }
     __attribute__((always_inline)) inline bool fill(size_t skip)
     {
-        d_data.resize(MAX_SIZE);
-
         size_t size;
         while (true) {
 #ifdef __linux__
-            size = unw_backtrace((void**)d_data.data(), MAX_SIZE);
+            size = unw_backtrace((void**)d_data.data(), d_data.size());
 #elif defined(__APPLE__)
-            size = ::backtrace((void**)d_data.data(), MAX_SIZE);
+            size = ::backtrace((void**)d_data.data(), d_data.size());
 #else
             return 0;
 #endif
-            if (size < MAX_SIZE) {
+            if (size < d_data.size()) {
                 break;
             }
 
-            MAX_SIZE = MAX_SIZE * 2;
-            d_data.resize(MAX_SIZE);
+            d_data.resize(d_data.size() * 2);
         }
         d_size = size > skip ? size - skip : 0;
         d_skip = skip;
@@ -166,12 +169,9 @@ class NativeTrace
     }
 
   private:
-    MEMRAY_FAST_TLS static thread_local size_t MAX_SIZE;
-
-  private:
     size_t d_size = 0;
     size_t d_skip = 0;
-    std::vector<ip_t> d_data;
+    std::vector<ip_t>& d_data;
 };
 
 /**
@@ -213,10 +213,13 @@ class Tracker
         }
         RecursionGuard guard;
 
-        NativeTrace trace;
+        std::optional<NativeTrace> trace{std::nullopt};
         if (Tracker::areNativeTracesEnabled()) {
+            if (!prepareNativeTrace(trace)) {
+                return;
+            }
             // Skip the internal frames so we don't need to filter them later.
-            trace.fill(1);
+            trace.value().fill(1);
         }
 
         std::unique_lock<std::mutex> lock(*s_mutex);
@@ -224,6 +227,24 @@ class Tracker
         if (tracker) {
             tracker->trackAllocationImpl(ptr, size, func, trace);
         }
+    }
+
+    static inline bool prepareNativeTrace(std::optional<NativeTrace>& trace)
+    {
+        auto t_trace_data_ptr = static_cast<std::vector<NativeTrace::ip_t>*>(
+                pthread_getspecific(s_native_unwind_vector_key));
+        if (!t_trace_data_ptr) {
+            t_trace_data_ptr = new std::vector<NativeTrace::ip_t>();
+            if (pthread_setspecific(s_native_unwind_vector_key, t_trace_data_ptr) != 0) {
+                Tracker::deactivate();
+                std::cerr << "memray: pthread_setspecific failed" << std::endl;
+                delete t_trace_data_ptr;
+                return false;
+            }
+            t_trace_data_ptr->resize(128);
+        }
+        trace.emplace(*t_trace_data_ptr);
+        return true;
     }
 
     __attribute__((always_inline)) inline static void
@@ -324,6 +345,7 @@ class Tracker
 
     // Data members
     static std::unique_ptr<std::mutex> s_mutex;
+    static pthread_key_t s_native_unwind_vector_key;
     static std::unique_ptr<Tracker> s_instance_owner;
     static std::atomic<Tracker*> s_instance;
 
@@ -341,7 +363,11 @@ class Tracker
     static size_t computeMainTidSkip();
     frame_id_t registerFrame(const RawFrame& frame);
 
-    void trackAllocationImpl(void* ptr, size_t size, hooks::Allocator func, const NativeTrace& trace);
+    void trackAllocationImpl(
+            void* ptr,
+            size_t size,
+            hooks::Allocator func,
+            const std::optional<NativeTrace>& trace);
     void trackDeallocationImpl(void* ptr, size_t size, hooks::Allocator func);
     void invalidate_module_cache_impl();
     void updateModuleCacheImpl();
--- a/tests/integration/test_extensions.py
+++ b/tests/integration/test_extensions.py
@@ -271,8 +271,52 @@ def test_valloc_at_thread_exit(tmpdir, monkeypatch):
         ctx.setattr(sys, "path", [*sys.path, str(extension_path)])
         from testext import run_valloc_at_exit  # type: ignore
 
-        with Tracker(output):
+        with Tracker(output, native_traces=True):
             run_valloc_at_exit()
+
+    # THEN
+    records = list(FileReader(output).get_allocation_records())
+    assert records
+
+    vallocs = [record for record in records if record.allocator == AllocatorType.VALLOC]
+    assert len(vallocs) == 1
+
+
+def test_valloc_at_thread_exit_in_subprocess(tmpdir, monkeypatch):
+    """Test tracking allocations in the destructor of a TLS variable.
+
+    Ensure that TLS variable is created before Memray is imported.
+    """
+    # GIVEN
+    output = Path(tmpdir) / "test.bin"
+    extension_name = "multithreaded_extension"
+    extension_path = tmpdir / extension_name
+    shutil.copytree(TEST_MULTITHREADED_EXTENSION, extension_path)
+    subprocess.run(
+        [sys.executable, str(extension_path / "setup.py"), "build_ext", "--inplace"],
+        check=True,
+        cwd=extension_path,
+        capture_output=True,
+    )
+
+    code = dedent(
+        f"""
+        from testext import run_valloc_at_exit
+        run_valloc_at_exit()  # First call creates the test extension TLS.
+
+        from memray import Tracker
+        with Tracker({str(output)!r}, native_traces=True):
+            run_valloc_at_exit()
+        """
+    )
+
+    # WHEN
+    with monkeypatch.context() as ctx:
+        ctx.setenv("PYTHONPATH", str(extension_path), prepend=":")
+        subprocess.run(
+            [sys.executable, "-c", code],
+            check=True,
+        )
 
     # THEN
     records = list(FileReader(output).get_allocation_records())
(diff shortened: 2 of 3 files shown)


pablogsal:
## Before

```
❯ hyperfine 'python -m memray run --trace-python-allocators --native -o /dev/null -m test test_list'
Benchmark 1: python -m memray run --trace-python-allocators --native -o /dev/null -m test test_list
  Time (mean ± σ):      3.517 s ±  0.035 s    [User: 3.414 s, System: 0.085 s]
  Range (min … max):    3.470 s …  3.602 s    10 runs
```

## After

```
❯ hyperfine 'python -m memray run --trace-python-allocators --native -o /dev/null -m test test_list'
Benchmark 1: python -m memray run --trace-python-allocators --native -o /dev/null -m test test_list
  Time (mean ± σ):      2.987 s ±  0.032 s    [User: 2.899 s, System: 0.061 s]
  Range (min … max):    2.948 s …  3.057 s    10 runs
```

So ~18% speedup!

pablogsal:
> Note also that we don't record stack traces for deallocations, only allocations, so it won't be a problem that our deallocation hook will be called for this thread-local vector once the thread dies.

I am not sure this works. A deallocation can allocate. Indeed, this happens in our tests:

https://github.com/bloomberg/memray/blob/960d4c9307ecd896b76e7a32e1b8eeae32d48906/tests/integration/multithreaded_extension/testext.cpp#L91

That would trigger an allocation on exit, which will recreate the TLS.

pablogsal:
Wow, I tested in my Linux box and this is the difference:

## Before
```
❯ hyperfine 'memray run --trace-python-allocators --native -o /dev/null -m test test_list'
Benchmark 1: memray run --trace-python-allocators --native -o /dev/null -m test test_list
  Time (mean ± σ):      5.069 s ±  0.066 s    [User: 5.020 s, System: 0.059 s]
  Range (min … max):    4.956 s …  5.201 s    10 runs
```

## After
```
❯ hyperfine 'memray run --trace-python-allocators --native -o /dev/null -m test test_list'
Benchmark 1: memray run --trace-python-allocators --native -o /dev/null -m test test_list
  Time (mean ± σ):      2.817 s ±  0.408 s    [User: 2.775 s, System: 0.048 s]
  Range (min … max):    2.471 s …  3.518 s    10 runs
```

pablogsal:
Hummmm, it seems that we need to still solve the tls problem:

```
{
   <insert_a_suppression_name_here>
   Memcheck:Leak
   match-leak-kinds: definite
   fun:_Znwm
   fun:trackAllocation
   fun:_ZN6memray9intercept6vallocEm
   fun:_ZN12_GLOBAL__N_1L15cleanup_handlerEPv
   fun:__nptl_deallocate_tsd
   fun:__nptl_deallocate_tsd
   fun:start_thread
   fun:clone
}
```

src/memray/_memray/tracking_api.h
```diff
@@ -213,10 +215,16 @@ class Tracker
         }
         RecursionGuard guard;
 
-        NativeTrace trace;
+        std::optional<NativeTrace> trace{std::nullopt};
```
godlygeek, line 218: Ooh, yeah, using `std::optional` is a good idea. I didn't think of that.

src/memray/_memray/tracking_api.h
```diff
@@ -116,23 +122,22 @@ class NativeTrace
     }
     __attribute__((always_inline)) inline bool fill(size_t skip)
     {
-        d_data.resize(MAX_SIZE);
+        d_data.resize(128);
```
godlygeek, line 125: You've changed this to call `d_data.resize()` on every call to `fill`, which defeats part of the optimization. We want to call this only in the constructor (if the vector is empty). Doing it here shrinks the vector if it was already larger than 128 items.

godlygeek:
> A deallocation can allocate.
Ugh, true.

> Hummmm, it seems that we need to still solve the tls problem
I wonder if this would fix it...

```diff
diff --git a/src/memray/_memray/tracking_api.cpp b/src/memray/_memray/tracking_api.cpp
index 34efe0a..b9b7f52 100644
--- a/src/memray/_memray/tracking_api.cpp
+++ b/src/memray/_memray/tracking_api.cpp
@@ -1144,7 +1144,7 @@ forget_python_stack()
         return;
     }

-    RecursionGuard guard;
+    RecursionGuard::isActive = true;
     PythonStackTracker::get().clear();
 }

```

pablogsal:
> I wonder if this would fix it...

Nah, the problem seems to be that we are re-creating the vector after the vector was destroyed on thread destruction. This can happen in any thread at any time while the tracker is active so I am not sure how setting the guard on `forget_python_stack` helps

godlygeek:
Mm. `forget_python_stack` is called when our profile function is uninstalled for a thread, which happens when the Python threadstate for that thread is destroyed, which happens before the thread dies. My thought was that setting the recursion guard there would let us ignore allocations that happen afterwards in that thread.

But that's wrong, anyway - we can create a new threadstate for a thread even after the last one was destroyed (that's normal for a native thread that's occasionally grabbing the GIL to make Python calls). So... bad idea all around.

pablogsal:
Seems that the monster I commited using `pthread_key` works

godlygeek:
Eh, OK: I see two ways to solve this problem. Both of them are solutions we applied to Python stack tracking at one point or another.

1. Spend an extra byte in TLS storing a bool representing whether the thread is dying (set it from the destructor of a TLS object). If that bool is set, don't re-create the TLS vector (and instead just supply a stack vector to the NativeTrace constructor for native unwinding).
2. Leveraging our current solution for Python stack tracking: create the vector of native frames when the first Python frame for a thread is entered, and destroy it when that frame is exited. That costs us no extra space in TLS, but it means that we only benefit from caching on threads that have Python stacks, so pure native threads would be slow.

Option 1 seems like a good solution to me, and shouldn't be tough to implement.

pablogsal:
> Leveraging our current solution for Python stack tracking: create the vector of native frames when the first Python frame for a thread is entered, and destroy it when that frame is exited. That costs us no extra space in TLS, but it means that we only benefit from caching on threads that have Python stacks, so pure native threads would be slow.

But we want this vector to be filled even if a thread never saw any Python frame whatsoever, no?

pablogsal:
> * Spend an extra byte in TLS storing a bool representing whether the thread is dying (set it from the destructor of a TLS object). If that bool is set, don't re-create the TLS vector (and instead just supply a stack vector to the NativeTrace constructor for native unwinding).

I think this may look uglier than the pthread thing, but I am not sure.

godlygeek:
> But we want this vector to be filled even if a thread never saw any Python frame whatsoever, no?

Option 2 means we'd get the caching benefit only for Python threads and not for native threads, yeah. Not ideal, but that probably gets us 90% of the benefit of this change. But option 1 gets us 100% of the benefit, at the cost of only 1 more byte of TLS, so that seems better.

> Seems that the monster I commited using `pthread_key` works

I don't understand why... unless... are you sure that you didn't just change the order, so that `key_destructor` is running _after_ `valloc_on_thread_exit`? I'm pretty sure that's what happened:

```text
tests/integration/test_extensions.py In cleanup_handler
in key_destructor
.

=== 1 passed, 448 deselected in 0.73s ===
```

pablogsal:
> Option 2 means we'd get the caching benefit only for Python threads and not for native threads, yeah. Not ideal, but that probably gets us 90% of the benefit of this change. But option 1 gets us 100% of the benefit, at the cost of only 1 more byte of TLS, so that seems better.
> 
> >

I think I have implemented a naive version of this idea correctly in the last commit, but it uses two new TLSs

pablogsal:
Ah, and also doesn't work:

```
==6807== 24 bytes in 1 blocks are definitely lost in loss record 117 of 10,110
==6807==    at 0x4849013: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==6807==    by 0x9BE43A6: trackAllocation (tracking_api.h:233)
==6807==    by 0x9BE43A6: memray::intercept::valloc(unsigned long) (hooks.cpp:254)
==6807==    by 0xB34F474: (anonymous namespace)::cleanup_handler(void*) (testext.cpp:61)
==6807==    by 0x4CAA710: __nptl_deallocate_tsd (nptl_deallocate_tsd.c:73)
==6807==    by 0x4CAA710: __nptl_deallocate_tsd (nptl_deallocate_tsd.c:22)
==6807==    by 0x4CAD9C9: start_thread (pthread_create.c:453)
==6807==    by 0x4D3EBB3: clone (clone.S:100)
```

src/memray/_memray/tracking_api.h
```diff
@@ -58,6 +58,15 @@ struct RecursionGuard
 
     const bool wasLocked;
     MEMRAY_FAST_TLS static thread_local bool isActive;
+    MEMRAY_FAST_TLS static thread_local bool threadDestroyed;
+};
+
+struct ThreadDestroyedGuard
+{
+    ThreadDestroyedGuard()
```
godlygeek, line 66: You're missing a `~` here 😅

pablogsal: :eyes: Well, time to stop today :sweat:

pablogsal: (Still doesn't work)

godlygeek:
It mostly works - it fails if the _first_ call to `trackAllocation` in some thread happens after that thread has already started to exit. As long as something touches the TLS variable before the thread starts to exit, everything is hunky dory, but if the first access happens while TLS is being torn down, then the guard doesn't work: the destructor wasn't scheduled to run at thread teardown because the constructor never ran for that thread.

Ugh.

godlygeek:
Maybe using the raw pthread APIs does work though. The docs have this snippet in them:

> Both  pthread_getspecific() and pthread_setspecific() may be called from a thread-specific data destructor function. A call to pthread_getspecific() for the thread-specific data key being destroyed shall return the value NULL, unless the value is changed (after the destructor starts) by a call to pthread_setspecific(). Calling pthread_setspecific() from a thread-specific data destructor routine may result either in lost storage (after at least PTHREAD_DESTRUCTOR_ITERATIONS attempts at destruction) or in an infinite loop.

That bit about `PTHREAD_DESTRUCTOR_ITERATIONS` seems to imply that, when using the raw pthread APIs rather than the C++ stdlib interface, creating the variable after the thread has started being torn down is OK: it will result in a second iteration of the destructors, and that one will destroy the value created in the first iteration.

So: I think your test case didn't correctly prove that using the raw pthread APIs works, because it just wound up calling the destructors in a different order than the test expected - _but_ it seems like it might indeed work, and we just need a different test case to prove it.

godlygeek:
And https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html seems to say that `PTHREAD_DESTRUCTOR_ITERATIONS` must be at least 4.

godlygeek:
This means our TLS `PythonStackTracker` may be wrong, too. There's nothing preventing a TLS variable's destructor function from acquiring the GIL and calling a Python function - which means that it's possible for a Python stack to get recorded by another TLS variable's destructor, after our TLS Python stack has been destroyed. 😬

godlygeek:
> There's nothing preventing a TLS variable's destructor function from acquiring the GIL and calling a Python function

We've decided this probably isn't true. Acquiring the GIL using the `PyGILState_Ensure` API makes use of a thread local variable inside of the Python interpreter, and relying on that TLS variable after TLS variables have already begun to be destroyed is no bueno. So, this case doesn't seem worth worrying about.

src/memray/_memray/tracking_api.h
```diff
@@ -213,10 +213,22 @@ class Tracker
...
 
-        NativeTrace trace;
+        std::optional<NativeTrace> trace{std::nullopt};
         if (Tracker::areNativeTracesEnabled()) {
+            auto t_trace_data_ptr = static_cast<std::vector<NativeTrace::ip_t>*>(
+                    pthread_getspecific(s_native_unwind_vector_key));
+            if (!t_trace_data_ptr) {
+                t_trace_data_ptr = new std::vector<NativeTrace::ip_t>();
+                if (pthread_setspecific(s_native_unwind_vector_key, t_trace_data_ptr) != 0) {
+                    fprintf(stderr, "memray: pthread_setspecific failed\n");
```
pablogsal, line 223: Should we deactivate tracking here?

tests/integration/test_extensions.py
```diff
@@ -282,6 +282,50 @@ def test_valloc_at_thread_exit(tmpdir, monkeypatch):
     assert len(vallocs) == 1
 
 
+def test_valloc_at_thread_exit_in_subprocess(tmpdir, monkeypatch):
```
pablogsal, line 285: Let's decorate this with the valgrind marker

src/memray/_memray/tracking_api.h
godlygeek: Sure, I suppose. I think there's very few ways this could happen:
1. We've got a bug that causes us to fail to create the key before we use it
2. We (or the program we're tracking) has a memory corruption bug that trashes our key
3. We've gotten an ENOMEM trying to allocate TLS for a new thread

I think any of those are pretty fantastically unlikely, but I suppose none of them are really recoverable, so deactivating tracking is fine. (Technically we could recover from (1) by falling back to a new vector for each call, but it doesn't seem worth the effort, given that it would represent some bug in our code.)

tests/integration/test_extensions.py
godlygeek: It wouldn't help, as the meat of it is run in a subprocess, and we don't run valgrind with `--trace-children=yes`.

It's run in a subprocess so that we can control the order in which the different TLS variables get created - the non-subprocess one will always have done an `import memray` first, so the subprocess one is for checking the case where an extension with a TLS destructor that allocates memory creates its TLS variable before Memray has created its.

godlygeek: That said - maybe we _should_ use `--trace-children=yes` for Valgrind?

pablogsal (approved):

tests/integration/test_extensions.py
pablogsal: Maybe in a different PR.

PR #299, opened 2023-01-18
Use consistently binary multiples in the reporters

pablogsal:
We are inconsistently using decimal and binary multiples in different
places where we report sizes.


godlygeek (approved):
LGTM

PR #301, opened 2023-01-19
Update example flamegraphs in the docs

pablogsal:
We have changed quite a lot the flamegraph reporter since the last time
these examples were rendered.


godlygeek (approved):
LGTM

PR #302, opened 2023-01-19
Add a GitHub action to check for changelog files

pablogsal:
We constantly forget to add news entries in Pull Requests and we need to
fix this after the Pull Requests are landed. With this action we can
easily automate the check so we can consciously decide to ignore if
needed.

### DIFF
--- a/.github/workflows/sanity-check.yml
+++ b/.github/workflows/sanity-check.yml
@@ -1,5 +1,12 @@
 name: Sanity check
-on: [pull_request]
+on: 
+  pull_request:
+    types:
+    - "opened"
+    - "reopened"
+    - "synchronize"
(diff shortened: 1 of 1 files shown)


.github/workflows/sanity-check.yml
```diff
@@ -1,5 +1,12 @@
 name: Sanity check
-on: [pull_request]
+on: 
+  pull_request:
+    types:
+    - "opened"
+    - "reopened"
+    - "synchronize"
+    - "labeled"
+    - "unlabeled"
```
godlygeek, line 9: What's the difference between the old and the new here? Are there some types that it used to run on but now won't?

godlygeek (approved):
LGTM

Issue #253, opened 2022-11-15
Document trace_python_allocators in the API docs


pablogsal:
We are missing adding some explanation of what `trace_python_allocators` does in the API docs

PR #303, opened 2023-01-19
Document ``trace_python_allocators`` in the memray API docs

pablogsal:
We allow users to pass this value but we currently do not include it in
our public API documentation.

Closes: #253


godlygeek (approved):
LGTM

PR #308, opened 2023-01-20
Add a new patch to libbacktrace to integrate with debuginfod

pablogsal:
- Add a new patch to libbacktrace to integrate with debuginfod
- Update the vendored libbacktrace files with the latest patch

### DIFF
--- a/src/vendor/libbacktrace/elf.c
+++ b/src/vendor/libbacktrace/elf.c
@@ -38,6 +38,7 @@ POSSIBILITY OF SUCH DAMAGE.  */
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <unistd.h>
+#include <dlfcn.h>
 
 #ifdef HAVE_DL_ITERATE_PHDR
 #include <link.h>
@@ -45,6 +46,7 @@ POSSIBILITY OF SUCH DAMAGE.  */
 
 #include "backtrace.h"
 #include "internal.h"
+#include "debuginfod_support.h"
 
 #ifndef S_ISLNK
  #ifndef S_IFLNK
@@ -850,6 +852,82 @@ elf_readlink (struct backtrace_state *state, const char *filename,
 }
 
 #define SYSTEM_BUILD_ID_DIR "/usr/lib/debug/.build-id/"
+
+typedef debuginfod_client*(*debuginfod_begin_t)(void);
+
+typedef int(*debuginfod_find_debuginfo_t)(debuginfod_client *client,
+                                            const unsigned char *build_id,
+                                            int build_id_len,
+                                            char **path);
+typedef void(*debuginfod_end_t)(debuginfod_client *client);
+
+
+static debuginfod_find_debuginfo_t the_debuginfod_find_debuginfo = NULL;
+static debuginfod_begin_t the_debuginfod_begin = NULL;
+static debuginfod_end_t the_debuginfod_end = NULL;
+static int debuginfod_guard = 0;
+static int debuginfod_dlopen_failed = 0;
+
+static int
+elf_open_debugfile_by_debuginfod (const char *buildid_data,
+			       size_t buildid_size,
+			       backtrace_error_callback error_callback,
+			       void *data)
+{
+  if (debuginfod_dlopen_failed) {
+      return -1;
+  }
+
+  if (the_debuginfod_find_debuginfo == NULL) {
+    void *handle = dlopen(DEBUGINFOD_SONAME, RTLD_LAZY);
+    if (!handle) {
+      debuginfod_dlopen_failed = 1;
+      return -1;
+    }
+
+    the_debuginfod_find_debuginfo = (debuginfod_find_debuginfo_t)dlsym(handle, "debuginfod_find_debuginfo");
+    if (!the_debuginfod_find_debuginfo) {
+      return -1;
+    }
+
+    the_debuginfod_begin = (debuginfod_begin_t)dlsym(handle, "debuginfod_begin");
+    if (!the_debuginfod_begin) {
+      the_debuginfod_find_debuginfo = NULL;
+      return -1;
+    }
+
+    the_debuginfod_end = (debuginfod_end_t)dlsym(handle, "debuginfod_end");
+    if (!the_debuginfod_end) {
+      the_debuginfod_find_debuginfo = NULL;
+      the_debuginfod_begin = NULL;
+      return -1;
+    }
+  }
+
+  char* path = NULL;
+  const unsigned char* bi = (const unsigned char*)buildid_data;
+
+  debuginfod_client* client = the_debuginfod_begin();
+  if (!client) {
+    return -1;
+  }
+
+  const int fd = the_debuginfod_find_debuginfo(client, bi, buildid_size, &path);
+
+  the_debuginfod_end(client);
+
+  if (!path || fd <= 0) {
+    return -1;
+  }
+
+  const int ret = backtrace_open (path, error_callback, data, NULL);
+
+  free(path);
+  close(fd);
+
+  return ret;
+}
+
 
 /* Open a separate debug info file, using the build ID to find it.
    Returns an open file descriptor, or -1.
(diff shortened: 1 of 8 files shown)


src/vendor/libbacktrace/elf.c
```diff
@@ -30,6 +30,8 @@ STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
 IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 POSSIBILITY OF SUCH DAMAGE.  */
 
+#define _GNU_SOURCE
```
godlygeek, line 33: 👀 What's this needed for?

src/vendor/libbacktrace_2446c66076480ce07a6bd868badcbceb3eeecc2e_debuginfod_patch.diff
```diff
@@ -0,0 +1,331 @@
...
++			       void *data)
++{
++  int ret;
++  int does_not_exist;
++
++  if (debuginfod_dlopen_failed) {
++      return -1;
++  }
++
++  if (the_debuginfod_find_debuginfo == NULL) {
```
godlygeek, line 246: Defensive programming nit: check for the last one to be assigned, instead of the first, so that if we enter this function with `the_debuginfod_find_debuginfo` set (because `dlopen` and `dlsym("debuginfod_find_debuginfo")` succeed but `dlsym("debuginfod_end")` fails, we don't skip this block and try calling a null pointer.
```suggestion
+  if (the_debuginfod_end == NULL) {
```

Or alternatively, set any that have been assigned to `nullptr` on error. That's perhaps more readable, though this suggestion is more succinct.

src/vendor/libbacktrace_2446c66076480ce07a6bd868badcbceb3eeecc2e_debuginfod_patch.diff
```diff
@@ -0,0 +1,331 @@
...
++    void *handle = dlopen(DEBUGINFOD_SONAME, RTLD_LAZY);
```
godlygeek, line 247: If `dlopen` succeeds but `dlsym` fails (shouldn't happen, but who knows), then every call to this function to resolve symbols will make a new call to `dlopen`.

I see that you've architected it to leak the handle even in the success case, and I guess that seems fair enough (we could clean it up in a `__attribute__(destructor)` or whatever, but it's probably not worth the trouble).

But: do we leak any resources in the case where `dlopen` is called repeatedly and succeeds every time? I _think_ no, that it'll keep returning the same handle on each call without allocating any new resources, but double check me on that.

src/vendor/libbacktrace_2446c66076480ce07a6bd868badcbceb3eeecc2e_debuginfod_patch.diff
```diff
@@ -0,0 +1,331 @@
...
++static int debuginfod_dlopen_failed = 0;
++ 
++static int
++elf_open_debugfile_by_debuginfod (const char *buildid_data,
++			       size_t buildid_size,
++			       backtrace_error_callback error_callback,
++			       void *data)
++{
++  int ret;
++  int does_not_exist;
```
godlygeek, line 240: Minor nit, but these variables could be declared closer to where they're first used.

That would also let you make `ret` const. I know you love those tasty consts.

src/vendor/libbacktrace_2446c66076480ce07a6bd868badcbceb3eeecc2e_debuginfod_patch.diff
```diff
@@ -0,0 +1,331 @@
...
++
++  int fd = the_debuginfod_find_debuginfo(client, bi, buildid_size, &path);
++
++  the_debuginfod_end(client);
++
++  if (!path || fd <= 0) {
++    return -1;
++  } 
++
++  ret = backtrace_open (path, error_callback, data, &does_not_exist);
```
godlygeek, line 285: > /* Open a file for reading.  Returns -1 on error.  If DOES_NOT_EXIST
>   is not NULL, *DOES_NOT_EXIST will be set to 0 normally and set to 1
>   if the file does not exist.  If the file does not exist and

Any reason not to just pass a nullptr instead of `&does_not_exist`, if we're not checking the value?

src/vendor/libbacktrace_2446c66076480ce07a6bd868badcbceb3eeecc2e_debuginfod_patch.diff
```diff
@@ -0,0 +1,331 @@
...
++
++  free(path);
```
godlygeek, line 287: The docs say:
> If a find family function is successful, the resulting file is saved
> to the client cache and a file descriptor to that file is returned.
> The caller needs to `close()` this descriptor.  Otherwise, a
> negative error code is returned.

I think we need:
```suggestion
+  free(path);
+  close(fd);
```

godlygeek (changes requested):

godlygeek:
Oh, I probably shouldn't have made suggestions on the patch - I was thinking that was the source of truth and it'd be easier to apply them there and then regenerate the code, which is true, but of course you need to also fix up the line numbers in the patch because of the new/removed lines. Oops.

src/vendor/libbacktrace_2446c66076480ce07a6bd868badcbceb3eeecc2e_debuginfod_patch.diff
pablogsal: Tasty tasty consts....:)

src/vendor/libbacktrace_2446c66076480ce07a6bd868badcbceb3eeecc2e_debuginfod_patch.diff
pablogsal: >But: do we leak any resources in the case where dlopen is called repeatedly and succeeds every time? I think no, that it'll keep returning the same handle on each call without allocating any new resources, but double check me on that.

Yes. The [dlopen(3)](http://linux.die.net/man/3/dlopen) linux man page says:

 >  If the same library is loaded again with dlopen(), the same file 
   handle is returned. The dl library maintains reference counts for 
   library handles, so a dynamic library is not deallocated until 
   dlclose() has been called on it as many times as dlopen() 
   has succeeded on it.

As we are not `dlclosing` this, I think this is fine.

src/vendor/libbacktrace_2446c66076480ce07a6bd868badcbceb3eeecc2e_debuginfod_patch.diff
pablogsal: No, will clean this

src/vendor/libbacktrace_2446c66076480ce07a6bd868badcbceb3eeecc2e_debuginfod_patch.diff
```diff
@@ -0,0 +1,331 @@
...
+-
+ 	  elf_release_view (state, &buildid_view, error_callback, data);
+ 	  if (debuglink_view_valid)
+ 	    elf_release_view (state, &debuglink_view, error_callback, data);
+@@ -4889,7 +4974,10 @@ backtrace_initialize (struct backtrace_state *state, const char *filename,
+   pd.exe_filename = filename;
+   pd.exe_descriptor = ret < 0 ? descriptor : -1;
+ 
++   
++  debuginfod_guard++;
```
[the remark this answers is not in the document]
pablogsal: Done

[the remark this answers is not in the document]
godlygeek: No, I'm happy enough with this note:

> Note this ONLY affects the symbol resolution when retrieving a backtrace,
> and it doesn't affect offline symbolication.

Since that suggests the solution we'll take if we start getting reports that users find this confusing.

godlygeek (approved):
While testing this locally I noticed `patch` was complaining about trailing whitespace on a few of the lines in the patch. I've force pushed a fix for that.

Otherwise, LGTM!

PR #313, opened 2023-02-04
Use the vanilla Alpine image instead of the python-based one

pablogsal:
The Python-based alpine image downloads and builds a Python executable
without symbols. This causes several tests to fail because we cannot
leverage the `python3-dbg`package that contains symbols for the
interpreter.

To fix this, use the vanilla Alpine docker image and also install the
`python3-dbg` package so we always have symbols available.


godlygeek (approved):
LGTM

PR #314, opened 2023-02-06
Reduce the file size of flame graphs

godlygeek:
These two changes combined result in a flame graph that's 85% smaller for the nbody example program in `--trace-python-allocators` and `--native` mode.


src/memray/reporters/flamegraph.py
```diff
@@ -126,26 +139,51 @@ def from_snapshot(
...
+            nodes["lineno"].append(frame["location"][2])
+            nodes["value"].append(frame["value"])
+            nodes["children"].append(frame["children"])
+            nodes["n_allocations"].append(frame["n_allocations"])
+            nodes["thread_id"].append(all_strings.register(frame["thread_id"]))
+            nodes["interesting"].append(int(frame["interesting"]))
+            nodes["import_system"].append(int(frame["import_system"]))
+
+        data = {
+            "unique_threads": [all_strings.register(t) for t in sorted(unique_threads)],
```
pablogsal, line 181: Nit:

```suggestion
            "unique_threads": tuple(all_strings.register(t) for t in sorted(unique_threads)),
```

src/memray/reporters/templates/__init__.py
```diff
@@ -13,9 +13,11 @@
 
 @lru_cache(maxsize=1)
 def get_render_environment() -> jinja2.Environment:
-    return jinja2.Environment(
+    env = jinja2.Environment(
         loader=jinja2.PackageLoader("memray.reporters"),
     )
+    env.policies["json.dumps_kwargs"] = {"sort_keys": True, "separators": (",", ":")}
```
pablogsal, line 19: jesus

pablogsal (approved):
LGTM

This is fantastic!

I left two nits

src/memray/reporters/templates/__init__.py
godlygeek: This change alone saves us about 2 MB on the size of the nbody example program's flamegraph - apparently we had 2 MB of useless spaces! 😅

src/memray/reporters/assets/flamegraph.js
```diff
@@ -15,8 +15,42 @@ import {
...
+    node["interesting"] = nodes["interesting"][i] != 0;
+    node["import_system"] = nodes["import_system"][i] != 0;
+  }
+
+  let root = node_objects[0];
+  root["unique_threads"] = unique_threads.map(
+    (thread_id) => strings[thread_id]
+  );
+  return root;
+}
```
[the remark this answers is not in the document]
godlygeek: > I think this does the same and is simpler but feel free to ignore or pick just the parts that you like:

Thanks, I learned some more JavaScript from this - I'd never seen the destructuring assignment stuff before (I'm pretty sure it didn't exist the last time I did much JS 😄).

I took most of this. I tweaked it a little bit, though. We need to maintain two copies of this function, one in Python and one in JS (since the flame graph tests need to undo our packing as well, in order to make assertions about the structure of the tree). So, I tried to make it as easy as possible to copy/paste things from JS to Python (or vice versa). I kept the same name for the local variables in both language (rather than using camel case in one and snake case in the other). I also used `obj["property"]` instead of `obj.property` for accessing object properties in the JS code, to make it easier to copy-paste into Python code where we're representing these objects as a dictionary. That makes for less idiomatic JavaScript, but easier copy-pasting.

PR #317, opened 2023-02-09
Temporal flamegraphs

godlygeek:
I'm leaving off the news and docs for now, as I'd still like to do some more refactoring here (possibly putting this under a new subcommand, rather than putting it under a flag for the existing flamegraph reporter).

I should probably add some automated tests as well... but other than that, this should be in a reviewable state.


godlygeek:
@pablogsal Since you've started looking at this, I pushed one commit as a fixup that corrects some minor issues noticed while testing, rather than force pushing fixes to code you've already started reviewing.

We'll need to make sure to apply the fixup before landing this.

Anyway, there's tests now! 🥳

pablogsal:
> Anyway, there's tests now! 🥳

Ah, but someone forgot to signup the commits :P

src/memray/reporters/assets/temporal_flamegraph.js
```diff
@@ -0,0 +1,299 @@
...
+      allocBefore <= rangeEnd &&
+      (deallocBefore === null || deallocBefore > rangeEnd)
+    ) {
+      while (nodeIndex !== undefined) {
+        node_objects[nodeIndex].n_allocations += count;
+        node_objects[nodeIndex].value += bytes;
+        nodeIndex = parent_index_by_child_index[nodeIndex];
+      }
+    }
+  });
```
pablogsal, line 70: A possible implementation of this using binary search (not tested):

```
console.log("finding leaked allocations");

packedData.intervals.forEach((interval) => {
  let [allocBefore, deallocBefore, nodeIndex, count, bytes] = interval;

  if (
    allocBefore >= rangeStart &&
    allocBefore <= rangeEnd &&
    (deallocBefore === null || deallocBefore > rangeEnd)
  ) {
    // Use binary search to find the parent index of the node
    let low = 0;
    let high = parent_index_by_child_index.length - 1;
    while (low <= high) {
      let mid = Math.floor((low + high) / 2);
      if (nodeIndex > parent_index_by_child_index[mid]) {
        low = mid + 1;
      } else if (nodeIndex < parent_index_by_child_index[mid]) {
        high = mid - 1;
      } else {
        nodeIndex = mid;
        break;
      }
    }

    // Update the node_objects for each parent node
    while (nodeIndex !== undefined) {
      node_objects[nodeIndex].n_allocations += count;
      node_objects[nodeIndex].value += bytes;
      nodeIndex = parent_index_by_child_index[nodeIndex];
    }
  }
});
```

pablogsal: To be honest it looks more complex than the linear search and I don't think it gives us that much 🤷

src/memray/reporters/assets/temporal_flamegraph.js
```diff
@@ -0,0 +1,299 @@
...
+
+    idx1 = 0;
+    let t1 = new Date(request_data.string2).getTime();
+    for (let i = memory_records.length - 1; i >= 1; i--) {
+      if (memory_records[i - 1][0] < t1) {
+        idx1 = i;
+        break;
+      }
+    }
+  }
```
pablogsal, line 186: What about

```
let idx0 = 0;
let idx1 = memory_records.length - 1;

if (request_data) {
  const t0 = new Date(request_data.string1).getTime();
  idx0 = memory_records.findIndex(([t, _]) => t >= t0);

  const t1 = new Date(request_data.string2).getTime();
  idx1 = memory_records.length - 1 - memory_records.slice().reverse().findIndex(([t, _]) => t < t1);
}

idx1 = Math.max(idx1, idx0);
```

?

src/memray/reporters/assets/temporal_flamegraph.js
```diff
@@ -0,0 +1,299 @@
...
+    let the_range = active_plot.layout.xaxis.range;
+    request_data = {
+      string1: the_range[0],
+      string2: the_range[1],
+    };
+  } else {
+    return;
+  }
+  return request_data;
+}
```
pablogsal, line 225: I was trying to simplify this and came out with this:

```
function getRangeData(event) {
  const range = event?.['xaxis.range'] || event?.['xaxis.range[0]'];
  if (!range && !active_plot) return;
  const [string1, string2] = Array.isArray(range) ? range : [range[0], range[1]];
  return { string1, string2 };
}
```

Unclear if is better, though. Is certainly shorter :)

src/memray/_memray/snapshot.cpp
```diff
@@ -19,6 +20,18 @@ HighWaterMarkLocationKey::operator==(const HighWaterMarkLocationKey& rhs) const
...
+operator<(const AllocationLifetime& lhs, const AllocationLifetime& rhs)
+{
+    // Sort by n_bytes if allocatedBefore/deallocatedBefore are equal,
+    // so that our test suite gets records in a predictable order.
+    return lhs.allocatedBeforeSnapshot != rhs.allocatedBeforeSnapshot
+                   ? lhs.allocatedBeforeSnapshot < rhs.allocatedBeforeSnapshot
+           : lhs.deallocatedBeforeSnapshot != rhs.deallocatedBeforeSnapshot
+                   ? lhs.deallocatedBeforeSnapshot < rhs.deallocatedBeforeSnapshot
+                   : lhs.n_bytes < rhs.n_bytes;
+}
```
pablogsal, line 33: I find the chain of ternary stuff a bit difficult to read. What about something like:

```
bool operator<(const AllocationLifetime& lhs, const AllocationLifetime& rhs) {
  if (lhs.allocatedBeforeSnapshot != rhs.allocatedBeforeSnapshot) {
    return lhs.allocatedBeforeSnapshot < rhs.allocatedBeforeSnapshot;
  } else if (lhs.deallocatedBeforeSnapshot != rhs.deallocatedBeforeSnapshot) {
    return lhs.deallocatedBeforeSnapshot < rhs.deallocatedBeforeSnapshot;
  } else {
    return std::tie(lhs.n_bytes) < std::tie(rhs.n_bytes);
  }
}
```

src/memray/reporters/assets/temporal_flamegraph.js
godlygeek: Hm. We could pull in d3-array and use https://github.com/d3/d3-array/blob/main/README.md#bisectLeft perhaps.

The linear search is fast enough using the samples I've tried with, but I'm not sure if that would be the case with a longer run or a huge capture file.

src/memray/reporters/assets/temporal_flamegraph.js
godlygeek: Hm, not opposed to using `findIndex`, but I think that should be something more like:
```js
let idx0 = 0;
let idx1 = memory_records.length - 1;

if (request_data) {
  const t0 = new Date(request_data.string1).getTime();
  const t0_idx = memory_records.findLastIndex((rec) => rec[0] >= t0);
  if (t0_idx != -1) idx0 = t0_idx;

  const t1 = new Date(request_data.string2).getTime();
  const t1_idx = memory_records.findIndex((rec) => rec[0] >= t1);
  if (t1_idx != -1) idx1 = t1_idx;
}
```

src/memray/reporters/assets/temporal_flamegraph.js
godlygeek: Well, this code came verbatim from your prototype - I'm a bit baffled by why plotly gives us the data in such an inconsistent format in the first place.

src/memray/_memray/snapshot.h
```diff
@@ -320,6 +320,64 @@ class HighWaterMarkAggregator
     reduced_snapshot_map_t getAllocations(bool merge_threads, bool stop_at_high_water_mark) const;
 };
 
+struct AllocationLifetime
+{
+    size_t allocatedBeforeSnapshot;
+    size_t deallocatedBeforeSnapshot;  // SIZE_MAX if never deallocated
+    HighWaterMarkLocationKey key;
```
pablogsal, line 327: Now that I am reviewing this, `HighWaterMarkLocationKey` is now a slightly confusing name. I opened an issue about the whole struct:

https://github.com/bloomberg/memray/issues/321

src/memray/_memray/snapshot.cpp
```diff
@@ -407,6 +420,163 @@ HighWaterMarkAggregator::visitAllocations(const allocation_callback_t& callback)
...
+    std::vector<AllocationLifetime> ret;
+
+    // Things that were leaked
+    for (const auto& [when_where, how_much] : leaks) {
+        const auto& [allocated_before, key] = when_where;
+        const auto& [n_allocations, n_bytes] = how_much;
+        ret.push_back({allocated_before, static_cast<size_t>(-1), key, n_allocations, n_bytes});
+    }
+
+    // Things that weren't leaked
```
pablogsal, line 569: I think we can discuss this offline, but I wonder if we can somehow handle true leaks in a closer fashion as the rest by just inserting them in `d_allocation_history` with `SIZE_MAX` or something in the `deallocatedBeforeSnapshot` field. Not that it really changes things a lot, but when reviewing this I kept wondering what makes true leaks so special with respect of the data model we are constructing here

src/memray/_memray/snapshot.cpp
```diff
@@ -407,6 +420,163 @@ HighWaterMarkAggregator::visitAllocations(const allocation_callback_t& callback)
...
+}
+
+void
+AllocationLifetimeAggregator::captureSnapshot()
+{
+    ++d_num_snapshots;
+}
+
+std::vector<AllocationLifetime>
+AllocationLifetimeAggregator::generateIndex()
```
pablogsal, line 524: Can we add a small comment with the general recipe for this? I think is clear enough without comments here but the fact that we have some like

```
// Leaked simple allocations
```

makes it seems the reader should now already what the process is going to be and is a bit confusing the first time you go through this. WDYT?

src/memray/_memray/snapshot.cpp
godlygeek: Discussed offline: we can't do that because the interface doesn't expose a way to signal "I'm done adding allocations". We could implicitly make the first call to `generateIndex` also mark that no more allocations can be added and merge all of the leaks into `d_allocation_history`, but that feels very weird. After our offline discussion, we decided to keep `generateIndex` side effect free.

src/memray/_memray/snapshot.cpp
godlygeek: I tried to improve the comments here. Take a look and see if you think there's anything that still needs to be explained better.

src/memray/_memray.pyx
```diff
@@ -124,18 +124,138 @@ def size_fmt(num, suffix='B'):
...
+
+    return hybrid_stack[:max_stacks]
+
 
 @cython.freelist(1024)
 cdef class AllocationRecord:
     cdef object _tuple
-    cdef object _stack_trace
-    cdef object _native_stack_trace
+    cdef object _stack_trace_cache
```
pablogsal, line 253: IIRC you may be able to use `cdef dict _stack_trace_cache` and cython will improve access code but I am not 100% sure that the last part is true.

pablogsal:
One thing I noticed, we should change the title in the flamegraph because right now it shows:

Screenshot 2023-02-23 at 14 03 39

src/memray/_memray.pyx
```diff
@@ -1111,3 +1419,37 @@ cdef class HighWaterMarkAggregatorTestHarness:
...
+        allocation.n_allocations = 1
+        self.aggregator.addAllocation(allocation)
+
+    def capture_snapshot(self):
+        return self.aggregator.captureSnapshot()
+
+    def get_allocations(self):
+        cdef shared_ptr[RecordReader] reader
+        cdef TemporalAllocationGenerator gen = TemporalAllocationGenerator()
+        gen.setup(self.aggregator.generateIndex(), reader)
```
pablogsal, line 1454: Why not using a `__init__/__new__` for this?

godlygeek: Because we can't pass C++ objects to `__init__` or `__cinit__` - every argument to `__init__` or `__new__` needs to be convertible to a Python object, and we don't want to do that with either of these.

Although... I was trying to avoid making a copy of this big vector, but it looks like Cython has thrown it into a local temporary anyway. I think I'll switch to a `shared_ptr` for both the index and the reader, to avoid making that useless copy.

src/memray/_memray.pyx
godlygeek: Yeah, looks like it actually does generate different code for that. I'll make that change.

src/memray/_memray.pyx
godlygeek: No, making it a smart pointer makes the code much uglier. Looks like I can just make the function take an r-value reference, and std::move the value out of Cython's temporary. Much less ugly.

godlygeek:
The force push I just did should address the new concerns you raised, plus some minor things I noticed: the unnecessary vector copy, our help button text not mentioning the sliders, our title not including "(memory leaks)", and the warning about generating the report without tracking Python allocators not showing up.

godlygeek:
Whoops, one more force push - I missed updating `TemporalAllocationRecord` to not be a subclass of `AllocationRecord` in the extension module's type stubs.

godlygeek:
> Whoops, one more force push - I missed updating `TemporalAllocationRecord` to not be a subclass of `AllocationRecord` in the extension module's type stubs.

Which also surfaced another type warning. I just shut it up with a `# type: ignore`, since that's code that I plan on refactoring as soon as we land this PR.

godlygeek:
And then, just for fun, I rebased on `main`. 😂

pablogsal (approved):

PR #325, opened 2023-02-24
Refactor `HighWatermarkCommand`

godlygeek:
This PR implements the cleanup of the `HighWatermarkCommand` interface that I promised when we were discussing ways to clean up #317 to pass the `TemporalAllocationRecord`'s through the layers more cleanly.

I think this is a huge improvement over what we've got. This is in a reviewable state, I'm only marking it as draft to remind us not to land it until we've landed #317 and rebased this on it.


godlygeek:
Rebased, now that #317 is landed. Should be good to review at your leisure.

pablogsal:
Flawless!

pablogsal (approved):

Issue #332, opened 2023-03-11
AssertionError when generating flamegraph from native run


toofar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I'm trying to generation a flamegraph based on data from a native run. I'm getting this output:

```
⚠  No debug information was found for the Python interpreter  ⚠

Without debug information reports showing native traces may not include file names and line numbers. Please use an interpreter 
built with debug symbols for best results. Check https://bloomberg.github.io/memray/native_mode.html for more information 
regarding how memray resolves symbols.

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/mnt/ssd/user/qute-basedirs/venv-6.5/lib/python3.11/site-packages/memray/__main__.py", line 6, in <module>
    sys.exit(main())
             ^^^^^^
  File "/mnt/ssd/user/qute-basedirs/venv-6.5/lib/python3.11/site-packages/memray/commands/__init__.py", line 128, in main
    arg_values.entrypoint(arg_values, parser)
  File "/mnt/ssd/user/qute-basedirs/venv-6.5/lib/python3.11/site-packages/memray/commands/common.py", line 165, in run
    self.write_report(
  File "/mnt/ssd/user/qute-basedirs/venv-6.5/lib/python3.11/site-packages/memray/commands/common.py", line 130, in write_report
    reporter = self.reporter_factory(
               ^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/ssd/user/qute-basedirs/venv-6.5/lib/python3.11/site-packages/memray/reporters/flamegraph.py", line 121, in from_snapshot
    tuple(record.hybrid_stack_trace())
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/memray/_memray.pyx", line 279, in memray._memray.AllocationRecord.hybrid_stack_trace
AssertionError
```

The .bin file is here (50MB): https://litter.catbox.moe/g8io46.bin
It's a 3 day link, I can put up another one if that times out.

[...]

PR #334, opened 2023-03-14
Handle missing eval symbols in the hybrid stack with non entry frames

pablogsal:
In Python 3.11, due to the presence of non-entry frames we are hitting an assertion when constructing the hybrid stack.

for Python 3.10 and earlier, `len(hybrid_stack) == len(native_stack)`, which is why we've never seen this `assert hidx == -1` fail before. In 3.11 there are now non-entry frames, and `hybrid_stack` is longer than `native_stack` by the number of non-entry frames in the Python stack. If we fail to recognize a `_PyEval_EvalFrameDefault` call in Python 3.10 and earlier it only causes us to fill a native frame into `hybrid_stack` instead of a Python frame, but from 3.11 on it _also_ causes us to totally fail to insert any non-entry Python frames that were evaluated by that call to `_PyEval_EvalFrameDefault`, meaning we write fewer total entries into `hybrid_stack` than we meant to.

After some discussion we have decided that the correct thing to do here is return the Python stack if we cannot construct the hybrid stack. This is because report files that include native information will always try to construct the hybrid stack and returning something that doesn't include the Python stack somehow will result in a very bad user experience as they won't see any Python code whatsoever in the reports.

Closes: #332

### DIFF
--- a/news/334.bugfix.rst
+++ b/news/334.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a bug that was hitting an assert when constructing hybrid stack frames in Python 3.11 when no eval symbols are available.
--- a/src/memray/_memray.pyx
+++ b/src/memray/_memray.pyx
@@ -236,13 +236,13 @@ cdef hybrid_stack_trace(
             hybrid_stack[hidx] = native_frame
             hidx -= 1
 
-    assert hidx == -1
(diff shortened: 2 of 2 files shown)


godlygeek (approved):
LGTM

PR #362, opened 2023-04-24
Remove tox dependancy

### LINKED ISSUE #350: Drop `tox` dependency
Since we've landed on using `make` for orchestrating maintenance commands for the repository, remove our `tox` configuration file and drop our dependency on `tox` in our GitHub actions. Also remove the reference to `tox.ini` in our coverage configuration in `Makefile`, as well as any other references to it that exist.

sjirwin:
*Issue number of the reported bug or feature request: #350*

**Describe your changes**
`make` is now used for orchestrating maintenance commands for the repository so `tox` is not needed anymore

**Testing performed**
verified new settings by running `make dev-install` and `make pycoverage` locally

**Additional context**
Closes #350


godlygeek (approved):
I dropped the changes to the benchmarks/ directory (those files are vendored directly from an upstream source), and needed to switch the Ubuntu image to use ./venv instead of just /venv since we're not running as root in it.

Otherwise, LGTM!

godlygeek (approved):
LGTM

PR #363, opened 2023-04-24
Only delete the *.so files that we create

### LINKED ISSUE #353: `make clean` breaks virtualenvs stored inside the repo clone
If you create a virtualenv inside a clone of the repo, running `make clean` will break it by removing all of the `*.so` files inside of it. We should improve our `make clean` to only remove files that we know we've created.

jay-z007:
*Issue number of the reported bug or feature request: #353 *

**Describe your changes**
Delete only the following:
- `_test_utils`
- `_memray`
- `_inject`
- `build/` dir

**Testing performed**
To make sure that the virtualenv is working after running `make clean`, I did the following,
- Installed some packages in the virtualenv
- Start the interpreter and import a few modules

**Additional context**
Closes #353

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -179,8 +179,12 @@ gh-pages:  ## Publish documentation on BBGitHub Pages
 
 .PHONY: clean
 clean:  ## Clean any built/generated artifacts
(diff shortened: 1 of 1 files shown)


godlygeek (approved):
LGTM

PR #365, opened 2023-04-24
ci: Only check for news entries if src/ changed 

stefmolin:
**Describe your changes**
- Separate the check for a news entry into its own action to apply only to `src/*` files
- Clean up `sanity-check.yml` workflow file

**Testing performed**
This PR will trigger the `sanity-check.yml` workflow file (which will now pass since there is no need for the `skip news` label), but not the `news-check.yml` workflow file since there are no changes to `src/memray/*`.

### DIFF
--- a/.github/workflows/news-check.yml
+++ b/.github/workflows/news-check.yml
@@ -1,0 +1,23 @@
+name: News entry check
+on:
+  pull_request:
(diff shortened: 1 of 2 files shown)


godlygeek (approved):
LGTM

Issue #352, opened 2023-04-23
Set up CI to build and test musl libc wheels


godlygeek:
Currently `pystack` publishes `musllinux` wheels, but `memray` doesn't. Update our GitHub Actions to build, test, and publish `musllinux` wheels.

salticus:
I would like to try this issue today.

gusmonod:
Closed by the above PR. Thanks @salticus !

salticus:
You're welcome!

PR #379, opened 2023-04-25
Build musl x86-64 wheels

salticus:
#352 Set up CI to build and test musl libc wheels

**Describe your changes**
Enable musl builds in the pyproject file. Dependencies for memray are supplied using comparable libraries from the Alpine main repository.

**Testing performed**
Running against GitHub actions is producing long build times. Further changes may be necessary.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -40,7 +40,7 @@ jobs:
           CIBW_ARCHS_LINUX: ${{ matrix.cibw_arch }}
           CIBW_PRERELEASE_PYTHONS: True
           CIBW_TEST_EXTRAS: test
-          CIBW_TEST_COMMAND: pytest {package}/tests
+          CIBW_TEST_COMMAND: python -m pytest {package}/tests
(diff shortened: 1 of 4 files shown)


pablogsal:
Hummmm, seems that `lldb` is missing from the alpine repo?

pablogsal:
I have updated the PR skipping 32 bit muslinux wheels

godlygeek:
> Hummmm, seems that `lldb` is missing from the alpine 32 bits repo?

We could make that install optional. We already do that for the manylinux builds. The test suite just skips that `memray attach` test when `lldb` isn't available.

> We probably don't want 32 bits muslc wheels in any case

But that's fine, too.

There were still a few failing tests even for musl x86-64. Looks like the old musl libc version in the musllinux_1_1 container implements `aligned_alloc` and `realloc` in terms of the `malloc` symbol in the PLT, so we needed to set the recursion guard in a few more places in the hooks to prevent double counting of allocations with those allocators.

godlygeek (approved):
Thanks a lot for the contribution, @salticus!

PR #380, opened 2023-04-28
Work around an IPython packaging bug

godlygeek:
This works around a packaging bug in IPython 8.13: it drops support for Python 3.8, but still declares support for 3.8 in its `python_requires`, so `pip` installs the new version even for Python versions that don't support it.

See https://github.com/ipython/ipython/issues/14053


gusmonod (approved):
LGTM

PR #381, opened 2023-05-01
Drop pins to old IPython versions

godlygeek:
This reverts commit 62fb9ad15a8456bc356cbc3ba42e33c6299eafa5. This is no longer necessary now that IPython upstream has yanked the release which incorrectly asserted it was compatible with 3.8.

Reverts #380


gusmonod (approved):
LGTM

PR #326, opened 2023-03-02
Support high water mark temporal flame graphs

godlygeek:
Building on top of the work to support our temporal flame graphs for analyzing memory leaked within a user-selected range, allow creating temporal flame graphs for analyzing the high water mark within a user-selected range.

PR #383, opened 2023-05-16
Pre-factoring in preparation for temporal high water mark reports

godlygeek:
Split some code out of #326. These changes lay some of the groundwork for changes in #326, but can be reviewed separately and should be less controversial.


pablogsal (approved):

pablogsal:
LGTM

PR #390, opened 2023-06-08
Use std::all_of instead of loop + boolean

pablogsal:
This makes use of more idiomatic C++ and is anecdotally faster.


godlygeek (approved):
LGTM

PR #391, opened 2023-06-09
docs: Document temporal flame graphs

godlygeek:
For your reviewing convenience, here's previews of the rendered [temporal flame graph docs](https://godlygeek.github.io/memray/flamegraph.html#temporal-flame-graphs) and [news](https://godlygeek.github.io/memray/changelog.html).


pablogsal (approved):

godlygeek:
> but we can do that later

That's what I'm thinking. We probably want a demo where there's some setup code that uses a bunch of temporary memory, and then some other code that uses whatever was produced by the setup code, and we could demo how you could see what's going on in the setup phase vs the run phase. But that requires us to come up with some new demo code...

If we can come up with something good, we could add it as one of the example applications, and include its temporal flame graph in the docs so that people can easily play around with it.

godlygeek:
I've opened #392. Let's come back to it, unless you can immediately think of a good program for showing this off.

PR #394, opened 2023-06-09
Build wheels from sdist

godlygeek:


godlygeek:
Ah right, that's why I didn't do this sooner. 😔

pablogsal (dismissed):

pablogsal (approved):

Issue #396, opened 2023-06-09
Flamegraph view gets stuck 


mgmacias95:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I made this simple script to test the new temporary flamegraphs feature:
```python
import random
import string
from collections import Counter

def set_up():
  b = random.choice(string.ascii_letters)
  return Counter(b)

def run():
  c = set_up()
  key = random.choice(list(c.keys()))
  print(f'{key}: {c[key]}')


run()
```

I executed the following commands:
```shellsession
$ python -m memray run test.py                                   
Writing profile results into memray-test.py.73578.bin
o: 1
[memray] Successfully generated profile results.

You can now generate reports from the stored allocation records.
Some example commands to generate reports:

/Users/mgmacias/Documents/GitHub/memray/env/bin/python -m memray flamegraph memray-test.py.73578.bin
$ python -m memray flamegraph memray-test.py.73578.bin --temporal 
Wrote memray-flamegraph-test.py.73578.html
```

Then opened the flamegraph on my chrome browser. The flamegraph gets stuck on this loading view:
<img width="1505" alt="image" src="https://github.com/bloomberg/memray/assets/8604906/1ec596cc-81b8-4796-a814-d252c6aebdfd">

When checking the chrome dev console I can see the following error message:
<img width="551" alt="image" src="https://github.com/bloomberg/memray/assets/8604906/314ad9b8-ef38-4215-a643-ace1ceab5e23">

### Expected Behavior

An error message or a flamegraph.

### Steps To Reproduce

1. Save the script to a python file
2. Run the commands mentioned above
3. Open the flamegraph in chrome

### Memray Version

1.8.0

### Python Version

3.11

### Operating System

macOS

pablogsal:
There are two possible problems that this bug surfaces:

* Is possible that the script runs so fast that the two ends of the graph are the same.
* If the memory is flat the high the `hwmBytes` is somehow `undefined`.

pablogsal:
This seems to happen because in this case `high_water_mark_by_snapshot` is an empty array.

pablogsal:
I think at the very least this:

https://github.com/bloomberg/memray/blob/c2c8d53393de25670a9da0325c34027a7ed4c015/src/memray/reporters/assets/temporal_flamegraph.js#L55

should be

```
  if (hwms && hwms.length > 0) {
```

pablogsal:
I think this is enough to fix it:

```diff
[...]

PR #399, opened 2023-06-12
Fix temporal flame graph handling of HWM before the first or after the last snapshot

pablogsal:
- Ensure we don't try to build a plot if we don't have data
- Update `package-lock.json` and binary files

closes: #396


src/memray/reporters/assets/temporal_flamegraph.js
```diff
@@ -52,7 +52,7 @@ function packedDataToTree(packedData, rangeStart, rangeEnd) {
   }
 
   const hwms = packedData.high_water_mark_by_snapshot;
-  if (hwms) {
+  if (hwms && hwms.length > 0 && memory_records.length > 0) {
```
godlygeek, line 55: This fix isn't correct. The `else` block of this `if` is for temporal leaks mode, rather than temporal high water mark mode. We don't want to go into that if `hwms` is set, even if it's set to an empty array.

godlygeek:
Oh boy, there was a lot here that wasn't working great. We had problems when the high water mark was before the first snapshot, when it was after the last snapshot, when there were exactly 0 snapshots, and when there was exactly 1 snapshot.

I think this PR now fixes everything - at least, it fixes all the problems I could find. Take a look when you can, @pablogsal - there was enough wrong here that we should probably cut a bug fix release once we land this.

tests/unit/test_high_water_mark_aggregator.py
```diff
@@ -49,7 +49,7 @@ def test_no_allocations_at_start():
     # THEN
     assert 0 == tester.get_current_heap_size()
     assert [] == tester.get_allocations()
-    assert [] == list(tester.high_water_mark_bytes_by_snapshot())
+    assert [0] == list(tester.high_water_mark_bytes_by_snapshot())
```
pablogsal, line 52: I find this a bit weird, to be honest, but I suppose it makes sense.

pablogsal:
I reviewed your commits but I cannot accept my own PR 😅

godlygeek (approved):
Well, it LGTM!

PR #401, opened 2023-06-14
Teach memray to patch libraries that are part of the macOS linker cache

pablogsal:


godlygeek (approved):
After my changes, this all LGTM, so if my changes all seem correct to you, I think this is good to land.

PR #409, opened 2023-06-21
ci: Test with Python 3.12

godlygeek:


godlygeek:
OK, this should be ready to be reviewed. The way I'm avoiding publishing the 3.12 wheels is hacky, but I don't see a non-hacky solution, and I manually tested that it works by forcing the `upload_pypi` stage to run and to skip the actual wheel publishing.

pablogsal (approved):
Let the hacks happen :)

PR #416, opened 2023-07-11
Build also aarch64 wheels

pablogsal:
Seems that cibuildwheel now allows building aarch64 wheels, so we should
try to distribute these as well.


pablogsal:
Screenshot 2023-07-12 at 12 26 38

This worked. I am switching the aarch64 to only run during pip promotion

godlygeek (approved):
LGTM

Issue #404, opened 2023-06-16
Expose a pause button for `memray live`


godlygeek:
Add a key binding for pausing refreshes, and another key binding for resuming the display. Ensure that records continue to be consumed from the socket, and only the refreshes of the TUIs data are suspended.

pablogsal:
Closed by https://github.com/bloomberg/memray/pull/418

PR #418, opened 2023-07-12
Add pause button to live view

bilbofroggins:
*Issue number of the reported bug or feature request: #404 *

**Describe your changes**
P key to pause, and when paused, U key to unpause.
Couldn't see anything built in to rich live that will automatically do this, so I just split the TUI class into a layout class (TUI) and a data class (TUIData) so that we can snapshot the data when the user pauses and show the snapshot while continuing to write to the live version of data in the background (in update_snapshot)

Previously used to put the whole _snapshot (Iterable[AllocationRecord]) within the TUI class and then extract what we cared about in the render stage (get_body) but there's a bunch of extra stuff in there it seems like is not needed. Instead, extracting the data we need in update_snapshot and then get_body can just pull from TUIData.

**Testing performed**
Tested an application that has a couple threads - made sure you can still arrow through them.
Sorting still works and carries over paused/unpaused states.
Ran pytest locally.


pablogsal:
I have not reviewed the code yet, but I noticed that when running ` memray run --live -m test test_asyncio` and pausing, hitting `U` or `P` to unpause doesn't continue the process and the table does not keep updating for me.

pablogsal:
Update: seems that it only updates when I keep pressing `U`

bilbofroggins:
> Update: seems that it only updates when I keep pressing U

Oh interesting, I'm not seeing that on my side but I just put in a flag to prevent data from being copied a bunch while holding the keys down - and pushed that up. Possibly related?

pablogsal:
> > Update: seems that it only updates when I keep pressing U
> 
> Oh interesting, I'm not seeing that on my side but I just put in a flag to prevent data from being copied a bunch while holding the keys down - and pushed that up. Possibly related?

Unfortunately still happens on my end after the latest commit

pablogsal:
Thanks a lot @bilbofroggins this looks fantastic. Very good job! 🚀 

Can you maybe add a small test that checks the `pause()` and `unpause()` functions simulating an update in the middle?

bilbofroggins:
> Thanks a lot @bilbofroggins this looks fantastic. Very good job! 🚀
> 
> Can you maybe add a small test that checks the `pause()` and `unpause()` functions simulating an update in the middle?

👍 Added

pablogsal:
Will make another pass tomorrow! Thanks for the ping

pablogsal (approved):
LGTM! Excellent work @bilbofroggins ✨ 

Thanks a lot for contributing to memray :)

PR #436, opened 2023-08-01
ci: Automatically publish docs for new releases

godlygeek:
An update of #223, rebased and PR'd from a branch instead of a fork so that it can (hopefully) request write permissions for the job that needs to publish the docs...

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -127,7 +127,6 @@ jobs:
       - name: Lint sources
         run: |
           make lint
-          python3 -m pre_commit run --all-files --hook-stage pre-push
       - name: Build docs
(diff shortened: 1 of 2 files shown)


godlygeek:
The new job has
```yaml
    if: github.event_name == 'release' && github.event.action == 'published'
```
so that it only runs for new releases, but I iterated on this without that guard to confirm that it works, and the docs currently up on https://bloomberg.github.io/memray/ were published by this job. Had to prove it works somehow, and it was a perfect time to do this since we just had a release and hadn't yet updated the docs 😄

pablogsal (approved):

pablogsal:
Wonderful!

Issue #293, opened 2023-01-13
Inverted flame graphs


godlygeek:
### Is there an existing proposal for this?

- [X] I have searched the existing proposals

### Is your feature request related to a problem?

The [Fil profiler](https://github.com/pythonspeed/filprofiler) has the ability to generate inverted flame graphs. In a normal flame graph, the roots are thread entry points, the children of any given node are the functions called by that node, and the leaf nodes are functions that allocated memory but did not call any functions that allocated memory. This can result in duplicated leaf nodes, if there are multiple call stacks that lead to a particular function that performs allocations. In an inverted flame graph, the roots are the functions that allocated memory, and the children of any given node represent the percentage of any of that node's allocations that can be attributed to a particular caller.

As Fil's documentation describes it:

> The second graph shows the reverse callgraph, from <tt>f()</tt> upwards.</p>
> <p>Why is the second graph useful? If <tt>f()</tt> is called from multiple places, in the first graph it will show up multiple times, at the bottom.
> In the second reversed graph all calls to <tt>f()</tt> will be merged together.</p>

### Describe the solution you'd like

We should implement this idea, as it can sometimes be very helpful in analyzing where memory is being spent in aggregate.

gusmonod:
@ivonastojanovic is taking this with #439. Could you please comment on this issue Ivona so that I can assign it to you?

ivonastojanovic:
Hey, a comment :)

PR #439, opened 2023-08-14
Inverted flame graph

ivonastojanovic:
Added inverted flame graph which is generated with --inverted command line argument. The tree's nodes with the same parent and location are merged when we check 'Hide Import System Frames' checkbox.

*Fixes: #293*

**Describe your changes**

**Testing performed**

**Additional context**


src/memray/reporters/flamegraph.py
```diff
@@ -75,6 +75,21 @@ def __init__(
         self.data = data
         self.memory_records = memory_records
 
+    @classmethod
+    def update_import_system_frames_for_inverted_flamegraph(
```
pablogsal, line 79: This may have two problems:

* For deep trees, recursion can be quite slow.
* In Python there is a recursion limit to protect from stack overflows. This means that if the tree is deep enough this may blow up with a `RecursionError`.

We could transform this function into something iterative that keeps the stack manually but on the other hand we are already limiting the stack depth here:

https://github.com/bloomberg/memray/blob/01542287336fa7b5386b6ad39996ab92cee318b6/src/memray/reporters/flamegraph.py#L164-L167

It may be good to add a test to see that we don't crash for a stack that's too deep. The test would just see what's the generated data for an input that's very nested.

src/memray/reporters/flamegraph.py
```diff
@@ -178,6 +195,11 @@ def _from_any_snapshot(
                     for interval in intervals
                 )
 
+        if inverted:
+            frames[0][
```
pablogsal, line 199: Is the assignment necessary? `update_import_system_frames_for_inverted_flamegraph` does this:

```
nodes[node_index]["import_system"] |= result
```

and on the first call `node_index` is 0 so it should update frames[0] already, no?

src/memray/reporters/templates/__init__.py
```diff
@@ -28,7 +28,9 @@ def include_file(name: str) -> Markup:
     return env
 
 
-def get_report_title(*, kind: str, show_memory_leaks: bool) -> str:
+def get_report_title(*, kind: str, show_memory_leaks: bool, inverted: bool) -> str:
```
pablogsal, line 31: I think you missed some updates of calls to this function for example here: `tests/unit/test_templates.py`

src/memray/reporters/assets/common.js
```diff
@@ -192,6 +192,32 @@ function filterFramesByFunc(root, func) {
   return _.defaults({ children: children }, root);
 }
 
+function mergeNodes(root) {
```
pablogsal, line 195: This function will likely need some tests in `src/memray/reporters/assets/common.test.js`

src/memray/reporters/flamegraph.py
```diff
@@ -7,10 +7,13 @@
 from typing import Iterable
 from typing import List
 from typing import Optional
+from typing import Sequence
 from typing import TextIO
 from typing import Tuple
+from typing import TypedDict
```
godlygeek, line 13: CI is failing because of this. We still support 3.7, and `typing.TypedDict` wasn't added until Python 3.8. We either need to use `from typing_extensions import TypedDict` on Python 3.7, or drop support for Python 3.7 - or maybe make this conditional on Python version... or maybe only use `TypedDict` when `typing.TYPE_CHECKING` is set.

src/memray/reporters/flamegraph.py
```diff
@@ -75,6 +105,21 @@ def __init__(
...
+        cls, nodes: Sequence[FrameNodeDict], node_index: int
+    ) -> bool:
+        if not nodes or not nodes[node_index]:
+            return False
+
+        for child_index in nodes[node_index]["children"]:
+            result = cls.update_import_system_frames_for_inverted_flamegraph(
+                nodes, child_index
+            )
+            nodes[node_index]["import_system"] |= result
```
godlygeek, line 119: The `|=` here is essentially saying: if any of this node's callers are part of the import system, then this is an import system node. That doesn't seem correct to me. Consider a test case like this:
```py
    def test_inverted_allocator_sometimes_called_under_importlib(self):
        # GIVEN
        peak_allocations = [
            MockAllocationRecord(
                tid=1,
                address=0x1000000,
                size=1024,
                allocator=AllocatorType.MALLOC,
                stack_id=1,
                n_allocations=1,
                _stack=[
                    ("malloc", "malloc.c", 12),
                    ("parent_one", "fun.py", 8),
                    ("main", "main.py", 4),
                ],
            ),
            MockAllocationRecord(
                tid=1,
                address=0x1000000,
                size=1024,
                allocator=AllocatorType.MALLOC,
                stack_id=1,
                n_allocations=1,
                _stack=[
                    ("malloc", "malloc.c", 12),
                    ("importer", "<frozen importlib>", 4),
                    ("parent_two", "fun.py", 10),
                    ("main", "main.py", 4),
                ],
            ),
        ]
```

In that example, `malloc` can be called from two different paths - either by `main` calling `parent_one` calling `malloc`, or by `main` calling `parent_two` calling `importer` calling `malloc`. `importer` is part of the import system, so there's one path to `malloc` that goes through the import system and one that doesn't. Each of those paths are responsible for half the bytes allocated by `malloc`.

Our non-inverted flame graph displays these stacks for that:
```text
|                root size=2048               |
|                main size=2048               |
| parent_one size=1024 | parent_two size=1024 |
|   malloc size=1024   |  importer size=1024  |
|                      |   malloc size=1024   |
```

When you click the "hide import system frames" checkbox, `importer` and the frames below it are hidden, and `parent_two` becomes a leaf:
```
|                root size=2048               |
|                main size=2048               |
| parent_one size=1024 | parent_two size=1024 |
|   malloc size=1024   |                      |
```

For our inverted flame graphs, I expect the exact same stacks to be shown, just with leaves and roots swapped. So, when import system frames are not hidden, I expect us to show this:
```text
|                root size=2048               |
|               malloc size=2048              |
| parent_one size=1024 |  importer size=1024  |
|    main size=1024    | parent_two size=1024 |
|                      |    main size=1024    |
```
But when import system frames are hidden, I expect it to instead show:
```text
|                root size=2048               |
|   malloc size=1024   | parent_two size=1024 |
| parent_one size=1024 |    main size=1024    |
|    main size=1024    |                      |
```

The `main -> parent_one -> malloc` path remains unchanged by "hide import system frames", since that path doesn't go through the import system, but the `main -> parent_two -> importer -> malloc` stack gets truncated to `main -> parent_two` when `importer` is reached (just like it is for non-inverted flame graphs), and the stack we show now ends with `parent_two` as a child of the root (because `parent_two` is a leaf in the non-inverted hide-import-system case, and inverted mode swaps leaves and roots). Which means that the `malloc` node shrinks from 2048 bytes to 1024 bytes, since half of its allocations are attributable to a path through the import system, and half aren't.

There's no way we can achieve that desired behavior if we update the `malloc` node to say that it's an import system frame because one of its callers is an import system frame, even though the other isn't. As this function is currently implemented, I think it would result in us hiding `malloc` entirely, effectively truncating the `main -> parent_one -> malloc` call stack even though it didn't pass through the import system.

I can think of 3 reasonable ways to handle this:
1. Build two separate trees in `--inverted` mode. For one, use full stacks as the input. For the other, truncate the stacks at import system calls before processing them into a tree. Include both trees in the inverted-mode HTML, and have the checkbox switch which one is rendered.
2. In the `FrameNodeDict`, instead of recording just a `bool` for whether or not the node was reached by the import system, record an `int`: the number of bytes allocated at that node through a path that went through the import system. That would change this `|=` into a `+=`. When the checkbox is unset the inverted-mode JS code would ignore that field. When the checkbox is checked the JS code would shrink every node by treating that node's size as `value - import_system_bytes`, instead of as `value` bytes. Any node where `value - import_system_bytes` is 0 would be removed, and its children would be hoisted up to the root.
3. Pre-process each stack to identify which frames are below a call into the import system. Include the is-import-system flag as part of the node key, so that we have two different `malloc` nodes in the dumped tree, each associated with 1024 bytes. When the checkbox is checked, our JS code drops nodes with is-import-system set, and hoists their children up a level. When the checkbox is unchecked, our JS code would need to traverse the tree and merge nodes together if they have the same parent and same location, but one is flagged is-import-system and the other isn't.

Of these options, I think 1 is probably the simplest, and 3 is probably the most complex.  2 might require the fewest changes to what's been implemented so far...

@pablogsal Please check what I've said here, and see if a) you agree with me about what the checkbox should do, and b) if my proposed solutions seem reasonable.

src/memray/reporters/templates/__init__.py
```diff
@@ -28,7 +28,11 @@ def include_file(name: str) -> Markup:
     return env
 
 
-def get_report_title(*, kind: str, show_memory_leaks: bool) -> str:
+def get_report_title(
+    *, kind: str, show_memory_leaks: bool, inverted: bool = False
+) -> str:
+    if inverted and kind == "flamegraph":
+        kind = f"inverted {kind}"
```
godlygeek, line 35: Don't we want to return "inverted flamegraph report (memory leaks)" if `show_memory_leaks` is true? I think what we want is:
```py
    parts = []
    if inverted:
        parts.append("inverted")
    parts.append(kind)
    parts.append("report")
    if show_memory_leaks:
        parts.append("(memory leaks)")
    return " ".join(parts)
```

Also: Don't worry about doing the `and kind == "flamegraph"` here. I think it's fine to return "inverted table report" if someone calls this with `kind="table"`. We know they won't, but if they for some reason do, that would be sane behavior, so there's no reason to avoid it.

godlygeek (changes requested):

godlygeek:
As a heads up, I haven't looked at the JS or the tests yet, so the above comments are what I have to say about the Python side of the implementation.

src/memray/reporters/flamegraph.py
```diff
@@ -75,6 +112,97 @@ def __init__(
...
+        is_import_system = False
+
+        for index, stack_frame in stack_it:
+            node_key = (current_frame_id, stack_frame, record.thread_name)
+            if node_key not in node_index_by_key:
+                if is_cpython_internal(stack_frame):
+                    num_skipped_frames += 1
+                    continue
+                # update import_system only if we're generting normal flamegraph
+                # and inverted flamegrap
```
godlygeek, line 174:
```suggestion
                # and inverted flamegraph
```

src/memray/reporters/flamegraph.py
```diff
@@ -75,6 +112,97 @@ def __init__(
...
+                # we don't need to calculate is_frame_from_import_system for inverted
+                # import system tree again
+                if (inverted and not is_inverted_import_system) or not is_import_system:
+                    is_import_system = is_frame_from_import_system(stack_frame)
```
godlygeek, line 178: I think you should just skip this call whenever `inverted` is true. When `inverted and is_inverted_import_system`, the caller already stripped the import system frames, so we know `is_frame_from_import_system` would always return false. But when `inverted and not is_inverted_import_system` - that is, the full inverted tree without import system nodes dropped - we know that this question isn't answerable. The flag doesn't make sense, because all the call paths that lead to one function are children of the same node. Taking my example from the other day, it doesn't make sense to flag `malloc` as either `is_import_system = True` or as `is_import_system = False`, because half of its callers are import system calls and half aren't. I think you should just skip this check entirely when `inverted is True`, and just leave every node with `is_import_system = False`, because we can't give any other answer accurately.
```suggestion
                if not inverted and not is_import_system:
                    is_import_system = is_frame_from_import_system(stack_frame)
```

src/memray/reporters/flamegraph.py
```diff
@@ -119,52 +260,65 @@ def _from_any_snapshot(
...
+                import_system_stack = list(
+                    reversed(
+                        list(
+                            itertools.takewhile(
+                                lambda e: not is_frame_from_import_system(e),
+                                reversed(stack),
+                            )
                         )
                     )
+                )
```
godlygeek, line 310: I like this! Very elegant.

src/memray/reporters/flamegraph.py
```diff
@@ -83,22 +211,35 @@ def _from_any_snapshot(
...
-            "location": [html.escape("<tracker>"), "<b>memray</b>", 0],
+            "location": (html.escape("<tracker>"), "<b>memray</b>", 0),
             "value": 0,
             "children": [],
             "n_allocations": 0,
             "thread_id": "0x0",
             "interesting": True,
             "import_system": False,
         }
+        inverted_import_system_root = deepcopy(root) if inverted else None
```
godlygeek, line 232: I think every variable name in this function with `_import_system_` in the name should be renamed to `_no_import_system_` (or just `_no_imports_`) instead.

`inverted_import_system_root` makes it sound like this is the root of an inverted tree that _does_ contain import system nodes, when actually its the root of an inverted tree that contains only non-import-system nodes.

src/memray/reporters/flamegraph.py
```diff
@@ -83,22 +211,35 @@ def _from_any_snapshot(
         memory_records: Iterable[MemorySnapshot],
         native_traces: bool,
         temporal: bool,
+        inverted: Optional[bool] = None,
     ) -> "FlameGraphReporter":
-        root: Dict[str, Any] = {
+        # TODO?
+        if temporal and inverted:
+            raise NotImplementedError(
```
godlygeek, line 218: Have you given any thought to what it would take to extend this to work for temporal mode? It should definitely be a separate PR, this one is long enough - but I don't think it would be too tough to implement.

src/memray/reporters/templates/base.html
```diff
@@ -139,8 +139,11 @@ <h5 class="modal-title" id="helpModalLabel">How to interpret {{ kind }} reports<
   <script type="text/javascript">
     const packed_data = {{ data|tojson }};
     var data = null;
+    var flamegraphData = null;
+    var invertedImportSystemData = null;
```
godlygeek, line 143: As above, this makes it sound like this one _does_ include import system nodes, rather than does not. I'd call it something like `importSystemHiddenFlamegraphData` or `noImportSystemData` or `flamegraphDataNoImportSystem` or something.

src/memray/reporters/assets/flamegraph.js
```diff
@@ -42,9 +42,29 @@ function packedDataToTree(packedData) {
...
+    unique_threads: unique_threads,
+  });
+
+  invertedImportSystemData = inverted
+    ? packedDataToTree({
+        strings: strings,
+        nodes: inverted_import_system_nodes,
+        unique_threads: unique_threads,
+      })
+    : {};
```
godlygeek, line 60: Let's do `null` here instead of an empty object. That's a better indicator for the particular sort of invalid that this is.
```suggestion
    : null;
```

godlygeek (commented):
This is looking great! I've taken a look at the Javascript this time around (though I still haven't looked closely at the tests), and most of the concerns I have now are pretty minor.

src/memray/reporters/flamegraph.py
```diff
@@ -83,22 +216,33 @@ def _from_any_snapshot(
...
+            inverted_no_imports_root = deepcopy(root)
+            inverted_no_imports_frames = [inverted_no_imports_root]
 
-        frames = [root]
         interval_list: List[Tuple[int, Optional[int], int, int, int]] = []
 
         node_index_by_key: Dict[Tuple[int, StackFrame, str], int] = {}
+        inverted_no_imports_node_index_by_key: Dict[
+            Tuple[int, StackFrame, str], int
+        ] = {}
```
godlygeek: I'm gonna switch this to use a type alias, to make things slightly easier to read:
```py
        NodeKey = Tuple[int, StackFrame, str]
        node_index_by_key: Dict[NodeKey, int] = {}
        inverted_no_imports_node_index_by_key: Dict[NodeKey, int] = {}
```
In particular, it's tough to see that the two variables have the same type given the way that Black wrapped the lines, and this change makes that more recognizable.

src/memray/reporters/flamegraph.py
```diff
@@ -117,87 +261,78 @@ def _from_any_snapshot(
...
-                root["n_allocations"] += n_allocations
-
-            current_frame_id = 0
-            current_frame = root
+            record_data: RecordData = {
+                "thread_name": thread_id,
+                "size": size,
+                "n_allocations": n_allocations,
+                "intervals": intervals,
+            }
```
godlygeek: I'm gonna move this to be defined in each branch of the `if` above, so that we don't need `intervals` and `size` and `n_allocations` and `thread_id` as local variables in addition to values in this dict, for the same reason as above - limiting the number of local variables here will help future readers.

src/memray/reporters/flamegraph.py
```diff
@@ -117,87 +261,78 @@ def _from_any_snapshot(
...
-                        create_framegraph_node_from_stack_frame(
-                            stack_frame,
-                            import_system=is_import_system,
-                            thread_id=thread_id,
+
+            # normal flamegraph
+            if not inverted:
+                stack_it = enumerate(reversed(stack))
+                cls.generate_frames(
+                    stack_it=stack_it,
```
godlygeek: I'm gonna drop this local variable as well, and just construct it in the call to the function below, since it's only used in one spot, and to keep limiting local state.

src/memray/reporters/flamegraph.py
```diff
@@ -117,87 +261,78 @@ def _from_any_snapshot(
...
+                no_imports_stack = list(
+                    reversed(
+                        list(
+                            itertools.takewhile(
+                                lambda e: not is_frame_from_import_system(e),
+                                reversed(stack),
+                            )
                         )
                     )
+                )
```
godlygeek: I'm gonna split this into a function, so that we can assign a nice name to it for the benefit of future maintainers.

godlygeek (approved):
I've made a few small changes, but this LGTM!

PR #441, opened 2023-08-16
ci: Allow publishing Python 3.12 wheels

godlygeek:
Python 3.12 is now ABI stable, so we should publish wheels for it.


gusmonod (approved):
LGTM

PR #452, opened 2023-09-07
Temporal inverted flame graph

ivonastojanovic:

**Describe your changes**

**Testing performed**

**Additional context**

### DIFF
--- a/src/memray/reporters/assets/temporal_flamegraph.js
+++ b/src/memray/reporters/assets/temporal_flamegraph.js
@@ -17,19 +17,22 @@ import {
 var active_plot = null;
 var current_dimensions = null;
 
-var parent_index_by_child_index = (function () {
-  let ret = new Array(packed_data.nodes.children.length);
+var parent_index_by_child_index = generateParentIndexes(packed_data.nodes);
+var inverted_no_imports_parent_index_by_child_index = inverted
+  ? generateParentIndexes(packed_data.inverted_no_imports_nodes)
+  : null;
+
+function generateParentIndexes(nodes) {
+  let ret = new Array(nodes.children.length);
   console.log("finding parent index for each node");
-  for (const [parentIndex, children] of packed_data.nodes.children.entries()) {
+  for (const [parentIndex, children] of nodes.children.entries()) {
     children.forEach((idx) => (ret[idx] = parentIndex));
   }
   console.assert(ret[0] === undefined, "root node has a parent");
   return ret;
-})();
+}
 
-function packedDataToTree(packedData, rangeStart, rangeEnd) {
-  const { strings, nodes, unique_threads } = packedData;
-
+function generateNodeObjects(strings, nodes) {
   console.log("constructing nodes");
   const node_objects = nodes.name.map((_, i) => ({
     name: strings[nodes["name"][i]],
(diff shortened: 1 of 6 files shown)


src/memray/reporters/templates/base.html
```diff
@@ -144,6 +144,7 @@ <h5 class="modal-title" id="helpModalLabel">How to interpret {{ kind }} reports<
     const merge_threads = {{ merge_threads|tojson }};
     const memory_records = {{ memory_records|tojson }};
     const inverted = {{ inverted|tojson }};
+    const temporal = packed_data["high_water_mark_by_snapshot"] != null;
```
godlygeek, line 147: This isn't fully correct - `--temporal --leaks` flamegraphs don't have `high_water_mark_by_snapshot`. Let's instead check:
```suggestion
    const temporal = packed_data.intervals != null;
```

That condition is true both with and without `--leaks`

src/memray/reporters/assets/temporal_flamegraph.js
```diff
@@ -27,9 +27,8 @@ var parent_index_by_child_index = (function () {
   return ret;
 })();
 
-function packedDataToTree(packedData, rangeStart, rangeEnd) {
-  const { strings, nodes, unique_threads } = packedData;
-
+function generateNodeObjects(packedData) {
+  const { strings, nodes } = packedData;
```
godlygeek, line 31: Since we only need 2 things, let's just have the caller pass two parameters, instead of forcing them to pass an object that we can unpack here

src/memray/reporters/assets/temporal_flamegraph.js
```diff
@@ -51,6 +50,88 @@ function packedDataToTree(packedData, rangeStart, rangeEnd) {
...
+    flamegraphNodeObjects: flamegraphNodeObjects,
+    invertedNoImportsNodeObjects: invertedNoImportsNodeObjects,
+  };
+}
+
+function findHWMAllocations(intervals, node_objects, hwmSnapshot) {
+  if (!node_objects) {
+    return;
+  }
+
```
godlygeek, line 92: Hm, this check is new - why do we need it?

src/memray/reporters/assets/temporal_flamegraph.js
```diff
@@ -51,6 +50,88 @@ function packedDataToTree(packedData, rangeStart, rangeEnd) {
...
+    if (
+      allocBefore >= rangeStart &&
+      allocBefore <= rangeEnd &&
+      (deallocBefore === null || deallocBefore > rangeEnd)
+    ) {
+      while (nodeIndex !== undefined) {
+        node_objects[nodeIndex].n_allocations += count;
+        node_objects[nodeIndex].value += bytes;
+        nodeIndex = parent_index_by_child_index[nodeIndex];
```
godlygeek, line 125: Ah, this is the bug you hit. You're starting with an index into `inverted_no_imports_nodes` and looking it up in `parent_index_by_child_index`, which is a mapping from index in `nodes` to index in `nodes`.

`parent_index_by_child_index` only knows about the main tree, not the inverted-no-import-system-nodes tree. We need a separate mapping table that maps child node index to parent node index for the new tree.

godlygeek (approved):
LGTM! I've squashed the fixup and slightly reworded the commit message, but otherwise I'm happy to land this.

PR #455, opened 2023-09-08
Added --aggregate option to attach

ivonastojanovic:
Added --aggregate option which allows user to request aggregated mode for in-memory aggregation.

**Describe your changes**

**Testing performed**

**Additional context**

### DIFF
--- a/src/memray/commands/attach.py
+++ b/src/memray/commands/attach.py
@@ -242,6 +242,13 @@ class AttachCommand:
         )
 
         parser.add_argument(
+            "--aggregate",
+            help="Write aggregated stats to the output file instead of all allocations",
+            action="store_true",
+            default=False,
+        )
+
+        parser.add_argument(
             "--native",
             help="Track native (C/C++) stack frames as well",
             action="store_true",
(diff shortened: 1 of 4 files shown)


src/memray/commands/attach.py
```diff
@@ -332,11 +340,19 @@ def run(self, args: argparse.Namespace, parser: argparse.ArgumentParser) -> None
             live_port = _get_free_port()
             destination = memray.SocketDestination(server_port=live_port)
 
+        if args.aggregate and not hasattr(args, "output"):
+            parser.error("Can't use aggregated mode without an output file.")
```
godlygeek, line 344: This `parser.error()` never fires, because `hasattr(args, "output")` is always `True`.

Sometimes the attribute is `None`, but it always exists. If it didn't, we'd get an `AttributeError` on this line above: https://github.com/bloomberg/memray/blob/2751eb4700f09700abebb668299260aef1b3b3d2/src/memray/commands/attach.py#L332

We should add a unit test that exercises this case, too, so that we're not introducing a new, uncovered line of code.

src/memray/commands/attach.py
```diff
@@ -332,11 +340,19 @@ def run(self, args: argparse.Namespace, parser: argparse.ArgumentParser) -> None
...
+
+        file_format = (
+            f"file_format={FileFormat.AGGREGATED_ALLOCATIONS}" if args.aggregate else ""
+        )
```
godlygeek, line 348: Oooh, this is interesting. This winds up passing `file_format=1`, because:
```py
>>> FileFormat.AGGREGATED_ALLOCATIONS
<FileFormat.AGGREGATED_ALLOCATIONS: 1>
>>> f"file_format={FileFormat.AGGREGATED_ALLOCATIONS}"
'file_format=1'
```

I'm surprised that works. I think it would be more robust to do:
```py
        file_format = (
            "file_format=memray.FileFormat.AGGREGATED_ALLOCATIONS" if args.aggregate else ""
        )
```

I'd rather pass around the enum by name than by value, in case a future version of the API changes the value associated with the enum, or changes the `Tracker` constructor to reject ints and only accept enum members.

I'd like to see an integration test that exercises this, too - that should be as easy as [parametrizing](https://docs.pytest.org/en/7.3.x/how-to/parametrize.html) our [current `attach` test](https://github.com/bloomberg/memray/blob/main/tests/integration/test_attach.py#L55) with a new flag for whether or not to use aggregated mode, passing `--aggregate` in the `attach_cmd` when the flag is set, and asserting that `get_allocation_records()` fails when `--aggregate` was passed but succeeds when it wasn't, and that `get_high_watermark_allocation_records()` works in both cases.

godlygeek (changes requested):

godlygeek (approved):
As we discussed offline I've modified the integration tests a bit, but otherwise this looks great to me! I'm going to squash and land this.

PR #458, opened 2023-09-12
Added new commands to memray attach

ivonastojanovic:
Letting memray attach supports deactivating an attached tracker manually, or after a specified heap size is reached, or after a specified time limit has elapsed.

**Describe your changes**

**Testing performed**

**Additional context**

### DIFF
--- a/src/memray/commands/attach.py
+++ b/src/memray/commands/attach.py
@@ -19,14 +19,49 @@ from memray._errors import MemrayCommandError
 from .live import LiveCommand
 from .run import _get_free_port
 
+try:
+    from typing import Literal
+except ImportError:
+    from typing_extensions import Literal  # type: ignore
+
+TrackingMode = Literal["ACTIVATE", "DEACTIVATE", "UNTIL_HEAP_SIZE", "FOR_DURATION"]
+
+
 GDB_SCRIPT = pathlib.Path(__file__).parent / "_attach.gdb"
 LLDB_SCRIPT = pathlib.Path(__file__).parent / "_attach.lldb"
 RTLD_DEFAULT = memray._memray.RTLD_DEFAULT
 RTLD_NOW = memray._memray.RTLD_NOW
 PAYLOAD = """
 import atexit
+import time
+import threading
+import resource
+import sys
+from contextlib import suppress
 
 import memray
+
+
+def _get_current_heap_size() -> int:
+    usage = resource.getrusage(resource.RUSAGE_SELF)
+    rss_bytes = usage.ru_maxrss * 1024  # Convert from KB to bytes
+    return rss_bytes
+
+
+class RepeatingTimer(threading.Thread):
+    def __init__(self, interval, function):
+        self._interval = interval
+        self._function = function
+        self._canceled = threading.Event()
+        super().__init__()
+
+    def cancel(self):
+        self._canceled.set()
+
+    def run(self):
+        while not self._canceled.wait(self._interval):
+            if self._function():
+                break
 
 
 def deactivate_last_tracker():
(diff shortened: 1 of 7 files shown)


src/memray/commands/attach.py
```diff
@@ -12,22 +12,49 @@
...
+import pathlib
+from typing import Set
+from typing import Union
+from contextlib import suppress
 
 import memray
+from memray.commands.attach import TrackingMode
+
+ACTIVE_THREADS: Set[Union[threading.Thread, threading.Timer]] = set()
+TRACKER_LOCK = threading.Lock()
```
godlygeek, line 50: This isn't going to work the way you want. You're trying to share these globals across multiple calls to attach, but each attach call runs its own script, with its own globals namespace. That's why we do the `memray._last_tracker = ...` thing below: we're sticking the tracker, which we need to be shared across calls to `memray attach`, in a global variable of the `memray` module, so that future calls from different namespaces can share it.

Honestly, this is probably a reason to split this out into a proper module. This is getting too complicated to manage as an inline script like this... I'm going to take a swing at refactoring this and giving you something easier to work with for basing your changes on top of...

src/memray/commands/attach.py
```diff
@@ -282,6 +350,22 @@ def prepare_parser(self, parser: argparse.ArgumentParser) -> None:
             choices=["auto", "gdb", "lldb"],
         )
 
+        parser.add_argument(
+            "--mode",
+            help="Activate or deactivate tracking.",
+            type=str,
+            default="activate",
```
godlygeek, line 357: Let's instead go with a boolean option named `--stop-tracking`, so that the user runs:
```shell
memray attach --stop-tracking $pid
```

src/memray/commands/attach.py
```diff
@@ -297,6 +381,18 @@ def prepare_parser(self, parser: argparse.ArgumentParser) -> None:
...
+        heap_size = None
+
+        if args.mode == "deactivate":
+            mode = TrackingMode.DEACTIVATE
+        elif args.heap_size:
+            mode = TrackingMode.UNTIL_HEAP_SIZE
+            heap_size = args.heap_size
+        elif args.duration:
+            mode = TrackingMode.FOR_DURATION
+            duration = args.duration
```
godlygeek, line 395: We should have some extra checking that these different modes can't be missed. As it is, the user is allowed to run `memray attach --mode=deactivate --duration=10 --output=foo.bin` which doesn't make any sense. We should prevent `--deactivate-tracking` from being used with any other option except `--method`, and we should prevent `--heap-limit` and `--duration` from being used together.

godlygeek (changes requested):

src/memray/commands/attach.py
godlygeek: OK, the last commit on https://github.com/bloomberg/memray/pull/459 gives you something saner to base this on. Rather than just a giant string literal, that PR will give you a new module named `memray._attach_callback` that we can put these changes in, to help make managing the global state easier.

The `memray._attach_callback.callback` function will be the entry point. It'll take both the arguments that are going to be passed to the `Tracker` constructor as well as the arguments determining which attach mode to use, and it will call other functions in the module to handle the tracker creation and destruction, as well as the threads to disable tracking when different thresholds are reached.

godlygeek:
I've pushed quite a few fixup commits here to make various improvements. Most of them are pretty small things. The biggest one is refactoring the thread for the heap size check so that it can be cancelled, like the duration check one can. I also moved the global state to be attributes of the `memray` module, as we discussed above.

@ivonastojanovic Please review my changes and let me know if you spot any mistakes, bugs, issues, or shortcomings.

godlygeek (approved):
At this point this LGTM, but since I made some fairly extensive tweaks here, I'd appreciate it if @pablogsal could give this a read before we land this and see if any issues jump out.

src/memray/commands/attach.py
```diff
@@ -304,6 +415,32 @@ def prepare_parser(self, parser: argparse.ArgumentParser) -> None:
...
+            if args.force:
+                parser.error("Can't use --stop-tracking with -f or --force")
+            if args.aggregate:
+                parser.error("Can't use --stop-tracking with --aggregate")
+            if args.native:
+                parser.error("Can't use --stop-tracking with --native")
+            if args.follow_fork:
+                parser.error("Can't use --stop-tracking with --follow-fork")
+            if args.trace_python_allocators:
+                parser.error("Can't use --stop-tracking with --trace-python-allocators")
```
pablogsal, line 434: Maybe

```python
incompatible_flags = {
    args.output: "--output",
    args.force: "--force",
    args.aggregate: "--aggregate",
    args.native: "--native",
    args.follow_fork: "--follow-fork",
    args.trace_python_allocators: "--trace-python-allocators",
    args.no_compress: "--no-compress",
}

if args.stop_tracking and any(incompatible_flags):
    conflicting_flags = [message for flag, message in incompatible_flags.items() if flag]
    parser.error(f"Can't use --stop-tracking with {', '.join(conflicting_flags)}")
```

pablogsal: @godlygeek what do you prefer?

pablogsal (approved):
Left a nit, otherwise LGTM

tests/unit/test_attach.py
```diff
@@ -18,4 +18,55 @@ def test_memray_attach_aggregated_without_output_file(
             main(["attach", "--aggregate", "1234"])
 
         captured = capsys.readouterr()
+        print("Error", captured.err)
         assert "Can't use aggregated mode without an output file." in captured.err
+
```
pablogsal, line 23: @godlygeek I assume testing these options may be too cumbersome but maybe we should do some smoke testing at least....

src/memray/commands/attach.py
godlygeek: Hm. Well, as it's implemented right now, we don't know whether `-o` or `--output` was used, but we print an error message that says "Can't use --stop-tracking with -o or --output".

If we went with your proposed approach, we'd get "Can't use --stop-tracking with --output" which I worry might confuse people if they used `-o` instead of `--output`.

I dunno. It's less verbose, but I'm not convinced it's better - and it's about the same amount of maintenance, I think.

tests/unit/test_attach.py
godlygeek: In the first version of this we weren't logging anything when detaching, and so it was essentially impossible to prove that the new code was reached - but now that we're logging, I suppose we could do some simple smoke tests that just check the stdio of the tracked process looking for "heap size has reached" and "seconds have elapsed". Think that's good enough?

src/memray/commands/attach.py
pablogsal: Ok, let's leave as it is 👍

tests/unit/test_attach.py
pablogsal: Yeah, that will certainly work

pablogsal: Check [de94a39](https://github.com/bloomberg/memray/pull/458/commits/de94a39591fd4f945f43dd98a3ed3256ef37fd0e)

godlygeek: Well, that doesn't pass CI 😅 

The more I think about this, the more I become convinced that this should be a separate subcommand instead. Rather than having a `memray attach --stop-tracking` that's mutually exclusive with almost every other `memray attach` argument, we should just have a separate `memray detach` command. At first I thought that wouldn't make sense because `memray detach` would still need to _attach_ to the process in order to uninstall the old tracker, but I think that's just me letting my knowledge of the underlying architecture and communication channels cloud my judgment. I think that a typical user would look at `memray attach` as attaching _Memray_ to the process, as opposed to gdb or lldb, and so users would find it perfectly intuitive for there to be a `memray detach` command that removes Memray from the process.

Let me know if you agree - if you do, I'll do the legwork of splitting this into a separate `memray detach` subcommand.

godlygeek:
OK, I've updated this PR to make `memray detach` a separate command, and to fix the small issue with your fixup that caused the CI to fail. Take a look when you can, @pablogsal

godlygeek:
Oh, and https://godlygeek.github.io/memray/attach.html#cli-reference shows how the docs changes render.

godlygeek:
After some offline discussions, @pablogsal and I have decided to drop `--heap-limit` from this PR. The name is misleading, since the implementation is looking at maximum RSS rather than at the size of the heap, and it refuses to work reliably on i686 for reasons we haven't managed to track down (operations that we expect to be raising `ru_maxrss` appear not to, though only on i686 and not on x86-64 or arm64).

Instead, our plan is to add a way to set these limits when `Tracker` is created and installed, and have it enforce the limits itself. That change will come in a future release, though. For now, we're going to remove the option from this PR to unblock the way for including the rest of this work in a release.

PR #459, opened 2023-09-12
Move the `memray attach` callback into a module

godlygeek:
Previously we were maintaining this as a string literal inside the
client-side `attach` code, but it's growing more complex, and we're
reaching the point where it would be advantageous to have it in a normal
module that linters and type checkers can analyze.

This commit is based on the code in #455, and will need to be rebased once we land that, but it's ready for review other than that. Marking it draft until #455 is landed.

### DIFF
--- a/src/memray/_attach_callback.py
+++ b/src/memray/_attach_callback.py
@@ -1,0 +1,34 @@
+import atexit
+
+import memray
+
+
+def deactivate_last_tracker() -> None:
+    global _last_tracker
+    tracker = _last_tracker
+
+    if tracker:
+        _last_tracker = None
+        tracker.__exit__(None, None, None)
+
+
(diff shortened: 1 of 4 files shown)


godlygeek:
Actually... I've realized this will break using a new version memray to attach to a Python process where an older version of memray is installed. Now, that's not really something we've ever officially supported, but it is something that has always mostly worked - the live TUI wouldn't work if the client and server use a different version of our record format, but I believe attaching a tracker that outputs to a capture file has always worked, and that's something that our users might be relying on without us realizing it. That's a strong argument for sticking with having the client inject the code to be executed, rather than having the server import it from its installed copy of memray.

LMMaster448 (approved):

godlygeek:
Let's not do this. It doesn't work as nicely as I thought it would.

PR #466, opened 2023-09-25
docs: Make flame graph docs reflect icicle default

tjmcewan:
Doc update to align with what appears to be a relatively new default flame graph orientation.

### DIFF
--- a/docs/flamegraph.rst
+++ b/docs/flamegraph.rst
@@ -5,54 +5,87 @@ The flame graph reporter generates an HTML file containing a flame graph
 representation of the allocations contributing to the tracked process's peak
 memory usage.
 
-Flame graphs are a way to visualize how your program is spending its time.
-A few important things about flame graphs:
+Flame graphs are a way to visualize where your program is spending its
+memory. A few important things about flame graphs:
 
 - The flame graph displays the superposition of all stack traces that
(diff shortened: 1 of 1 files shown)


godlygeek:
> Doc update to align with what appears to be a relatively new default flame graph orientation.

You'd think that, but nah - I'm pretty sure we've been using icicle orientation by default since our very first release, and this has been wrong forever 😅

Thanks for catching this! I'm going to go for a more extensive fix than you proposed, though. I've updated the first section to avoid absolute spatial references entirely, and only make references relative to the location of the root. Then, I've yanked up the section about flame graphs vs icicle graphs to right below that section, explaining the flame vs icicle distinction earlier, explaining that the root could appear in either of two positions, and making it clear that `memray flamegraph` produces icicle graphs by default despite its name. And I've updated the prose for the first of the two simple examples to make it clear that it's a flame graph as opposed to an icicle graph. I didn't bother for the second, since I think it's fair to assume anyone trying to read and interpret the second example flame graph will have already understood the first example.

Since I've made such extensive changes to your small original patch, I'm going to leave this for someone else to review my work.

https://godlygeek.github.io/memray/flamegraph.html shows this page with my proposed changes.

pablogsal (approved):

PR #467, opened 2023-09-26
Miscellaneous improvements to our documentation and help messages

godlygeek:
Before adding a `description=` to our `ArgumentParser`'s:
```text
$ memray parse --help
usage: memray parse [-h] results

positional arguments:
  results     Results of the tracker run

options:
  -h, --help  show this help message and exit

Please submit feedback, ideas, and bug reports by filing a new issue at https://github.com/bloomberg/memray/issues
```

After:
```text
$ memray parse --help
usage: memray parse [-h] results

Debug a results file by parsing and printing each record in it

positional arguments:
  results     Results of the tracker run

options:
  -h, --help  show this help message and exit

Please submit feedback, ideas, and bug reports by filing a new issue at https://github.com/bloomberg/memray/issues
```

Before setting `COLUMNS` in our `docs/conf.py`:
image

After:
image

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,5 +1,7 @@
 """Sphinx configuration file for memray's documentation."""
 
+import os
+
 # -- General configuration ------------------------------------------------------------
 
 # The name of a reST role (builtin or Sphinx extension) to use as the default role,
@@ -57,3 +59,10 @@ intersphinx_mapping = {
         (None,),
     ),
 }
+
+# -- Options for sphinx-argparse ------------------------------------------------------
+
+# Limit the width of usage messages. argparse defaults to the terminal width,
+# but we don't want different output depending on the terminal width where the
+# docs were built.
+os.environ["COLUMNS"] = "88"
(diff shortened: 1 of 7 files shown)


docs/conf.py
```diff
@@ -57,3 +59,10 @@
         (None,),
     ),
 }
+
+# -- Options for sphinx-argparse ------------------------------------------------------
+
+# Limit the width of usage messages. argparse defaults to the terminal width,
+# but we don't want different output depending on the terminal width where the
+# docs were built.
+os.environ["COLUMNS"] = "88"
```
gusmonod, line 68: Much better, thank you for this change it is indeed much clearer this way 👍🏼

gusmonod (approved):

docs/conf.py
godlygeek: I _think_ the last push to the docs was done manually by me, in a fullscreen terminal on a widescreen monitor, and I think that's why the usage summaries are currently a one-line mess with a scrollbar. I don't think we would have gotten a scrollbar had we let CI do an automated publish, because I don't think it would have wound up with the idea that the screen is so wide.

But also to be fair, I had no idea my environment was bleeding into the generated docs like this, so yeah, it's definitely much wiser to hardcode a sane value instead.

PR #490, opened 2023-11-03
Fix debugger output parsing in attach command

pablogsal:
When using lldb in macOS, the output of lldb is slighly different and we
are not properly matching that is what we expect. To avoid having to
rely on weird hacks to not detect the actual command we issue, filter
the output to not contain the lines that have the debugger promt to only
check against the actual debugger output. This simplifies the check and
provides an easier way to do assert that we are in the situation we
expect.

**Describe your changes**

**Testing performed**

**Additional context**

### DIFF
--- a/news/490.bugfix.rst
+++ b/news/490.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a bug that was caussing attaching to fail in macOS Sonoma.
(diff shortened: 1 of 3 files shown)


godlygeek (approved):
LGTM

PR #491, opened 2023-11-03
Limit the number of memory records reported

pablogsal:


src/memray/_memray.pyx
```diff
@@ -915,6 +919,10 @@ cdef class FileReader:
                     self._memory_snapshots.push_back(reader.getLatestMemorySnapshot())
                 else:
                     break
+
+        if len(self._memory_snapshots) > max_memory_records:
+            self._memory_snapshot_bucket = len(self._memory_snapshots) // max_memory_records
+            self._memory_snapshots = self._memory_snapshots[::self._memory_snapshot_bucket]
```
godlygeek, line 925: Why is `self._memory_snapshot_bucket` an instance attribute? It seems like it's only set in this one place, and only used on the next line. Can't it be a local variable?

Also... this math doesn't seem to work. Let's say `max_memory_records == 10_000`, and `len(self._memory_snapshots) == 19_999`, then `len(self._memory_snapshots) // max_memory_records` will be `math.floor(1.9999)`, and so we'll use `1` as the stride and copy every record and still have just under twice as many records as the supposed max.

If we want to use a constant stride, I think we need a `math.ceil` in here:
```suggestion
            stride = math.ceil(len(self._memory_snapshots) / max_memory_records)
            self._memory_snapshots = self._memory_snapshots[::stride]
```

But using a constant stride might not be optimal... If we've got 10,001 snapshots, we'd need to choose between using a stride of 1 (leaving us with 10,001 snapshots, more than the maximum) or using a stride of 2 (leaving us with 5,000 snapshots, only _half_ of the maximum).

An alternative approach would be seeing how many milliseconds elapsed between the timestamp in the first snapshot and the last snapshot, dividing that total number of elapsed milliseconds by our maximum number of snapshots to retain, and keeping only the first snapshot for each value of `(curr_snapshot.time - first_snapshot.time) // granularity`.  That is, conceptually: divide the timeline into 10k buckets, and keep only the first snapshot that falls into each bucket.

Something like:
```py
        cdef vector[_MemorySnapshot] temp_snapshots
        if max_memory_records == 0:
            self._memory_snapshots.resize(0)
        elif len(self._memory_snapshots) > max_memory_records:
            delta = (
                self._memory_snapshots.back().ms_since_epoch
                - self._memory_snapshots.front().ms_since_epoch
            )
            granularity = delta // max_memory_records
            next_snapshot_time = self._memory_snapshots[0].ms_since_epoch

            temp_snapshots.reserve(max_memory_records)
            for snapshot in self._memory_snapshots:
                if snapshot.ms_since_epoch >= next_snapshot_time:
                    temp_snapshots.push_back(snapshot)
                    next_snapshot_time += granularity
```

That compiles and passes the test suite, but I haven't really invested any effort in exercising it thoroughly.

src/memray/commands/common.py
```diff
@@ -127,9 +127,13 @@ def write_report(
         merge_threads: Optional[bool] = None,
         inverted: Optional[bool] = None,
         temporal: bool = False,
+        max_memory_records: int = 0,
     ) -> None:
         try:
-            reader = FileReader(os.fspath(result_path), report_progress=True)
+            kwargs = {}
+            if max_memory_records > 0:
+                kwargs["max_memory_records"] = max_memory_records
```
godlygeek, line 135: Huh - so you're making it impossible to set the max to 0, even though the parameter defaults to 0? That seems... weird. I'd have expected 0 to mean "don't include any memory records", and I'd have expected a different sentinel for "unlimited" (I'd use `None`, because `max_memory_records=None` does seem to read as "no limit" to me, but I could live with `-1` as a sentinel if you don't want to make this `Optional`)

src/memray/_memray.pyi
```diff
@@ -133,7 +133,10 @@ class FileReader:
     @property
     def metadata(self) -> Metadata: ...
     def __init__(
-        self, file_name: Union[str, Path], *, report_progress: bool = False
+        self,
+        file_name: Union[str, Path],
+        *,
+        report_progress: bool = False,
```
godlygeek, line 139: Didn't you mean to add `max_memory_records` here?

godlygeek (changes requested):

src/memray/_memray.pyi
pablogsal: Yup, something went wrong with the rebase :(

src/memray/commands/common.py
pablogsal: Yeah I agree, let's mark this as Optional as use None to mark the default

pablogsal: Also, note the default is set to a very big number, not "No limit"

src/memray/_memray.pyx
pablogsal: Talked offline about this, we will go with the simpler stride-based approach here. The PR was missing some updates that I didn't include after a rebase so that's why the `_memory_snapshot_bucket` was an instance attribute.

src/memray/_memray.pyx
```diff
@@ -1100,6 +1110,7 @@ cdef class FileReader:
 
         cdef AllocationLifetimeAggregator aggregator
         cdef _Allocation allocation
+        cdef int memory_records_seen
```
godlygeek, line 1113: Doesn't this need to be explicitly initialized to 0? I don't think Cython value-initializes local variables like it does instance attributes, and I can't immediately spot any place in the generated C++ code where this is getting set to 0.

src/memray/_memray.pyx
```diff
@@ -855,8 +856,9 @@ cdef class FileReader:
     cdef HighWatermark _high_watermark
     cdef object _header
     cdef bool _report_progress
+    cdef int _memory_snapshot_stride
```
godlygeek, line 859: I wonder if we should make this an `ssize_t` instead of an `int`... it probably doesn't matter too much (a program would need to run for more than 6 months before the number of generated memory snapshots was bigger than int max), but it'd be slightly more correct.

src/memray/_memray.pyx
```diff
@@ -917,6 +923,10 @@ cdef class FileReader:
                     self._memory_snapshots.push_back(reader.getLatestMemorySnapshot())
                 else:
                     break
+
+        if len(self._memory_snapshots) > max_memory_records:
+            self._memory_snapshot_stride = int(ceil(len(self._memory_snapshots) / max_memory_records))
```
godlygeek, line 928: This isn't working the way we want it to - I think we need to cast one of these to a double. The generated code contains:
```cpp
__pyx_t_9 = PyObject_Length(__pyx_t_6);
...
__pyx_v_self->_memory_snapshot_stride = ((int)ceil((__pyx_t_9 / ((Py_ssize_t)__pyx_v_max_memory_records))));
```
So this is dividing one `size_t` by another (truncating to an integer) and then calling `ceil` on that integer, always returning the same integer.

godlygeek (changes requested):

src/memray/_memray.pyx
pablogsal: Yeah, I don't mind to settle with size_t

godlygeek (approved):
LGTM

PR #496, opened 2023-11-13
Add a "memory overview" section to the docs

pablogsal:
Many users find resident vs heap memory confusing and they struggle to
analyze properly programs where understanding the difference between the
two is important. Furthermore, is very common that internal
allocator-specific behaviours confuse users when they need to understand
what is causing the resident size to increase or decrease.

To help users understand these problems and concepts, add a new memory
overview section to the docs that covers most typical concepts and
problems when analyzing memory.


godlygeek (changes requested):

docs/memory.rst
```diff
@@ -0,0 +1,186 @@
...
+------------------------
+
+Another thing that makes it difficult to relate heap memory and resident memory is
+that memory can be fragmented. This means that the memory that is allocated by
+the system allocator may be spread around the address space in different
+fragments. This means that the resident memory size will increase/decrease in
+unpredictable ways. This happens because the system allocator may not be able to
+reuse memory that has been freed before. This is a common problem in languages
+like C or C++ where you need to manually call the deallocation API
+(``free`` for example) but it can also happen in languages like Python.
```
pablogsal, line 124:
```suggestion
reuse memory that has been freed before.

godlygeek (approved):
LGTM

PR #500, opened 2023-11-24
Make minor performance improvements using ruff rules C4 and PERF

cclauss:

**Describe your changes**
Make minor performance improvements using [ruff](https://docs.astral.sh/ruff/) rules [C4](https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4) and [PERF](https://docs.astral.sh/ruff/rules/#perflint-perf).
```
9	PERF102	[*] When using only the keys of a dict use the `keys()` method
9	PERF401	[ ] Use a list comprehension to create a transformed list
4	C401   	[*] Unnecessary generator (rewrite as a `set` comprehension)
4	C402   	[*] Unnecessary generator (rewrite as a `dict` comprehension)
4	C408   	[*] Unnecessary `dict` call (rewrite as a literal)
3	C405   	[*] Unnecessary `list` literal (rewrite as a `set` literal)
1	C416   	[*] Unnecessary `dict` comprehension (rewrite using `dict()`)
```

**Testing performed**
% `ruff --select=C4,PERF .`

**Additional context**

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
       - id: rst-inline-touching-normal
 
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: "v0.0.262"
+    rev: "v0.1.6"
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,13 +11,16 @@ build-backend = 'setuptools.build_meta'
 
 [tool.ruff]
 line-length = 95
-select = ["F", "E", "W", "I001"]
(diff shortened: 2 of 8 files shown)


godlygeek:
Everything under benchmarks/ is vendored code that we don't want to modify - please revert the changes to those files. We should instead configure ruff to not lint that directory. Also, we should configure ruff to always apply the C4 and PERF rules if we're going to make these changes, since otherwise there's nothing stopping us from regressing on them.

All the changes to our source code and tests seem fine, though. I think some of the dictionary literals are a bit less readable than the calls to the dict() constructor they're replacing, but the difference is probably not enough to worry about.

godlygeek (approved):
LGTM! Thanks for the contribution, @cclauss!

PR #507, opened 2023-12-04
More tweaks for the tree reporter TUI

godlygeek:
- Use a gradient for rows in the table, rather than just 4 color buckets by size
- Hide the location and thread labels for the root node, since they don't apply to it
- Show 200 allocation locations in the tree by default, rather than the previous default of 10
- Add a node to the tree showing information about the allocations other than the 200 that made the cut. Selecting it gives information about how to rerun the reporter to include more of them.
- Update the tests to reflect these changes


src/memray/reporters/tree.py
```diff
@@ -75,41 +93,87 @@ async def update_text_area(self) -> None:
...
+
+        _, file, line = self.frame.location
         delta = text.size.height // 2
         lines = linecache.getlines(file)[line - delta : line + delta]
 
         text.text = "\n".join(tuple(line.rstrip() for line in lines))
         text.select_line((delta - 1))
         text.show_line_numbers = False
 
+    def _get_content_by_label_id(self) -> Dict[str, str]:
```
pablogsal, line 110: Can we confirm that we have a visual test that covers something where this text appears?

godlygeek: Yep, we have tests exercising all 3 cases of this. There's one for the root node (which most of our pre-existing snapshot tests capture), one for a normal frame node (which a few of our pre-existing snasphot tests exercise), and one for the artificial excluded allocations row (which a newly added snapshot test exercises)

pablogsal (approved):
LGTM

PR #510, opened 2023-12-14
Publish manylinux2010 wheels for Python 3.11/3.12

godlygeek:
We can't get this for free from the manylinux2010 image, because support for it was dropped before Python 3.11 came out. We can spin up a container based on that image, install Python 3.11 and 3.12 in it, and then ask cibuildwheel to use it for building some wheels for us (and running `auditwheel repair` and tests).

We'll do it only for x86_64, not i686, since 64-bit is much more heavily used.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -52,14 +52,40 @@ jobs:
 
   build_linux_wheels:
     needs: [build_sdist, choose_linux_wheel_types]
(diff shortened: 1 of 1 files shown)


pablogsal:
LGTM

Hopefully we don't need to maintain this for long as it's quite involved. I reviewed and also tested this locally and in a container so I am confident it works but it's indeed a bit complex.

PR #513, opened 2023-12-15
Update `upload-artifact` and `download-artifact`

gusmonod:
Replace #511 and #512

PR #516, opened 2024-01-03
Update `upload-artifact` and `download-artifact`: one artifact per OS + arch + platform tag series

gusmonod:
Another approach for #513 where we use one artifact per OS + architecture instead of one per wheel.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -24,12 +24,12 @@ jobs:
       - name: Build sdist
         run: pipx run build --sdist
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
-          name: dist
+          name: sdist
           path: dist
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
           name: tests
           path: tests
@@ -50,22 +50,20 @@ jobs:
 
   build_wheels:
     needs: [build_sdist, choose_architectures]
-    name: Wheel for Linux-${{ matrix.cibw_python }}-${{ matrix.cibw_arch }}
+    name: Wheel for Linux ${{ matrix.cibw_arch }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        cibw_python:
-          ["cp37-*", "cp38-*", "cp39-*", "cp310-*", "cp311-*", "cp312-*"]
         cibw_arch: ${{ fromJSON(needs.choose_architectures.outputs.cibw_arches) }}
 
     steps:
-      - uses: actions/download-artifact@v3
+      - uses: actions/download-artifact@v4
         with:
-          name: dist
+          name: sdist
           path: dist
-      - uses: actions/download-artifact@v3
+      - uses: actions/download-artifact@v4
         with:
           name: tests
           path: tests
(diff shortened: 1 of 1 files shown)


gusmonod:
@godlygeek this is also ready to review. A test in Alpine is failing, but I am pretty confident it has nothing to do with this PR.

godlygeek:
> A test in Alpine is failing, but I am pretty confident it has nothing to do with this PR.

Passed on re-run. Yeah, I've seen that fail a few times - a segfault in the `memray attach` tests... 😔

.github/workflows/build_wheels.yml
```diff
@@ -123,32 +121,33 @@ jobs:
...
           CIBW_TEST_EXTRAS: test
           CIBW_TEST_COMMAND: pytest {package}/tests
           CIBW_BUILD_VERBOSITY: 1
           MACOSX_DEPLOYMENT_TARGET: "10.14"
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
-          name: dist
+          name: macos-${{ matrix.cibw_arch }}-wheels
```
gusmonod, line 134:
```suggestion
          name: macosx_${{ matrix.cibw_arch }}-wheels
```

.github/workflows/build_wheels.yml
```diff
@@ -81,34 +80,33 @@ jobs:
...
+      - uses: actions/upload-artifact@v4
         with:
-          name: dist
+          name: ${{ matrix.wheel_type }}-wheels
           path: ./wheelhouse/*.whl
 
   build_wheels_macos:
     needs: [build_sdist]
-    name: Wheel for MacOS-${{ matrix.cibw_python }}-${{ matrix.cibw_arch }}
+    name: ${{ matrix.cibw_arch }} wheels
```
gusmonod, line 96:
```suggestion
    name: macosx_${{ matrix.cibw_arch }} wheels
```

gusmonod (commented):
I would change this to use `macosx_$architecture` as is the convention in wheel names

.github/workflows/build_wheels.yml
gusmonod: Without these changes, we see the following in PR checks:

```
Wheels / x86_64 wheels (pull_request)
Wheels / arm64 wheels (pull_request)
Wheels / manylinux_x86_64 wheels (pull_request)
Wheels / manylinux_aarch64 wheels (pull_request)
```

I think this is confusing.

.github/workflows/build_wheels.yml
```diff
@@ -123,32 +120,33 @@ jobs:
...
-      - uses: actions/download-artifact@v3
+      - uses: actions/download-artifact@v4
         with:
-          name: dist
+          # with no name set, it downloads all of the artifacts
           path: dist
-      - uses: pypa/gh-action-pypi-publish@release/v1
-        with:
-          skip_existing: true
-          password: ${{ secrets.PYPI_PASSWORD }}
```
gusmonod, line 151: This still needs to be put back before merging the PR

godlygeek:
OK, I think this should be good to go (other than updating the required status checks, which I'll do last). Double check the work on my commits for me, please, @gusmonod! 🚀

.github/workflows/build_wheels.yml
```diff
@@ -130,12 +130,20 @@ jobs:
 
       - uses: actions/upload-artifact@v4
         with:
-          name: macos-${{ matrix.cibw_arch }}-wheels
+          name: macosx_${{ matrix.cibw_arch }}-wheels
           path: ./wheelhouse/*.whl
 
+  build_and_test_wheels:
```
gusmonod, line 136: Maybe we could add a comment explaining why it is useful to have that job there, since it literally does nothing except making it easier to set required checks.

Perhaps we could call it something different? Maybe something like `wait_for_required_jobs`.

godlygeek: I added the comment, but as we discussed offline I left the name the same because it would be strange to have:
```yaml
  upload_pypi:
    needs: [wait_for_required_jobs, build_sdist]
```
instead of
```yaml
  upload_pypi:
    needs: [build_and_test_wheels, build_sdist]
```

godlygeek (approved):
@gusmonod's commit looks good to me, and he's confirmed my commits look good to him, so let's land this.

PR #521, opened 2024-01-09
ci: Fix coverage reporting on Alpine

godlygeek:
Coverage.py sometimes reports events with the line number as 0. We need to ignore these.


gusmonod (approved):
LGTM

PR #523, opened 2024-01-10
Create memray.1

alexmyczko:
*Issue number of the reported bug or feature request: #<number>*

**Describe your changes**
add a  manual page

**Testing performed**
yes `man ./memray.1`

**Additional context**
classical systems supply a printable manual to commands with options and their description

PR #524, opened 2024-01-12
docs: Allow producing a `memray.1` man page

godlygeek:
This man page will contain a small bit of prose explaining what Memray is and where to find the full documentation, followed by CLI usage information scraped from the `--help` of each of our subcommands. Distros can build the man page for any given version to include in distro packages of Memray.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,6 +124,7 @@ lint:  ## Lint all files
 docs:  ## Generate documentation
 	$(MAKE) -C docs clean
 	$(MAKE) -C docs html
+	$(MAKE) -C docs man
 
 .PHONY: docs-live
 docs-live:  ## Serve documentation on localhost:8000, with live-reload
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -25,6 +25,12 @@ html:
 	@echo
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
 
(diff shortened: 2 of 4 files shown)


godlygeek:
https://gist.github.com/godlygeek/e3ab347806010596cab9784ccc33f1a8 shows the memray.1 that's generated by running `make -C docs man` from the repo root (assuming that a virtualenv has been activated and `pip install -r requirements-docs.txt` has been run to install the docs generation dependencies).

@alexmyczko Can you check if that man page meets your needs? It's more informative than what you propose in #523 and it shouldn't require too much more work to keep up to date.

gusmonod:
Should we exercise this in CI in some way? At least to ensure that the manual page can be generated and isn't empty for example.

godlygeek:
> Should we exercise this in CI in some way?

Yes, let's.

> At least to ensure that the manual page can be generated and isn't empty for example.

I'll make sure it's generated. I'm not gonna check for emptiness - we don't do that for the HTML output, so doing it for the man output seems superflous. But I will update `make docs` to build the man page, and CI already tests that `make docs` succeeds.

Let's not worry about the "empty output" case unless we ever find a case where Sphinx generates an empty file and returns a success rcode to us. 😃

gusmonod (approved):
LGTM

Issue #212, opened 2022-09-16
Could not load library libcudnn_ops_infer.so.8.


kadirnar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Code: 
```
import yolov5

def pip_load():
    return yolov5.load('yolov5s.pt')

pip_load()
```
Error Message:
```
Writing profile results into speed/memray-pip_speed.py.15691.bin
Memray WARNING: Correcting symbol for aligned_alloc from 0x7f6d26393cc0 to 0x7f6d279a4250
Could not load library libcudnn_ops_infer.so.8. Error: libcudnn_ops_infer.so.8: cannot open shared object file: No such file or directory
Please make sure libcudnn_ops_infer.so.8 is in your library path!
```

### Steps To Reproduce

```
pip install memray
```
Code :
```
memray run file.py 
```

### Memray Version

1.3.1

### Python Version

3.8

### Operative System

Linux

pablogsal:
Thanks @kadirnar for opening an issue.

This doesn't seem like a `memray` error. Can you paste here what is the output of the program if you run `python file.py` and `python -m memray file.py`?

kadirnar:
I solved the problem thank you.

ash2703:
@kadirnar Could you please post the solution, faving same issue when profiling pytorch model

bnawras:
@pablogsal 

- Ubuntu 20.04.5 
- python3.8
- memray 1.11.0

```
$ python3.8 -m memray train.py

Memray WARNING: Correcting symbol for malloc from 0x425490 to 0x7fd52516e0e0
Memray WARNING: Correcting symbol for free from 0x425910 to 0x7fd52516e6d0
Memray WARNING: Correcting symbol for aligned_alloc from 0x7fd524941ca0 to 0x7fd52516f250
Could not load library libcudnn_ops_infer.so.8. Error: libcudnn_ops_infer.so.8: cannot open shared object file: No such file or directory
Aborted (core dumped)
```

bnawras:
### Dockerfile
```
FROM nvidia/cuda:11.1.1-runtime-ubuntu20.04

ENV DEBIAN_FRONTEND noninteracrive
RUN apt-get update
RUN apt-get upgrade -y
RUN apt-get install -y \
        build-essential git python3 python3-pip \
        ffmpeg libsm6 libxext6 libxrender1 libglib2.0-0

WORKDIR /app
COPY requirements.txt .
[...]

PR #525, opened 2024-01-16
Override dlsym instead of dlopen to correctly honour RPATH/RUNPATHS

pablogsal:
We need to override dlopen/dlclose to account for new shared libraries
being loaded in the process memory space. This is needed so we can
correctly track allocations in those libraries by overriding their PLT
entries and also so we can properly map the addresses of the symbols in
those libraries when we resolve later native traces. Unfortunately, we
can't just override dlopen directly because of the following edge case:
when a shared library dlopen's another by name (e.g.
dlopen("libfoo.so")), the dlopen call will honor the RPATH/RUNPATH of
the calling library if it's set. Some libraries set an RPATH/RUNPATH
based on $ORIGIN (the path of the calling library) to load dependencies
from a relative directory based on the location of the calling library.
This means that if we override dlopen, we'll end up loading the library
from the wrong path or more likely, not loading it at all because the
dynamic loader will think the memray extenion it's the calling library
and the RPATH of the real calling library will not be honoured.

To work around this, we override dlsym instead and override the symbols
in the loaded libraries only the first time we have seen a handle passed
to dlsym. This works because for a symbol from a given dlopen-ed library
to appear in a call stack, *something* from that library has to be
dlsym-ed first. The only exception to this are static initializers, but
we cannot track those anyway by overriding dlopen as they run within the
dlopen call itself.

Closes: #212

### DIFF
--- a/src/memray/_memray/hooks.cpp
+++ b/src/memray/_memray/hooks.cpp
@@ -1,5 +1,7 @@
 #include <cassert>
 #include <cstdio>
+#include <mutex>
+#include <unordered_set>
 
 #include "hooks.h"
 #include "tracking_api.h"
@@ -304,21 +306,75 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
     return ret;
 }
 
-void*
-dlopen(const char* filename, int flag) noexcept
+// We need to override dlopen/dlclose to account for new shared libraries being
+// loaded in the process memory space. This is needed so we can correctly track
+// allocations in those libraries by overriding their PLT entries and also so we
+// can properly map the addresses of the symbols in those libraries when we
+// resolve later native traces. Unfortunately, we can't just override dlopen
+// directly because of the following edge case: when a shared library dlopen's
+// another by name (e.g. dlopen("libfoo.so")), the dlopen call will honor the
+// RPATH/RUNPATH of the calling library if it's set. Some libraries set an
+// RPATH/RUNPATH based on $ORIGIN (the path of the calling library) to load
+// dependencies from a relative directory based on the location of the calling
+// library. This means that if we override dlopen, we'll end up loading the
+// library from the wrong path or more likely, not loading it at all because the
+// dynamic loader will think the memray extenion it's the calling library and
+// the RPATH of the real calling library will not be honoured.
+//
+// To work around this, we override dlsym instead and override the symbols in
+// the loaded libraries only the first time we have seen a handle passed to
+// dlsym. This works because for a symbol from a given dlopen-ed library to
+// appear in a call stack, *something* from that library has to be dlsym-ed
+// first. The only exception to this are static initializers, but we cannot
+// track those anyway by overriding dlopen as they run within the dlopen call
+// itself.
+// There's another set of cases we would miss: if library A has a static initializer
+// that passes a pointer to one of its functions to library B, and library B stores
+// that function pointer, then we could see calls into library A via the function pointer
+// held by library B, even though dlsym was never called on library A. This should be
+// very rare and will be corrected the next time library B calls dlsym so this should
+// not be a problem in practice.
+
+class DlsymCache
 {
-    assert(hooks::dlopen);
+  public:
+    auto insert(const void* handle)
+    {
+        std::unique_lock lock(mutex_);
+        return d_handles.insert(handle);
+    }
+
+    void erase(const void* handle)
+    {
+        std::unique_lock lock(mutex_);
+        d_handles.erase(handle);
+    }
+
+  private:
+    mutable std::mutex mutex_;
+    std::unordered_set<const void*> d_handles;
+};
+
+static DlsymCache dlsym_cache;
+
+void*
+dlsym(void* handle, const char* symbol) noexcept
+{
+    assert(hooks::dlsym);
     void* ret;
     {
         tracking_api::RecursionGuard guard;
-        ret = hooks::dlopen(filename, flag);
+        ret = hooks::dlsym(handle, symbol);
     }
     if (ret) {
-        tracking_api::Tracker::invalidate_module_cache();
-        if (filename
-            && (nullptr != strstr(filename, "/_greenlet.") || nullptr != strstr(filename, "/greenlet.")))
-        {
-            tracking_api::Tracker::beginTrackingGreenlets();
+        auto [_, inserted] = dlsym_cache.insert(handle);
+        if (inserted) {
+            tracking_api::Tracker::invalidate_module_cache();
+            if (symbol
+                && (0 == strcmp(symbol, "PyInit_greenlet") || 0 == strcmp(symbol, "PyInit__greenlet")))
+            {
+                tracking_api::Tracker::beginTrackingGreenlets();
+            }
         }
     }
     return ret;
@@ -334,6 +390,7 @@ dlclose(void* handle) noexcept
         tracking_api::RecursionGuard guard;
         ret = hooks::dlclose(handle);
     }
+    dlsym_cache.erase(handle);
     tracking_api::NativeTrace::flushCache();
     if (!ret) tracking_api::Tracker::invalidate_module_cache();
     return ret;
--- a/src/memray/_memray/hooks.cpp
+++ b/src/memray/_memray/hooks.cpp
@@ -90,14 +90,14 @@ isDeallocator(const Allocator& allocator)
     __builtin_unreachable();
 }
 
-#define FOR_EACH_HOOKED_FUNCTION(f) SymbolHook<decltype(&::f)> f(#f, &::f);
+#define FOR_EACH_HOOKED_FUNCTION(f) SymbolHook<decltype(&::f)> MEMRAY_ORIG_NO_NS(f)(#f, &::f);
 MEMRAY_HOOKED_FUNCTIONS
 #undef FOR_EACH_HOOKED_FUNCTION
 
 void
 ensureAllHooksAreValid()
 {
-#define FOR_EACH_HOOKED_FUNCTION(f) f.ensureValidOriginalSymbol();
+#define FOR_EACH_HOOKED_FUNCTION(f) MEMRAY_ORIG(f).ensureValidOriginalSymbol();
     MEMRAY_HOOKED_FUNCTIONS
 #undef FOR_EACH_HOOKED_FUNCTION
 }
@@ -166,12 +166,12 @@ pymalloc_free(void* ctx, void* ptr) noexcept
 void*
 malloc(size_t size) noexcept
 {
-    assert(hooks::malloc);
+    assert(MEMRAY_ORIG(malloc));
 
     void* ptr;
     {
         tracking_api::RecursionGuard guard;
-        ptr = hooks::malloc(size);
+        ptr = MEMRAY_ORIG(malloc)(size);
     }
     if (ptr) {
         tracking_api::Tracker::trackAllocation(ptr, size, hooks::Allocator::MALLOC);
@@ -182,7 +182,7 @@ malloc(size_t size) noexcept
 void
 free(void* ptr) noexcept
 {
-    assert(hooks::free);
+    assert(MEMRAY_ORIG(free));
 
     // We need to call our API before we call the real free implementation
     // to make sure that the pointer is not reused in-between.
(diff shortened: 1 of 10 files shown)


src/memray/commands/_attach.lldb
```diff
@@ -15,7 +15,7 @@ breakpoint set -b malloc -b calloc -b realloc -b free -b PyMem_Malloc -b PyMem_C
 # Set commands to execute when breakpoint is reached
 breakpoint command add -e true
 breakpoint disable
-expr auto $dlsym = (void* (*)(void*, const char*))&dlsym
+expr auto $dlsym = (void* (*)(void*, const char*))dlsym.d_original
```
pablogsal, line 18: This is extremely annoying but seems that `lldb` on Linux resolves `dlsym` to our hook and in macOS it resolves it to the real `dlsym`. I have played wit some alternatives but I cannot make a single script work in both (I also tried fully namespacing it, which fails in macOS because it doesn't resolve `memray`).

@godlygeek Ideas welcomed ;)

src/memray/_memray/hooks.cpp
```diff
@@ -304,19 +306,76 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
+    }
+
+    bool lookup(const void* handle) const
+    {
+        std::shared_lock lock(mutex_);
+        auto it = d_handles.find(handle);
+        if (it != d_handles.end()) {
+            return true;
+        }
+        return false;
```
godlygeek, line 348:
```suggestion
        return d_handles.find(handle) != d_handles.end();
```

src/memray/_memray/hooks.cpp
```diff
@@ -304,19 +306,76 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
+    assert(hooks::dlsym);
     void* ret;
     {
         tracking_api::RecursionGuard guard;
-        ret = hooks::dlopen(filename, flag);
+        ret = hooks::dlsym(handle, symbol);
     }
-    if (ret) {
+    if (ret && !dlsym_cache.lookup(handle)) {
+        dlsym_cache.addEntry(handle);
```
godlygeek, line 374: Hm, this is a TOCTOU issue - 2 threads can both get a false return from lookup and then both try to insert the same handle, and both think they succeeded.

I think instead we should have an `insert` that returns a `bool` indicating whether it succeeded or failed because the thing already existed. Which is the contract of `std::set<T>::insert`, heh.

src/memray/_memray/hooks.cpp
```diff
@@ -304,19 +306,76 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
         tracking_api::Tracker::invalidate_module_cache();
-        if (filename
-            && (nullptr != strstr(filename, "/_greenlet.") || nullptr != strstr(filename, "/greenlet.")))
+        if (symbol
+            && (nullptr != strstr(symbol, "PyInit_greenlet")
+                || nullptr != strstr(symbol, "PyInit__greenlet")))
```
godlygeek, line 378: This is gross, but it might actually be slightly less gross than the original version, heh.

But don't we want strcmp instead of strstr?
```suggestion
            && (0 == strcmp(symbol, "PyInit_greenlet")
                || 0 == strcmp(symbol, "PyInit__greenlet")))
```

src/memray/commands/_attach.lldb
godlygeek: This fails if `memray` hasn't already been imported:
```
error: Aborting reading of commands after command #1: 'expr auto $dlsym = (void* (*)(void*, const char*))dlsym.d_original'
failed with error: <user expression 11>:1:46: 'dlsym' has unknown type; cast it to its declared type to use it
    1 | auto $dlsym = (void* (*)(void*, const char*))dlsym.d_original
      |                                              ^~~~~
```
The only reason it works for the test suite is that the test program does `from memray._test import MemoryAllocator` which imports `memray._memray` as a side effect, heh. `memray._test` is in the `memray` package, which runs `memray/__init__.py`, which imports `memray._memray`.

godlygeek: I _think_ this is happening because we're violating ODR, actually.

We're using `decltype` to declare new functions with the same signature as the functions we're hooking, but `extern "C"` is part of the type, so even though we're putting these things in a namespace, they're mangling to just the original name!

I think the right fix here is namespacing our hooks the C way, by prefixing `memray_` before the original function name.

godlygeek:
Hm. What I pushed seems to work on glibc Linux and macOS, but is failing on musl linux. I'll probably need to spin up a container to debug that. 😔

src/memray/_memray/hooks.cpp
```diff
@@ -304,19 +306,76 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
+// dynamic loader will think the memray extenion it's the calling library and
+// the RPATH of the real calling library will not be honoured.
+//
+// To work around this, we override dlsym instead and override the symbols in
+// the loaded libraries only the first time we have seen a handle passed to
+// dlsym. This works because for a symbol from a given dlopen-ed library to
+// appear in a call stack, *something* from that library has to be dlsym-ed
+// first. The only exception to this are static initializers, but we cannot
+// track those anyway by overriding dlopen as they run within the dlopen call
+// itself.
```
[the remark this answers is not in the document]
pablogsal: I added a paragraph mentioning this at the end of the comment

src/memray/_memray/hooks.cpp
```diff
@@ -304,19 +306,76 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
+class DlsymCache
+{
+  public:
+    void addEntry(const void* handle)
+    {
+        std::unique_lock lock(mutex_);
+        d_handles.insert(handle);
+    }
+
+    bool lookup(const void* handle) const
```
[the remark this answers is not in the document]
pablogsal: I am going to nuke this function and have only a protected "insert"

tests/integration/rpath_extension/setup.py
```diff
@@ -0,0 +1,32 @@
...
+    name="ext",
+    version="1.0",
+    ext_modules=[
+        Extension(
+            "ext",
+            sources=["ext.c"],
+            extra_link_args=["-Wl,-rpath,$ORIGIN/sharedlibs"],
+        )
+    ],
+    cmdclass={"build_ext": CustomBuildExt},
```
godlygeek, line 31: Looks like this (and `CustomBuildExt` itself) are no longer needed, since you're using a `subprocess.run()` at module scope to do the building of the .so

tests/integration/test_extensions.py
```diff
@@ -354,3 +355,34 @@ def test_hard_exit(tmpdir, py_finalize):
 
     # THEN
     # No assertions, just check that the program exits without error
+
+
+@pytest.mark.skipif(
+    sys.platform == "darwin", reason="Test requires a linker that supports -rpath"
```
godlygeek, line 361: Shouldn't this be:
```suggestion
    sys.platform == "darwin", reason="Test requires a linker that supports $ORIGIN"
```

tests/integration/test_extensions.py
```diff
@@ -354,3 +355,34 @@ def test_hard_exit(tmpdir, py_finalize):
...
+
+    # WHEN
+    with monkeypatch.context() as ctx:
+        ctx.setattr(sys, "path", [*sys.path, str(extension_path)])
+        from ext import hello_world  # type: ignore
+
+        try:
+            hello_world()
+        except RuntimeError:
+            pytest.skip("Test requires a linker that supports -rpath")
```
godlygeek, line 384: Hm - is this still necessary given the `@pytest.mark.skipif` above for Darwin? What platform do we support that's not Darwin but where `hello_world()` would fail?

I'm a bit worried that this would mask real errors (turning them into skips instead of test failures) if we screw something up in the future.

src/memray/_memray/hooks.cpp
```diff
@@ -304,21 +306,75 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
+class DlsymCache
+{
+  public:
+    auto insert(const void* handle)
+    {
+        std::unique_lock lock(mutex_);
+        return d_handles.insert(handle);
+    }
+
+    void deleteEntry(const void* handle)
```
godlygeek, line 347: Tiny nit, but perhaps we should call this `erase` instead, to make it clearer that the semantics of this function are the same as set::set::erase just like the above one returns the same pair as std::set::insert does.

src/memray/_memray/hooks.cpp
```diff
@@ -1,5 +1,7 @@
 #include <cassert>
 #include <cstdio>
+#include <mutex>
+#include <unordered_set>
```
godlygeek, line 4: Hm - you did `#include <unordered_set>` but used `std::set` in the implementation.

Either you included the wrong thing or you used the wrong thing, I think.

src/memray/_memray/hooks.cpp
```diff
@@ -304,21 +306,75 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
     if (ret) {
-        tracking_api::Tracker::invalidate_module_cache();
-        if (filename
-            && (nullptr != strstr(filename, "/_greenlet.") || nullptr != strstr(filename, "/greenlet.")))
-        {
-            tracking_api::Tracker::beginTrackingGreenlets();
+        auto [_, inserted] = dlsym_cache.insert(handle);
+        if (inserted) {
+            tracking_api::Tracker::invalidate_module_cache();
+            if (symbol
```
godlygeek, line 373: I don't mind it, but just to check my own understanding: do we actually need `if (symbol)` here? Is there a case where `dlsym` is allowed to return something non-null despite a null pointer being passed for `symbol`?

godlygeek (commented):
I've got a few small nits, but otherwise this looks good to me.

In addition to these, would you mind squashing the first two commits together (and leaving the 3rd and 4th separate)? Commit 2 just un-does something done in commit 1, mostly (and adds a `&::` that wasn't there before). Maybe update the commit message to mention the rationale for the `&::` to force a lookup in the global namespace.

tests/integration/test_extensions.py
pablogsal: This is needed for old versions of the linker like the ones in the many Linux images.

godlygeek: Ah right, I remember now - I just didn't connect the dots.

OK, let's leave it then. I've been thinking about restructuring this CI to work like PyStack's, using cibuildwheel to build the wheels but _not_ test them, and downloading the built wheels onto more modern containers and testing them there (both to reintroduce a bit more parallelism, and to prove that the wheels are working correctly on newer distros).

If I make that change, I should be able to remove this at that point.

godlygeek: > I've been thinking about restructuring this CI to work like PyStack's, using cibuildwheel to build the wheels but _not_ test them, and downloading the built wheels onto more modern containers and testing them there (both to reintroduce a bit more parallelism, and to prove that the wheels are working correctly on newer distros).

Actually... I've investigated this a bit and decided not to do it, because it _won't_ introduce more parallelism - at least, not without switching back to multiple matrix jobs. As things stand today, I'd need to make a test_on_ubuntu job that depends on our build_linux_wheels job, but the build_linux_wheels job is a matrix job that doesn't finish until all its sub-jobs are done, so we wouldn't wind up testing any Linux wheels until we're done building all of them across all architectures, which is actually probably worse than the current behavior.

src/memray/_memray/hooks.cpp
pablogsal: I checked the glibc implementation and re-read the docs and seems that will never happen, so we can probably simplify this. Would you like me to do that?

src/memray/_memray/hooks.cpp
pablogsal: Makes sense

src/memray/_memray/hooks.cpp
godlygeek: eh, I don't care too much. It's harmless, I was just checking that I wasn't missing anything. Let's just leave it.

godlygeek (approved):
LGTM. I've squashed the fixups and rebased.

Issue #519, opened 2024-01-04
%%memray_flamegraph magic options


lcrmorin:
### Is there an existing proposal for this?

- [X] I have searched the existing proposals

### Is your feature request related to a problem?

I am trying to personnalise memray for more advanced usage. With the %%memray magic I can't (1) specify ouptut path (--output option) and (2) cannot specify I am mainly interested in the plots (--temporal ?). 

### Describe the solution you'd like

I'd like the %%memray magic to have the same options as the command line tool.

godlygeek:
I'd happily take a PR that adds `--temporal` support. `%%memray_flamegraph` has its own copy of the argument parsing from `src/memray/commands/common.py`, so this probably would just require copy-pasting the code for handling temporal mode from there into `src/memray/_ipython/flamegraph.py`, and seeing if we can add a test for it in `tests/integration/test_ipython.py`.

`--output-path` might be a little bit trickier, though. Currently `%%memray_flamegraph` is designed under the assumption that you want to display the generated flamegraph inside the Jupyter notebook. Are you saying that you'd like to have an option where _instead_ it writes that generated flame graph as a `.html` file on disk?

lcrmorin:
Thanks,
I'll look into this. For path maybe the way to go is to allow for customizing task id, now I get random strings (cell id maybe ?). 
Now I have another problem: I ran code overnight and memray seems to have bloated the disk. Any options to reduce memory size ? (typically sampling frequency ?)
Best,

godlygeek:
Memray is a tracing profiler, not a sampling profiler, so there's no concept of "sampling frequency" that you can apply. `memray run` itself supports a `--aggregate` mode that performs in-memory aggregation, resulting in higher memory usage and lower disk usage, but it's incompatible with the `--temporal` report you said you're interested in. Temporal reports inherently need point-in-time data about what is allocated at each moment in time, while the aggregated capture file format only retains information about what's on the heap at the high water mark and at the point when tracking is terminated.

In any event, it doesn't look like we added support for `--aggregate` to `%%memray_flamegraph`. That's an oversight - it probably should be the default. Maybe it shouldn't even be possible to turn it off... as things stand today, `%%memray_flamegraph` doesn't retain the capture file, only the flamegraph, and it can generate the flamegraph, both in `--leaks` mode and the default high water mark mode, with the smaller `--aggregate` capture file.

godlygeek:
[...]

PR #538, opened 2024-01-30
IPython improvements

### LINKED ISSUE #520: Make the `%%memray_flamegraph` IPython magic use aggregated capture files
See #519 for context. The `%%memray_flamegraph` IPython magic predates [aggregated capture files](https://bloomberg.github.io/memray/run.html#aggregated-capture-files), but considering that the capture file produced by `%%memray_flamegraph` is only used for the reports that aggregated capture files support, we should probably use aggregated capture files by default for the `%%memray_flamegraph` magic.

godlygeek:
Closes #520 
See #519

### DIFF
--- a/src/memray/_ipython/flamegraph.py
+++ b/src/memray/_ipython/flamegraph.py
@@ -107,6 +107,12 @@ def argument_parser() -> argparse.ArgumentParser:
         action="store_true",
         default=False,
     )
+    parser.add_argument(
+        "--max-memory-records",
+        help="Maximum number of memory records to display",
+        type=int,
+        default=None,
+    )
 
     return parser
 
@@ -149,7 +155,11 @@ class FlamegraphMagics(Magics):
 
         reporter = None
 
-        with FileReader(dump_file, report_progress=True) as reader:
+        kwargs = {}
+        if options.max_memory_records is not None:
+            kwargs["max_memory_records"] = options.max_memory_records
+
+        with FileReader(dump_file, report_progress=True, **kwargs) as reader:
             if reader.metadata.has_native_traces:
                 warn_if_not_enough_symbols()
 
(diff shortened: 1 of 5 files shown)


src/memray/_ipython/flamegraph.py
```diff
@@ -118,6 +133,15 @@ def memray_flamegraph(self, line: str, cell: str) -> None:
...
             return
 
+        if options.temporal and options.temporary_allocation_threshold >= 0:
+            raise UsageError(
+                "Can't create a temporal flame graph of temporary allocations"
+            )
+        elif options.temporal or options.temporary_allocation_threshold >= 0:
+            file_format = "ALL_ALLOCATIONS"
+        else:
+            file_format = "AGGREGATED_ALLOCATIONS"
```
pablogsal, line 143: Ah, this will be a nice improvement 👍  (I think we were not doing it already)

pablogsal (approved):

tests/integration/test_ipython.py
```diff
@@ -137,3 +137,35 @@ def test_passing_valid_argument(self, tmpdir, capsys):
...
+    @pytest.mark.parametrize(
+        "args,title",
+        [
+            ("--temporary-allocation-threshold=2", "flamegraph report"),
+            ("--leaks", "flamegraph report (memory leaks)"),
+            ("--temporal", "temporal flamegraph report"),
+            ("--leaks --temporal", "temporal flamegraph report (memory leaks)"),
+        ],
+    )
+    def test_report_title_by_report_type(self, tmpdir, capsys, args, title):
```
pablogsal, line 150: Maybe is worth checking the error when --temporal and `temporary_allocation_threshold>=0`

godlygeek: Ah, good call, can do.

src/memray/_ipython/flamegraph.py
godlygeek: 💯

pablogsal: Now that I think about it we could do it in the pytest plugin as well 🤔

godlygeek: > Now that I think about it we could do it in the pytest plugin as well 🤔

https://github.com/bloomberg/pytest-memray/pull/107

PR #543, opened 2024-02-02
Handle changes in Textual 0.48

godlygeek:
Textual 0.48 makes several changes to the `TextArea` that we need to adapt to
- Disable line wrapping
- Move from `DEFAULT_CSS` to using a CSS file (I'm not entirely sure why, but some of our CSS rules weren't being applied, and just moving them into a CSS file was enough to fix that. I know the priority of rules in `DEFAULT_CSS` is funny...)
- Update a few snapshots that have trivial differences... whatever this is:
image
It looks like some characters are being positioned ever so slightly differently within their cell, but whatever's happening there appears to be specific to the conversion to SVG, and not something that's visible through a real terminal.

Even with the SVG, outside of the "show differences" mode I can't see the difference between the old:
image
and the new:
image


godlygeek:
Oof, there's more that needs changing here it seems.

godlygeek:
... Huh. The 3 snapshots I needed to regenerate to get them to pass locally passed on Alpine, Ubuntu 22.04, and macOS, but failed in the manylinux and musllinux containers. What the heck? ☹️

godlygeek:
Ah - those 3 are testing with CPython 3.7, which Textual dropped support for in 0.44...

OK, I think we need to skip these 3 snapshot tests for Python 3.7, then.

darrenburns:
Hey 👋 

Textual snapshot tests will fail if there is a mismatch in the escape codes Textual writes to the terminal. A failing snapshot test does not always mean there _is_ a visual difference, it's more of a nudge to double check things, because different escape sequences have been produced.

In the case of the TextArea, we modified how it renders, which will result in some failing tests.

As for the artefacts you see in the diff, I think that's just down to how the browser is rendering the SVG. We actually overlay the two SVGs and the diff is done via CSS, so I think it's just a rendering anomaly at the browser level 😄 

For the `TextArea` changes, if you want to retain the exact same behaviour as before you can simply substitute calls to the `TextArea` constructor with `TextArea.code_editor`, which is a convenience constructor we added which retains the old defaults (wrapping disabled, syntax highlighting enabled, etc.)

godlygeek:
> different escape sequences have been produced.
> 
> In the case of the TextArea, we modified how it renders, which will result in some failing tests.

Fair enough, that makes sense.

> As for the artefacts you see in the diff, I think that's just down to how the browser is rendering the SVG. We actually overlay the two SVGs and the diff is done via CSS, so I think it's just a rendering anomaly at the browser level 😄

Sounds plausible. I was thinking some characters might have been aligned a few pixels off in the SVG from where they were in a previous version. Either way, I agree that it doesn't seem like anything to worry about.

What I'm _much_ more curious about is why switching from `DEFAULT_CSS` to `CSS_PATH` fixed some tests that were failing! I know there's some subtle differences between how `DEFAULT_CSS` is applied versus how CSS in a file is applied. I know `DEFAULT_CSS` is scoped to the widget and its children by default while `CSS_PATH` isn't... But I don't think that accounts for the differences, since the widget with the `DEFAULT_CSS` was the `App`, and so every other widget is one of its descendants...

> For the `TextArea` changes, if you want to retain the exact same behaviour as before you can simply substitute calls to the `TextArea` constructor with `TextArea.code_editor`, which is a convenience constructor we added which retains the old defaults

That was actually a bit uglier to do, since I don't want to drop support for Textual < 0.48 just yet: we'd need to special case on the existence or non-existence of that attribute. I started off writing `getattr(TextArea, "code_editor", TextArea)(...)` but that seemed arcane, and it seemed clearer just to set `soft_wrap = False` directly (knowing that attribute would just be ignored in old versions).

> the old defaults (wrapping disabled, syntax highlighting enabled, etc.)

I _didn't_ notice that there were more differences between the new defaults and the ones set by `code_editor` than just soft wrapping, but it turns out we're getting them all right anyway. We're already setting a theme and disabling line numbers and disabling focusing entirely (via `can_focus = False`), and that seems to be the sum total of the differences.

gusmonod (approved):
LGTM

PR #544, opened 2024-02-08
Fix tree reporter line highlighting

godlygeek:
Update tree reporter snapshots for Textual 0.49 (the TextArea no longer shows a cursor - yay!) and fix two bugs I noticed while looking at the snapshot diff: sometimes we don't show any lines, and sometimes we highlight the wrong line! Whoops.


godlygeek:
My test program for making sure the line number handling is now right was:
```
import mmap


def a():
    b()
    b()


def b():
    c()


def c():
    e()


def e():
    f()


def f():
    g()


def g():
    h()


def h():
    i()


def i():
    j()


def j():
    mmap.mmap(-1, 1024 * 1024)


a()
```

That made it easy to test both the case where there's not enough lines above the active frame's line to fill out our context and the case where there are (and the case where the active frame's line is close to the end of the file rather than the start, too, but that doesn't need any special handling).

pablogsal (approved):

PR #545, opened 2024-02-08
Stop using `IF` in our Cython code

godlygeek:
This has been deprecated for a while. Let's switch to using runtime checks for things that can be checked at runtime, and inline C++ code with conditional compilation for things that need to be checked at compile time.

### DIFF
--- a/src/memray/_memray_test_utils.pyx
+++ b/src/memray/_memray_test_utils.pyx
@@ -46,17 +46,29 @@ from libcpp.vector cimport vector
 
 from ._destination import Destination
 
-IF UNAME_SYSNAME == "Linux":
-    cdef extern from "sys/prctl.h":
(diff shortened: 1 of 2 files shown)


godlygeek:
This'll need to be rebased after #546 lands.

pablogsal (approved):

src/memray/_memray_test_utils.pyx
```diff
@@ -46,17 +46,29 @@ from libcpp.vector cimport vector
 
 from ._destination import Destination
 
-IF UNAME_SYSNAME == "Linux":
-    cdef extern from "sys/prctl.h":
-        int prctl(int, char*, char*, char*, char*)
+
+cdef extern from *:
```
pablogsal, line 50: Oh, nice approach ❤️

PR #549, opened 2024-02-13
Try to respect RPATHS of calling dlopen modules with dlinfo

pablogsal:
This commit is just for backup on our main strategy of using dlinfo
instead of dlopen.

**Describe your changes**

**Testing performed**

**Additional context**

### DIFF
--- a/src/memray/_memray/hooks.cpp
+++ b/src/memray/_memray/hooks.cpp
@@ -306,75 +306,84 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
     return ret;
 }
 
-// We need to override dlopen/dlclose to account for new shared libraries being
-// loaded in the process memory space. This is needed so we can correctly track
-// allocations in those libraries by overriding their PLT entries and also so we
-// can properly map the addresses of the symbols in those libraries when we
-// resolve later native traces. Unfortunately, we can't just override dlopen
-// directly because of the following edge case: when a shared library dlopen's
-// another by name (e.g. dlopen("libfoo.so")), the dlopen call will honor the
-// RPATH/RUNPATH of the calling library if it's set. Some libraries set an
-// RPATH/RUNPATH based on $ORIGIN (the path of the calling library) to load
-// dependencies from a relative directory based on the location of the calling
-// library. This means that if we override dlopen, we'll end up loading the
-// library from the wrong path or more likely, not loading it at all because the
-// dynamic loader will think the memray extenion it's the calling library and
-// the RPATH of the real calling library will not be honoured.
-//
-// To work around this, we override dlsym instead and override the symbols in
-// the loaded libraries only the first time we have seen a handle passed to
-// dlsym. This works because for a symbol from a given dlopen-ed library to
-// appear in a call stack, *something* from that library has to be dlsym-ed
-// first. The only exception to this are static initializers, but we cannot
-// track those anyway by overriding dlopen as they run within the dlopen call
-// itself.
-// There's another set of cases we would miss: if library A has a static initializer
-// that passes a pointer to one of its functions to library B, and library B stores
-// that function pointer, then we could see calls into library A via the function pointer
-// held by library B, even though dlsym was never called on library A. This should be
-// very rare and will be corrected the next time library B calls dlsym so this should
-// not be a problem in practice.
-
-class DlsymCache
-{
-  public:
-    auto insert(const void* handle)
-    {
-        std::unique_lock lock(mutex_);
-        return d_handles.insert(handle);
-    }
-
-    void erase(const void* handle)
-    {
-        std::unique_lock lock(mutex_);
-        d_handles.erase(handle);
-    }
-
-  private:
-    mutable std::mutex mutex_;
-    std::unordered_set<const void*> d_handles;
-};
-
-static DlsymCache dlsym_cache;
-
 void*
-dlsym(void* handle, const char* symbol) noexcept
+dlopen(const char* filename, int flag) noexcept
 {
-    assert(MEMRAY_ORIG(dlsym));
-    void* ret;
+    assert(MEMRAY_ORIG(dlopen));
+    void* ret = nullptr;
     {
         tracking_api::RecursionGuard guard;
-        ret = MEMRAY_ORIG(dlsym)(handle, symbol);
+#if defined(__GLIBC__)
+        // In GLIBC, dlopen() will respect the RPATH/RUNPATH of the caller when searching for the
+        // library, which won't work if we intercept dlopen() as we will be the caller. This means that
+        // callers that rely on RUNPATH to find their dependencies will fail to load. To work around
+        // this, we need to manually find our caller and walk the linker search path to know what we need
+        // to dlopen().
+        if (filename != nullptr && filename[0] != '\0' && std::strchr(filename, '/') == nullptr) {
+            void* const callerAddr = __builtin_extract_return_addr(__builtin_return_address(0));
+
+            Dl_info info;
+            if (dladdr(callerAddr, &info)) {
+                const char* dlname = info.dli_fname;
+                {
+                    // Check if we are being called from the main executable
+                    Dl_info main_info;
+                    void* main_sym = NULL;
+                    void* self_handle = MEMRAY_ORIG(dlopen)(nullptr, RTLD_LAZY | RTLD_NOLOAD);
+                    if (self_handle) {
+                        main_sym = dlsym(self_handle, "main");
+                        MEMRAY_ORIG(dlclose)(self_handle);
+                    }
+                    if (main_sym && dladdr(main_sym, &main_info)
+                        && strcmp(main_info.dli_fname, info.dli_fname) == 0)
+                    {
+                        dlname = nullptr;
+                    }
+                }
+
+                void* caller = MEMRAY_ORIG(dlopen)(dlname, RTLD_LAZY | RTLD_NOLOAD);
+                if (caller != nullptr) {
+                    Dl_serinfo size;
+                    if (dlinfo(caller, RTLD_DI_SERINFOSIZE, &size) == 0) {
+                        std::vector<char> paths_buf;
+                        paths_buf.resize(size.dls_size);
+                        auto paths = reinterpret_cast<Dl_serinfo*>(paths_buf.data());
+                        *paths = size;
+                        if (dlinfo(caller, RTLD_DI_SERINFO, paths) == 0) {
+                            for (unsigned int i = 0; i != paths->dls_cnt; ++i) {
+                                const char* name = paths->dls_serpath[i].dls_name;
+                                if (name == nullptr || name[0] == '\0') {
+                                    continue;
+                                }
+                                std::string dir = name;
+                                if (dir.back() != '/') {
+                                    dir += '/';
+                                }
+
+                                dir += filename;
+                                ret = MEMRAY_ORIG(dlopen)(dir.c_str(), flag);
+                                if (ret) {
+                                    break;
+                                }
+                            }
+                        }
+                    }
+                    MEMRAY_ORIG(dlclose)(caller);
+                }
+            }
+        }
+#endif
+        // Fallback if we found nothing
+        if (ret == nullptr) {
+            ret = MEMRAY_ORIG(dlopen)(filename, flag);
+        }
     }
     if (ret) {
-        auto [_, inserted] = dlsym_cache.insert(handle);
-        if (inserted) {
-            tracking_api::Tracker::invalidate_module_cache();
-            if (symbol
-                && (0 == strcmp(symbol, "PyInit_greenlet") || 0 == strcmp(symbol, "PyInit__greenlet")))
-            {
-                tracking_api::Tracker::beginTrackingGreenlets();
-            }
+        tracking_api::Tracker::invalidate_module_cache();
+        if (filename
+            && (nullptr != strstr(filename, "/_greenlet.") || nullptr != strstr(filename, "/greenlet.")))
+        {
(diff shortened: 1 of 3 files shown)


src/memray/_memray/hooks.cpp
```diff
@@ -306,75 +306,71 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
+                    Dl_serinfo size;
+                    if (dlinfo(caller, RTLD_DI_SERINFOSIZE, &size) == 0) {
+                        auto* paths = reinterpret_cast<Dl_serinfo*>(new char[size.dls_size]);
+                        *paths = size;
+                        if (dlinfo(caller, RTLD_DI_SERINFO, paths) == 0) {
+                            for (unsigned i = 0; i != paths->dls_cnt; ++i) {
+                                const char* name = paths->dls_serpath[i].dls_name;
+                                if (name == nullptr || name[0] == '\0') {
+                                    continue;
+                                }
```
godlygeek, line 344: I don't think this is correct - an empty entry is relative to $PWD, and skipping over that case but continuing to search other directories after it might mean we load the wrong file.

src/memray/_memray/hooks.cpp
```diff
@@ -306,75 +306,71 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
+            if (dladdr(callerAddr, &info)) {
+                const char* dlname = info.dli_fname;
+                {
+                    struct stat dlstat, exestat;
+                    // Check fi we are being called from the main executable
+                    if (stat(dlname, &dlstat) == 0 && stat("/proc/self/exe", &exestat) == 0
+                        && dlstat.st_dev == exestat.st_dev && dlstat.st_ino == exestat.st_ino)
+                    {
+                        dlname = nullptr;
+                    }
```
godlygeek, line 330: This is potentially error prone - it'll miss the case where the executable has been unlinked, at least (`/proc/self/exe` is a broken symlink in that case).

What if we instead checked if `dladdr(dlsym(dlopen(nullptr), "main"))` has the same `dli_fname` as `dladdr(callerAddr)`?

godlygeek:
We've decided to keep this option in our back pocket, but for now we're going to move forward with interposing dlsym instead of dlopen.

src/memray/_memray/hooks.cpp
```diff
@@ -306,75 +306,82 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
+dlopen(const char* filename, int flag) noexcept
 {
-    assert(MEMRAY_ORIG(dlsym));
-    void* ret;
+    assert(MEMRAY_ORIG(dlopen));
+    void* ret = nullptr;
     {
         tracking_api::RecursionGuard guard;
-        ret = MEMRAY_ORIG(dlsym)(handle, symbol);
+#if defined(__GLIBC__)
```
godlygeek, line 316: So we're doing this only for glibc, not for musl libc? musl behaves differently than glibc for this case?

src/memray/_memray/hooks.cpp
```diff
@@ -306,75 +306,82 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
+
+            Dl_info info;
+            if (dladdr(callerAddr, &info)) {
+                const char* dlname = info.dli_fname;
+                {
+                    struct stat dlstat, exestat;
+                    // Check fi we are being called from the main executable
+                    Dl_info main_info;
+                    void* main_sym = NULL;
+                    void* self_handle = dlopen(nullptr, RTLD_LAZY | RTLD_NOLOAD);
```
godlygeek, line 328:
```suggestion
                    void* self_handle = MEMRAY_ORIG(dlopen)(nullptr, RTLD_LAZY | RTLD_NOLOAD);
```

src/memray/_memray/hooks.cpp
```diff
@@ -306,75 +306,82 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
+                    if (self_handle) {
+                        main_sym = dlsym(self_handle, "main");
+                        dlclose(self_handle);
```
godlygeek, line 331:
```suggestion
                        MEMRAY_ORIG(dlclose)(self_handle);
```

src/memray/_memray/hooks.cpp
```diff
@@ -306,75 +306,82 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
+                                path += filename;
+                                ret = MEMRAY_ORIG(dlopen)(path.c_str(), flag);
+                                if (ret) {
+                                    break;
+                                }
+                            }
+                        }
+                        delete[] reinterpret_cast<char*>(paths);
+                    }
+                    dlclose(caller);
```
godlygeek, line 369:
```suggestion
                    MEMRAY_ORIG(dlclose)(caller);
```

src/memray/_memray/hooks.cpp
```diff
@@ -306,75 +306,82 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
+                    {
+                        dlname = nullptr;
+                    }
+                }
+
+                void* caller = MEMRAY_ORIG(dlopen)(dlname, RTLD_LAZY | RTLD_NOLOAD);
+                if (caller != nullptr) {
+                    Dl_serinfo size;
+                    if (dlinfo(caller, RTLD_DI_SERINFOSIZE, &size) == 0) {
+                        auto* paths = reinterpret_cast<Dl_serinfo*>(new char[size.dls_size]);
```
godlygeek, line 344: The `reinterpret_cast` on the `delete[]` and the lack of RAII makes me nervous... Let's instead do this:
```suggestion
                        std::vector<char> paths_buf;
                        paths_buf.resize(size.dls_size);
                        auto paths = reinterpret_cast<Dl_serinfo*>(&paths_buf[0]);
```

src/memray/_memray/hooks.cpp
```diff
@@ -306,75 +306,82 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
+                                    path += '/';
+                                }
+                                path += filename;
+                                ret = MEMRAY_ORIG(dlopen)(path.c_str(), flag);
+                                if (ret) {
+                                    break;
+                                }
+                            }
+                        }
+                        delete[] reinterpret_cast<char*>(paths);
```
godlygeek, line 367: If we use a `vector` as suggested above, this goes away:
```suggestion
```

src/memray/_memray/hooks.cpp
```diff
@@ -306,75 +306,82 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
+                }
+
+                void* caller = MEMRAY_ORIG(dlopen)(dlname, RTLD_LAZY | RTLD_NOLOAD);
+                if (caller != nullptr) {
+                    Dl_serinfo size;
+                    if (dlinfo(caller, RTLD_DI_SERINFOSIZE, &size) == 0) {
+                        auto* paths = reinterpret_cast<Dl_serinfo*>(new char[size.dls_size]);
+                        *paths = size;
+                        if (dlinfo(caller, RTLD_DI_SERINFO, paths) == 0) {
+                            for (unsigned i = 0; i != paths->dls_cnt; ++i) {
```
godlygeek, line 347: Minor stylistic nit, but I prefer
```suggestion
                            for (unsigned int i = 0; i != paths->dls_cnt; ++i) {
```

src/memray/_memray/hooks.cpp
```diff
@@ -306,75 +306,82 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
+                                if (name == nullptr || name[0] == '\0') {
+                                    // In the dynamic linking search path, an
+                                    // empty entry typically represents the
+                                    // current working directory ($PWD).
+                                    path = filename;
+                                }
+                                path = name;
+                                if (path.back() != '/') {
+                                    path += '/';
+                                }
```
godlygeek, line 359: This isn't right. You're conditionally assigning to `path`, and then unconditionally overwriting it with a different value. I'm thinking this should be:
```suggestion
                                    path = "./";
                                } else {
                                    path = name;
                                    if (path.back() != '/') {
                                        path += '/';
                                    }
                                }

```

godlygeek (changes requested):

src/memray/_memray/hooks.cpp
godlygeek: Ah, and moreover, musl doesn't support `RTLD_DI_SERINFOSIZE`. OK, fair enough.

godlygeek:
One downside I just noticed for this approach is that we're screwing up the contract for `dlerror()`: we're making a bunch of `dlopen` and `dlsym` and `dlclose` calls that the application doesn't know about, which could result in the return value from `dlerror` being clobbered by one of our internal attempts, rather than the last attempt the user made.

I'm not sure that's worth trying to work around, but it's worth noting, at least...

src/memray/_memray/hooks.cpp
godlygeek: As a minor nit, I'd suggest that renaming the `name` variable to `dir` would help make this a little more readable, too.

src/memray/_memray/hooks.cpp
```diff
@@ -306,75 +306,82 @@ posix_memalign(void** memptr, size_t alignment, size_t size) noexcept
...
+    void* ret = nullptr;
     {
         tracking_api::RecursionGuard guard;
-        ret = MEMRAY_ORIG(dlsym)(handle, symbol);
+#if defined(__GLIBC__)
+        if (filename != nullptr && filename[0] != '\0' && std::strchr(filename, '/') == nullptr) {
+            void* const callerAddr = __builtin_extract_return_addr(__builtin_return_address(0));
+
+            Dl_info info;
+            if (dladdr(callerAddr, &info)) {
```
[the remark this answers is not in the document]
pablogsal: IIRC `Python.h` doesn't expose it on purpose, so we may need to do it ourselves to be safe (not sure why is not blowing up currently).

pablogsal:
> One downside I just noticed for this approach is that we're screwing up the contract for `dlerror()`: we're making a bunch of `dlopen` and `dlsym` and `dlclose` calls that the application doesn't know about, which could result in the return value from `dlerror` being clobbered by one of our internal attempts, rather than the last attempt the user made.

We can clear the possible error before calling the real `dlopen` if you want by calling `dlerror()` ourselves as this is documented to clean existing errors when called: https://github.com/bminor/glibc/blob/96d0bf98cafd0b63721f369ca21ec64590551d47/dlfcn/dlerror.c#L120

src/memray/_memray/hooks.cpp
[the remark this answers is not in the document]
pablogsal: Ah, it's already defined:

```
  g++ -Wsign-compare -DNDEBUG -g -fwrapv -O2 -Wall -g -fstack-protector-strong -Wformat -Werror=format-security -g -fwrapv -O2 -fPIC -D_FILE_OFFSET_BITS=64 -DUSE_MEMRAY_TLS_MODEL=1 -Isrc/memray/_memray -Isrc -I/src/src/vendor/libbacktrace/install/include -I/usr/include/aarch64-linux-gnu -I/venv/include -I/usr/include/python3.11 -c src/memray/_memray.cpp -o /tmp/tmpuj28rc0q.build-temp/src/memray/_memray.o -std=c++17 -Wall -Og -fvisibility=hidden
  In file included from src/memray/_memray/records.h:15,
                   from src/memray/_memray.cpp:1337:
  src/memray/_memray/hooks.h:13: warning: "_GNU_SOURCE" redefined
     13 | #    define _GNU_SOURCE
        |
  <command-line>: note: this is the location of the previous definition
```

well, I'm going to ifdef around it to be explicit.

pablogsal:
@godlygeek Btw, I made some experiments regarding the null values of `paths->dls_serpath[i].dls_name`. Looks like empty entries are already resolved to ".":

```
LD_LIBRARY_PATH=":a:b" python -m pytest tests/integration/test_extensions.py -v -s -k test_dlopen_with_rpath
...
...
Num paths 8
Checking .
Checking a
Checking b
Checking /tmp/pytest-of-root/pytest-23/test_dlopen_with_rpath0/sharedlibs/sharedlibs
Checking /lib/aarch64-linux-gnu
Checking /usr/lib/aarch64-linux-gnu
Checking /lib
Checking /usr/lib
```

godlygeek:
> maybe we should bail out like I did before on `nullptr`.

I don't think it's possible to get `paths->dls_serpath[i].dls_name == nullptr`, so I don't think it matters too much what we do, beyond that we shouldn't crash if it somehow happens. I'd suggest we just `continue` and ignore that entry, but treating it the same as `""` seems fine too.

Checking the glibc code, it looks like it never assigns a null pointer to `dls_name`, and the docs don't suggest that it ever could. In fact (in retrospect unsurprisingly, considering there's no call to free this structure) every `dls_name` is assigned to some pointer into the `paths` buffer. That is, `RTLD_DI_SERINFOSIZE` sets `dls_size` to a size that's big enough not only for an array of `dls_cnt` entries of type `dls_serpath`, _but also_ for `dls_cnt` paths following the `dls_serpath` array, all concatenated together with `'\0'` bytes between them, and `paths->dls_serpath[i].dls_name` is set to the i'th string in that buffer.

pablogsal:
> I don't think it's possible to get `paths->dls_serpath[i].dls_name == nullptr`, so I don't think it matters too much what we do, beyond that we shouldn't crash if it somehow happens. I'd suggest we just `continue` and ignore that entry, but treating it the same as `""` seems fine too.

I pushed a fixup for that

godlygeek (approved):
LGTM. I've squashed the fixup and slightly reworded the news entry.

PR #560, opened 2024-03-06
Expose the main thread ID in the reader

pablogsal:
To be able to filter allocations that happened only on the thread that
started the tracking, expose the main thread ID in the reader class.

**Describe your changes**

**Testing performed**

**Additional context**

### DIFF
--- a/src/memray/_metadata.py
+++ b/src/memray/_metadata.py
@@ -13,6 +13,7 @@ class Metadata:
     peak_memory: int
     command_line: str
     pid: int
+    main_thread_id: int
     python_allocator: str
     has_native_traces: bool
     trace_python_allocators: bool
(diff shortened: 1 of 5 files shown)


src/memray/_metadata.py
```diff
@@ -11,6 +11,7 @@ class Metadata:
     peak_memory: int
     command_line: str
     pid: int
+    main_thread_tid: int
```
godlygeek, line 14: This feels a bit nitpicky, but - I think we should instead call this field
```suggestion
    main_thread_id: int
```

Now that it's going into the metadata it becomes a part of our public interface, and our JSON reports. I worry that using "tid" will mislead people, since it's not a tid in the sense of `gettid()`, or even in the sense of `pthread_self()` - it's a monotonically increasing integer which we've assigned.

Think there's anything to that concern, @pablogsal?

pablogsal: I am happy to make this change, will do it tomorrow!

godlygeek (approved):
LGTM

Issue #561, opened 2024-03-07
Include thread name in Memray live tracking view 


gibsondan:
### Is there an existing proposal for this?

- [X] I have searched the existing proposals

### Is your feature request related to a problem?

memray attach is incredibly useful - but for multi-threaded applications, once you've determined that a particular thread is the source of a memory issue, figuring out which thread in the application corresponds to that "Thread X of Y" number displayed in the live view is challenging.

### Describe the solution you'd like

Each thread has a string name - it would be very useful if the thread name was included alongside the "Thread X of Y" output in the live view. 

### Alternatives you considered

We can use py-spy to map thread numbers to thread names (I think?) but it would be much more convenient to have that information already available in memray

Thanks for building an amazing tool!

godlygeek:
> Each thread has a string name

Can you clarify which string name you're referring to, @gibsondan ?

gibsondan:
@godlygeek when you create a thread in python, you can supply a name: https://docs.python.org/3/library/threading.html#thread-objects

(Or a threadpoolexecutor can specify a prefix which is applied to all threads created in that pool)

I don't know offhand how that is translated into the operating system internals but I suspect it's through something like this: https://man7.org/linux/man-pages/man3/pthread_setname_np.3.html

py-spy, for example, includes these names in its output via commands like `py-spy dump` or the flamegraph produced by `py-spy record`:

```
Thread 41 (idle): "schedule_daemon_worker_7"
```

godlygeek:
We _do_ have access to the name set through `pthread_setname_np`, but we _don't_ have access to the name set using `threading.Thread.name`, and setting the name on a Python `Thread` object doesn't automatically cause `pthread_setname_np` to be called with the same name (or even a leading prefix of it).

That said, it would be interesting to try to collect the `name` attribute set on a `threading.Thread`... I wonder if we're able to...

PR #562, opened 2024-03-07
Print the thread name in the live TUI

pablogsal:
To allow users to determine thread in a multithreaded application
corresponds to the "Thread X of Y" number displayed in the live view,
include the thread name that we collect in the TUI.

Closes: #561


godlygeek (approved):
As we discussed offline, I made a _bunch_ of changes.

I renamed `AllocationRecord.thread_name` to `pretty_thread_name` to make it clear that it's doing some string formatting and not just returning data from the record, and then added a new `thread_name` attribute that _is_ just the data from the record.

Given that `AllocationRecord` is intended to (eventually) be part of our public interface, it probably shouldn't have `pretty_thread_name` at all, and we should instead move that to a function in `common.py` that our reporters can use, but that's a problem for another day.

Anyway, on top of that change, I updated this to only show a thread name if we have one, so the changes to the Textual snapshot cases are reverted (none of those threads have names). I also updated one of the tests to exercise the case where threads do have names, and I tested it against a Bloomberg service where threads do have names to show that it does work as intended.

I'm approving this since you won't be able to approve your own PR, but please do check my work before you land this, @pablogsal.

pablogsal:
Not really a fan of having `pretty_thread_name` and `thread_name`. Also, I think this PR makes the output nicer but it doesn't really solve that is requested in https://github.com/bloomberg/memray/issues/561 yet, so I would prefer to not merge this until we find a way to solve that as well.

godlygeek:
> Not really a fan of having `pretty_thread_name` and `thread_name`.

Me neither, but I don't think `pretty_thread_name` should be an attribute of the record at all. It's obscuring the actual thread name of the record by doing formatting, and presentation should be a concern of the reporter, rather than something exposed as an attribute of the record.

If you'd like me to revert the changes to have both `pretty_thread_name` and `thread_name`, I could, but the end result would be that the TUI has to parse the thread name out of the `record.thread_name` by doing something like
```py
thread_name = record.thread_name
thread_name = thread_name[thread_name.find("(") : thread_name.rfind(")") + 1]
```
which, yuck.

Alternatively, I could push forward with my cleanup by removing `pretty_thread_name` from the record and factoring it into a `def pretty_thread_name(record) -> str:` function in `reporters/__init__.py` or a new `reporters/common.py`, and update our reporters to get it from there.

> I think this PR makes the output nicer but it doesn't really solve that is requested in #561 yet

It doesn't, but it does display the thread name when we have the thread name, which is more than never. It also makes it so that the only piece missing for solving #561 is for us to capture the Python thread name in the tracker as well.

> I would prefer to not merge this until we find a way to solve that as well

Well, flipping this to draft until we decide, but it seems to me that this is already a reasonably sized self-contained PR that moves the ball forward and displays information that we already collect and which can be useful to people using the TUI.

I think #561 can really be broken down into 2 related work items:

1. capture Python thread names as well as OS thread names
2. displaying the best name we have for each thread on the TUI

We need both pieces, and I don't think both need to be in the same PR (or even the same release). This is already useful today in applications that use `pthread_setname_np` or `prctl(PR_SET_NAME)`, and some Python libraries do (`confluent_kafka` does, for instance).

godlygeek:
> Alternatively (and I think preferably), I could push forward with my cleanup by removing `pretty_thread_name` from the record and factoring it into a `def format_thread_name(record) -> str:` function in `reporters/__init__.py` or a new `reporters/common.py`, and update our reporters to get it from there.

I've done this (last commit on this PR)

> I think #561 can really be broken down into 2 related work items:
> 
> 1. capture Python thread names as well as OS thread names
> 2. displaying the best name we have for each thread on the TUI

I've implemented item 1 (second to last commit on this PR), using the descriptor-based plan we discussed offline earlier today. It turned out to be a bit more involved than it seemed at first glance, because we needed to worry about names being set before a thread is started, or one thread's name being set from a different thread. It turned out not to be too tough to support both of those things, though.

Before you ask: there's a C-style cast in this PR to convert a `pthread_t` to a `uint64_t`. I couldn't figure out any other way to do this conversion:

- On macOS, `pthread_t` is an opaque pointer, and we need to `reinterpret_cast` to get a `uint64_t`
- On i686, `pthread_t` is an `unsigned long int`, and we need to `static_cast` to get either a `uint64_t` or even a `uintptr_t`

I stopped wrestling with it after a while and just switched to a C-style cast, so that it reinterprets when we need reinterpreting and statically casts when we need a conversion between integer types.

src/memray/_memray/tracking_api.cpp
```diff
@@ -963,12 +965,37 @@ void
...
 }
 
+void
+Tracker::registerCachedThreadName()
+{
+    if (d_cached_thread_names.empty()) {
+        return;
+    }
+
+    auto it = d_cached_thread_names.find((uint64_t)(pthread_self()));
```
pablogsal, line 982:
```suggestion
    auto it = d_cached_thread_names.find(static_cast<uint64_t>(pthread_self()));
```

src/memray/_memray/tracking_api.cpp
```diff
@@ -963,12 +965,37 @@ void
...
+            deactivate();
+        }
+        d_cached_thread_names.erase(it);
+    }
+}
+
+void
+Tracker::dropCachedThreadName()
+{
+    d_cached_thread_names.erase((uint64_t)(pthread_self()));
```
pablogsal, line 996:
```suggestion
    d_cached_thread_names.erase(static_cast<uint64_t>(pthread_self()));
```

src/memray/_memray.pyx
```diff
@@ -690,6 +686,14 @@ cdef class Tracker:
             raise RuntimeError("Attempting to use stale output handle")
         writer = move(self._writer)
 
+        for attr in ("_name", "_ident"):
+            assert not hasattr(threading.Thread, attr)
+            setattr(
+                threading.Thread,
+                attr,
+                ThreadNameInterceptor(attr, NativeTracker.registerThreadNameById),
```
pablogsal, line 694: Ah, good old friend :P

src/memray/_memray/tracking_api.h
```diff
@@ -290,6 +290,27 @@ class Tracker
...
+    {
+        if (RecursionGuard::isActive || !Tracker::isActive()) {
+            return;
+        }
+        RecursionGuard guard;
+
+        std::unique_lock<std::mutex> lock(*s_mutex);
+        Tracker* tracker = getTracker();
+        if (tracker) {
+            if (thread == (uint64_t)(pthread_self())) {
```
pablogsal, line 303:
```suggestion
            if (thread == static_cast<uint64_t>(pthread_self())) {
```

Issue #559, opened 2024-03-05
Ctrl-Z in "memray tree" doesn't work


kshpytsya:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

It is impossible to use `Ctrl-Z` to background the "memray tree" TUI.

### Expected Behavior

I would like to be able to use the standard `Ctrl-Z` to send memray's TUI to background.

### Steps To Reproduce

1. Obtain some memray tracker run results
2. Run `memray tree <file>` in terminal
3. Press `Ctrl-Z`

Actual behavior: `Ctrl-Z` press is ignored.
Expected behavior: memray's TUI is sent to background and I see shell prompt.

### Memray Version

1.11.0

### Python Version

3.8, 3.11

### Operating System

Linux

willmcgugan:
ctrl+z to suspend isn't enabled by default, but you can [add it with a keybinding](https://textual.textualize.io/guide/app/#suspending-from-foreground). 

This is relatively new, so likely didn't exist when you built your TUI.

PR #581, opened 2024-04-17
Add a binding for Ctrl-Z to memray tree

pablogsal:
CTRL+z to suspend isn't enabled by default, but we can add it with a keybinding.

Closes: #559

### DIFF
--- a/src/memray/reporters/tree.py
+++ b/src/memray/reporters/tree.py
@@ -222,6 +222,7 @@ def node_is_not_import_system(node: Frame) -> bool:
 
 class TreeApp(App[None]):
     BINDINGS = [
+        Binding("ctrl+z", "suspend_process"),
         Binding(key="q", action="quit", description="Quit the app"),
         Binding(
             key="i", action="toggle_import_system", description="Hide import system"
--- a/news/581.feature.rst
+++ b/news/581.feature.rst
@@ -1,0 +1,1 @@
+Allow using Ctrl+Z to suspend ``memray tree`` and the live mode TUI.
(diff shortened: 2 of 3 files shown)


src/memray/reporters/tree.py
```diff
@@ -232,6 +232,7 @@ class TreeApp(App[None]):
         Binding(
             key="e", action="expand_linear_group", description="Expand linear group"
         ),
+        [Binding("ctrl+z", "suspend_process")],
```
godlygeek, line 235: Hm. _Technically_ this should use the `stty` `susp` binding, not hardcode `ctrl+z`... though to be fair I imagine almost no one changes it, so maybe it's not worth the bother.

setup.py
```diff
@@ -92,7 +92,7 @@ def build_js_files(self):
     "jinja2 >= 2.9",
     "typing_extensions; python_version < '3.8.0'",
     "rich >= 11.2.0",
-    "textual >= 0.41.0",
+    "textual >= 0.48.0",
```
godlygeek, line 95: We'd need to first drop support for Python 3.7, because modern Textual no longer supports 3.7.

src/memray/reporters/tree.py
godlygeek, line 235: Extra brackets here as well:
```suggestion
        Binding("ctrl+z", "suspend_process"),
```

setup.py
godlygeek: It looks like, if we don't require Textual 0.48 but make the binding unconditionally, things work "fine" even on old versions of Textual. Pressing ctrl-z at runtime just leads to a message logged to the Textual logs:
```
[14:11:50] INFO                                                                                            app.py:2794
<action> 'suspend_process' has no target. Could not find methods '_action_suspend_process' or 'action_suspend_process'
```

That seems acceptable to me in exchange for letting us continue to support 3.7 for a while longer.

godlygeek (approved):
LGTM

PR #582, opened 2024-04-20
Allow spinning a Memray environment up in codespaces

godlygeek:
Update our Dockerfile to the latest Debian stable, and add a devcontainer.json based on it, with our build and test dependencies installed.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,9 @@
+{
+  "name": "Memray Dockerfile",
+  "build": {
+    "context": "..",
+    "dockerfile": "../Dockerfile"
+  },
+  "runArgs": ["--cap-add=SYS_PTRACE", "--security-opt", "seccomp=unconfined"],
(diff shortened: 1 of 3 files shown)


jcarnaxide:
First comment I would like to add, spinning up the codespace took ~7 minutes. This is likely due to all the dependencies required for this currently considered codespace. This makes me wonder if a slimmed down version of this codespace might make sense for folks attempting to follow the tutorial (and not necessarily needing all of the dependencies)

I want to be clear, my suggestion is not to modify this codespace, it probably makes sense to leave this one as is, just some food for thought if we want to potentially shrink the time to create a codespace for folks. Then again, 7 minutes is not too long, so maybe optimization is overkill.

Just food for thought 😄

pablogsal:
Do we know where these 7 minute are being spent? The dependencies are not that many so I am a bit surprised

godlygeek:
@gusmonod Give this PR an approval, if you wouldn't mind!

gusmonod (approved):

PR #584, opened 2024-04-28
ci: Update to Ubuntu 24.04

godlygeek:


godlygeek:
Guess I'm jumping the gun here, GitHub Actions doesn't have Ubuntu 24 runners yet.

PR #589, opened 2024-05-05
feat: Merge threads in live view

mgmacias95:
*Issue number of the reported bug or feature request: https://github.com/bloomberg/memray/discussions/587*

**Describe your changes**
This PR adds a new option in the live view called "merge threads" that summarizes all allocations in a single view. It also hides the buttons ">" and "<" because you can't move between threads in that moment.

In addition, I added some missing docs to the readme.

**Testing performed**
Manual testing + unit tests included in this PR


README.md
```diff
@@ -148,6 +148,9 @@ positional arguments:
     parse               Debug a results file by parsing and printing each record in it
     summary             Generate a terminal-based summary report of the functions that allocate most memory
     stats               Generate high level stats of the memory usage in the terminal
+    transform           Generate reports files in different formats
+    attach              Begin tracking allocations in an already-started process
+    detach              End the tracking started by a previous ``memray attach`` call
```
godlygeek, line 153: Would you mind splitting this change out into a separate commit of its own, since it's for a separate purpose than the rest of this PR?

src/memray/reporters/tui.py
```diff
@@ -290,6 +290,7 @@ class AllocationTable(Widget):
     sort_column_id = reactive(default_sort_column_id)
     snapshot = reactive(_EMPTY_SNAPSHOT)
     current_thread = reactive(0)
+    merge_threads = False
```
godlygeek, line 293: Hm, I would think we'd want this to be a reactive attribute, since we need to repopulate the table whenever it changes.

godlygeek (changes requested):
This seems like a great start, but I'm spotting lots of little issues as I try to use it.

For one, if the process is still running and output isn't paused, then threads seem to become un-merged again as soon as the next update happens.

For another, this disables the `<` and `>` keys, but not the left and right arrow keys, which do the same thing (as long as the grid isn't wide enough to scroll - if it is, they scroll the grid instead).

Also, if after the tracked process dies you use the `>` key to select a specific thread and then enable thread merging, the table doesn't get redrawn to include allocations from all threads, even though the header gets redrawn to say that allocations from all threads are being shown together.

Can you spend some time tracking down and fixing these issues?

mgmacias95:
Hello @godlygeek,

I plan working on this by the end of the week. Thank you for your review! :)

mgmacias95:
Hey @godlygeek,

So when I said I would work on this by the end of the week, I meant by the end of the day 😅 😂  

I fixed the issues you pointed out. Can you please review it again? 

Thanks!

godlygeek:
It looks like, while it's in "merge threads" mode, the header winds up incorrectly showing a thread count again the first time a new thread is created after entering "merge threads" mode. See this screenshot, for instance:
image
You'll see that it shows "Thread 1 of 23" at the top, but "**M** Unmerge threads" at the bottom, so it's still in merge threads mode, but incorrectly showing a thread index and thread count.

Try using this test program:
```py
import mmap
import threading
import time


def allocate():
    with mmap.mmap(-1, 1024 * 1024):
        time.sleep(1)


threads = [threading.Thread(target=allocate) for _ in range(3)]
for thread in threads:
    time.sleep(3)
    thread.start()
```
and pressing "m" to merge threads as soon as the TUI shows up. You'll see that the heading switches to saying "Displaying all threads." but as soon as a new thread shows up, it switches back to displaying a thread count.

mgmacias95:
Hello @godlygeek,

Thank you for your review again, I just pushed a commit fixing that problem. Can you take another look? Thanks!

Cheers!

mgmacias95:
Hello @godlygeek,

I have fixed some conflicts that appeared in  2f50fa7b0ec605b1fa99ef69515d4db2de34bda6. Can you take a look again please? 🥺 

Thanks!

godlygeek (approved):
I picked this up and tweaked it a small bit, but this was great work. (Even more so because you were trying to keep up to date with some of the hacks being applied in other PRs 😅)

Thanks so much for the contribution, @mgmacias95! 🎉

PR #597, opened 2024-05-19
Fix live updates to footers on Textual TUIs

godlygeek:
In Textual 0.61, `App.namespace_bindings` was removed in favor of `Screen.active_bindings`. Update our implementation of dynamically updating footer descriptions to work for both, since we still support Textual versions below 0.61.


gusmonod (approved):
LGTM, but as discussed I would prefer for `update_key_description` to be a shared function in a new file, e.g. `src/memray/reporters/_textual_hacks.py`

PR #602, opened 2024-05-20
Publish docs on every PR, not just releases

godlygeek:
I don't want to merge this PR, but letting CI run it will allow updated docs to be published.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,38 @@
+name: Docs
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  publish_docs:
+    name: Publish docs
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


godlygeek:
Oops, this doesn't work the way I hoped it would - we don't have the token to publish the docs from a PR, so this would need to be merged in order to work...

.github/workflows/docs.yml
```diff
@@ -0,0 +1,38 @@
...
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  publish_docs:
+    name: Publish docs
+    runs-on: ubuntu-latest
+    #if: github.event_name == 'release' && github.event.action == 'published'
```
gusmonod, line 12: I'm not sure whether we would want this to be like this forever, but at least while sprinting in PyCon this is super useful.

gusmonod (approved):
Thanks!

Issue #557, opened 2024-03-03
Move the runner into a separate package with minimal dependencies


kshpytsya:
### Is there an existing proposal for this?

- [X] I have searched the existing proposals

### Is your feature request related to a problem?

The system I am currently working on involves some confidential customer data to which regular developers do not have access. Due to the complex nature of ML algorithms and libraries involved, sometimes we encounter anomalous memory usage which we cannot easily reproduce with our test data set. I am considering integration of some memory profiling mechanism, possibly involving memray which would be selectively enabled in production environment to help us to get insights into problematic cases without requiring access to customer data.

For multiple reasons (security, reliability, docker image size to name a few) we try to minimize the number of our dependencies. Memray has a number of dependencies which are unnecessary for trace collection:

https://github.com/bloomberg/memray/blob/578e02d20bc63bddc0fc25c28bf7569519b89838/setup.py#L91-L96

and those have their own dependencies as well.

### Describe the solution you'd like

It would be great to split memray into two packages, tentatively named `memray-core` and `memray`. The former would only be able to perform collection, and the later would do everything memray does today.

### Alternatives you considered

Another possibility would be having all the dependencies not related to `memray.Tracker` to be under "extras" but that would probably be a nuisance for most ordinary users who would want a simple `pip install memray` to install everything.

godlygeek:
Hm. It's true that Jinja2, Rich, and Textual aren't needed for collecting profile data, but they are needed for generating reports from the profile data, and as a general rule, we expect the reporters to be run on the same system as the report was generated on... In the case of `--native` mode reports, it can be extremely difficult to reproduce an environment on a different machine where the instruction pointers and shared library names stored in the capture file are able to be interpreted correctly. Because of that, the idea of a `memray-core` package with no support for interpreting the capture file seems problematic. It encourages a workflow that's likely to cause at least a lot of friction, and that may prevent certain types of reports from being generated from the captured file at all.

kshpytsya:
I do realize it may need extra refactoring but I do believe it would be possible to implement the collection phase (including the extraction the required information from the running environment, e.g. from shared libraries) without those libraries, and make the resulting files completely portable.
Of course, the code is yours, but to me this would seem a better architecture.

To clarify, I envision roughly the following flow:

1. `with memray_core.Tracker: ...` exactly what it does today. `memray.Tracker` would be just a re-export of `memray_core.Tracker`.
[...]

PR #606, opened 2024-05-20
import commands only when dependencies are available

micheleAlberto:
*Issue number of the reported bug or feature request: #557*

**Describe your changes**
This PR allows the `memray` cli to run with a reduced set of dependencies. The cli will only list commands that are supported by installed packages and ignore the commands that require additional packages.

**Testing performed**
**without textual** 
```
pip uninstall textual
memray --help
```
results in 
```
usage: memray [-h] [-v] [-V] {flamegraph,parse,stats,table,transform} ...

Memory profiler for Python applications

Run `memray run` to generate a memory profile report, then use a reporter command
such as `memray flamegraph` or `memray table` to convert the results into HTML.

Example:

    $ python3 -m memray flamegraph output.bin

positional arguments:
  {flamegraph,parse,stats,table,transform}
                        Mode of operation
    flamegraph          Generate an HTML flame graph for peak memory usage
    parse               Debug a results file by parsing and printing each record in it
    stats               Generate high level stats of the memory usage in the terminal
    table               Generate an HTML table with all records in the peak memory usage
    transform           Generate reports files in different formats

options:
  -h, --help            show this help message and exit
  -v, --verbose         Increase verbosity. Option is additive and can be specified up to 3 times
  -V, --version         Displays the current version of Memray

Please submit feedback, ideas, and bug reports by filing a new issue at
https://github.com/bloomberg/memray/issues
```
**without jinja2**
```
pip uninstall jinja2
memray --help
```

results in 
```
usage: memray [-h] [-v] [-V] {run,attach,live,parse,stats,summary,transform,tree} ...

Memory profiler for Python applications

Run `memray run` to generate a memory profile report, then use a reporter command
such as `memray flamegraph` or `memray table` to convert the results into HTML.

Example:

    $ python3 -m memray run -o output.bin my_script.py

positional arguments:
  {run,attach,live,parse,stats,summary,transform,tree}
                        Mode of operation
    run                 Run the specified application and track memory usage
    attach              Begin tracking allocations in an already-started process
    live                Remotely monitor allocations in a text-based interface
    parse               Debug a results file by parsing and printing each record in it
    stats               Generate high level stats of the memory usage in the terminal
    summary             Generate a terminal-based summary report of the functions that allocate most memory
    transform           Generate reports files in different formats
    tree                Generate a tree view in the terminal for peak memory usage

options:
  -h, --help            show this help message and exit
  -v, --verbose         Increase verbosity. Option is additive and can be specified up to 3 times
  -V, --version         Displays the current version of Memray

Please submit feedback, ideas, and bug reports by filing a new issue at
https://github.com/bloomberg/memray/issues
```
**without jinja2 and textual**
```
pip uninstall jinja2 textual
memray --help
```
results in 
```
usage: memray [-h] [-v] [-V] {parse,stats,transform} ...

Memory profiler for Python applications

Run `memray run` to generate a memory profile report, then use a reporter command
such as `memray flamegraph` or `memray table` to convert the results into HTML.

Example:

positional arguments:
  {parse,stats,transform}
                        Mode of operation
    parse               Debug a results file by parsing and printing each record in it
    stats               Generate high level stats of the memory usage in the terminal
    transform           Generate reports files in different formats

options:
  -h, --help            show this help message and exit
  -v, --verbose         Increase verbosity. Option is additive and can be specified up to 3 times
  -V, --version         Displays the current version of Memray

Please submit feedback, ideas, and bug reports by filing a new issue at
https://github.com/bloomberg/memray/issues
```
**Additional context**


godlygeek:
@micheleAlberto Would you mind updating your commit to include a `Signed-off-by`? See https://github.com/bloomberg/memray/blob/main/CONTRIBUTING.md#contribution-licensing

Since this is multiple commits, the easiest way to do that would be using a `git rebase -i` against the main branch and then doing `reword` for each commit to add the `Signed-off-by` line.

godlygeek:
Thank you for the work you put in here, @micheleAlberto, but after some reflection, I've decided against going ahead with this approach.

The risk that this approach introduces is that, because the set of reporters to be exposed in any given environment is determined implicitly by whether they can be successfully imported, we might introduce a new dependency on a library that's not guaranteed to be installed, and accidentally trigger the removal of a reporter that we meant to include in our default set. This case isn't as far fetched as it might seem, and some of these dependencies are quite subtle. For instance, `memray summary` can't be imported unless Textual is installed. This isn't because it uses Textual, but rather because it imports a single function from `memray.reporters.tui`, to avoid duplication, which causes `textual` to be imported. The fix for that is factoring the function that needs to be shared into a common module that both `memray.reporters.tui` and `memray.reporters.summary` can depend upon, but just ignoring every module that can't be imported would lead to us not identifying problems like that, and potentially cause regressions if similar problems are introduced in the future.

Going further, another problem with making detection of optional libraries implicit rather than explicit is that we don't have a good way to notify the user of the missing dependency. I think what we want is for `memray tree` to be present in the `--help` and runnable from the CLI, but to emit a message if it's run explaining that it needs Textual to work, and explaining how to go about installing that dependency. But, emitting that message is only doable if we know what package to tell the user to install: we need to know that `memray tree` depends on Textual but `memray flamegraph` depends on Jinja2 to emit the right error message for each.

I do appreciate the work that you put in here, and it's certainly helped me solidify my understanding of the problem and what a solution needs to look like, but I don't think it's a good starting point to get to the best solution.

Sorry!

Issue #600, opened 2024-05-20
Prevent the test suite from rendering Textual apps to the screen while running


godlygeek:
When running `make check` or `make pycoverage`, some of the tests in `test_main.py` cause Textual TUIs to be rendered to the screen.

The graphical output of these tests should be suppressed, so that only the pytest output appears, even when running with `-s`. (We may want to reconsider running with `-s` by default, too, but the current behavior is wrong and should be fixed, anyway.)

terapyon:
I tried two command. I looked below.

```
make chack
```
<img width="682" alt="screenshot-chack2024-05-20 12 26 53" src="https://github.com/bloomberg/memray/assets/485372/667c0f0f-57d7-4ca4-a29b-583cd4f8356e">


```
make pycoverage
```
<img width="682" alt="screenshot-pycoverage2024-05-20 12 29 19" src="https://github.com/bloomberg/memray/assets/485372/e4543156-a11f-46c0-b21f-0d0ef6836d35">



No need to show Memray GUI, We want to show only pytest text.

terapyon:
I used `pytest.mark.skip` to investigate the scope of impact. When I disabled all the methods in `class TestTreeSubCommand` and `TestLiveRemoteSubcommand`, the issue stopped appearing. It seems that these classes affect the `tree` and `remote`,  `live-remote` switches.

https://github.com/bloomberg/memray/blob/main/tests/integration/test_main.py

vitorespindola:
I wiil take a look on it

gusmonod:
Thank you @terapyon for the preliminary research! Vitore will now take a look

PR #608, opened 2024-05-20
Prevent test suite from renderering Textual UI

vitorespindola:
Send client stdout and stderr to PIPE so Textual UI will not be shown on tests output.

#600 

**Describe your changes**
Change tree and live subcommands client output on some tests to PIPE, to prevent Textual UI to be shown.

**Testing performed**
make check
pytest with -s flag was tested as well


gusmonod (approved):
Thanks @vitorespindola !

Issue #591, opened 2024-05-06
Flame graph time axis offset (perhaps due to timezone)


braingram:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I generated a profile and flamegraph on a remote machine (that uses UTC). The flamegraph page `stats` include:
```
Start time: 2024-05-03 12:35:24.695000
End time: 2024-05-06 02:40:31.130000
```
However, when I view the page on my laptop (ET) the first point on the graph reports a time of `May 3rd, 2024, 08:35`.

Is there an option to set maybe the timezone of the graph so it will match the times reported in the `stats` (and in my case times from other logs generated by the program, which are all in the system UTC time)?

Thanks and let me know if any more information is helpful. Also, memray is awesome :)

### Steps To Reproduce

I think this could be reproduced by generating a log on a system using UTC then viewing it on a system uses a different timezone.

### Memray Version

1.12.0

### Python Version

3.12

### Operating System

Linux

vitimm144:
I can try to help on this issue

PR #611, opened 2024-05-21
Fix timezone on stats modal

vitimm144:
*Issue number of the reported bug or feature request: #591*

**Describe your changes**
Used vanilla javascritpt to format the datetime on stats table


gusmonod (approved):
LGTM

godlygeek (approved):
This was only half the fix, but I've pushed a separate commit with the other half. In addition to having making the HTML reporters localize the date/time when the page is loaded, we also need to embed the original time zone in the report. Previously we were embedding a non-localized time stamp, which didn't contain enough information to properly localize it to a different time zone.

After adding that commit (and a news entry), this looks good to me!

Thank you for the contribution!

PR #615, opened 2024-05-28
Fix a segfault on macOS when _dyld_shared_cache_contains_path is not available

pablogsal:
The _dyld_shared_cache_contains_path is only available in macOS 11.0
or later. As we are compiling in a new system that's newer than 11.0 but
setting the deployment target to an older version, it's possible that
the _dyld_shared_cache_contains_path is not available in the target
platform and in that case we will segfault when calling it. It's also
possible that when compiling memray in an old system
_dyld_shared_cache_contains_path is not even prototyped in the headers.

To handle this, we change to fetch the function via a dlopen/dlsym approach
that will be generic in all macOS versions at the cost of an extra call
to dlopen the first time we interpose symbols.

### DIFF
--- a/src/memray/_memray/macho_shenanigans.cpp
+++ b/src/memray/_memray/macho_shenanigans.cpp
@@ -296,7 +296,20 @@ patch_symbols_in_shared_object(
             unsigned int section_type = section->flags & SECTION_TYPE;
             LOG(DEBUG) << "Considering section " << i << " (" << section->segname << ":"
                        << section->sectname << ")";
-            if (_dyld_shared_cache_contains_path(image_name) && strcmp(section->segname, SEG_TEXT) == 0)
+
+            // dlopen with null and check for _dyld_shared_cache_contains_path but do it only
+            // once with std::once_flag
+            static std::function<bool(const char*)> dyld_shared_cache_contains_path;
+            static std::once_flag _dyld_shared_cache_check_flag;
+            std::call_once(_dyld_shared_cache_check_flag, [&]() {
+                void* handle = dlopen(nullptr, RTLD_LAZY);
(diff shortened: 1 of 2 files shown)


src/memray/_memray/macho_shenanigans.cpp
```diff
@@ -296,7 +298,15 @@ patch_symbols_in_shared_object(
             unsigned int section_type = section->flags & SECTION_TYPE;
             LOG(DEBUG) << "Considering section " << i << " (" << section->segname << ":"
                        << section->sectname << ")";
-            if (_dyld_shared_cache_contains_path(image_name) && strcmp(section->segname, SEG_TEXT) == 0)
+                
+#if defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 110000
```
godlygeek, line 302: I don't think I understand what's happening here. I get that the `#include <AvailabilityMacros.h>` should cause `__MAC_OS_X_VERSION_MAX_ALLOWED` to be defined, but won't it always be defined as `101400` (macOS 10.14) on x86-64 since that's what we set as `MACOSX_DEPLOYMENT_TARGET`?

I'm not really following what this is supposed to be guarding against, actually. Is it that `__builtin_available` wasn't available in 10.14?

godlygeek: > I'm not really following what this is supposed to be guarding against, actually. Is it that `__builtin_available` wasn't available in 10.14?

If so, I think we could guard against that with `#if __has_builtin(__builtin_available)` instead.

godlygeek (approved):
LGTM

PR #641, opened 2024-07-01
Tutorial: Fixes for Exercise 3

statkute:
The PR fixes the issues with the exercise 3 from the tutorial:
- There was an inconsistency between the code in the solution vs the exercise, as the exercise used `@classmethod` and the solution did not. Now that's been made consistent
- The 2nd solution did not lower the memory usage. Not sure why it was added there - the PR removes it
- An edit to the for consistency to utilise the same global variables instead of magic numbers


godlygeek (approved):
LGTM

PR #643, opened 2024-07-02
Update tests to support Textual 0.71

godlygeek:
- Work around a Textual deadlock in our tests. There's a deadlock on shutdown when the pilot exits before the message pump has quiesced, but it doesn't occur when we wait for the message pump to become idle.
- Update our tests' Textual snapshots. The latest version of Textual uses a new SVG generation algorithm that results in some slight differences compared to the old algorithm's output, but should be more stable going forward. Additionally, the two trailing spaces removed by the previous commit show up as differences in these snapshots.
- Avoid extra trailing spaces on <ROOT> node in the tree reporter tree. If trailing spaces are included, they are highlighted with an orange background when the line is selected.


pablogsal (approved):

godlygeek:
Deadlock reported at https://github.com/Textualize/textual/issues/4699

PR #656, opened 2024-08-01
Retain inverted-ness when redrawing graph.

### LINKED ISSUE #655: Changing the flamegraph filter switches to icicles

- [X] I have searched the existing issues

### Current Behavior

Changing the filter on Memray's flamegraph always switches the view back to icicles, and leaves the selector on flames.

### Expected Behavior

It would be nice if it just kept it on flames, but if it was too complicated to do that for whatever reason, then it should at least switch the selector to icicles. 

### Steps To Reproduce

1. Open any Memray flamegraph output
2. Click on "flames"
3. Check or uncheck "hide import system frames"

### Memray Version

1.13.4

### Python Version

3.12

### Operating System

Linux

### Anything else?

I'll submit a PR for this :D

ZeroIntensity:
Closes #655.

**Describe your changes**
When the flamegraph is redrawn, keep the inversion from the previous graph, if it exists.

**Testing performed**
I opened a flamegraph and made sure that it worked :)


godlygeek (approved):
LGTM, thanks for the fix @ZeroIntensity!

PR #673, opened 2024-09-03
Use scikit-build 

pablogsal:
- **Try to respect RPATHS of calling dlopen modules with dlinfo**
- **Use scikit-build-core**

**Describe your changes**

**Testing performed**

**Additional context**


src/memray/_memray.pyx
```diff
@@ -69,12 +69,12 @@ from libcpp.unordered_map cimport unordered_map
...
-from ._destination import SocketDestination
-from ._metadata import Metadata
-from ._stats import Stats
-from ._thread_name_interceptor import ThreadNameInterceptor
+from memray._destination import Destination
+from memray._destination import FileDestination
+from memray._destination import SocketDestination
+from memray._metadata import Metadata
+from memray._stats import Stats
+from memray._thread_name_interceptor import ThreadNameInterceptor
```
godlygeek, line 77: Something about how `scikit-build-core` does editable installs is causing this to be necessary... Things work fine with a non-editable install without these changes, but not with an editable install.

godlygeek: Ah, presumably this:

> The default `editable.mode`, "redirect", uses a custom redirecting finder to combine the static CMake install dir with the original source code.

It's probably winding up wrong about what its package is due to that...

godlygeek: But somehow it's working for `memray._test_utils`, which _also_ does
```py
from ._destination import Destination
```
Hm, there's still something I don't understand here...

godlygeek: Ah, the thing that's throwing it off is that there's both a `memray/_memray/` directory and a `memray._memray` module, and they're not the same. I think this is a bug in `scikit-build-core`'s editable mode. I'll try to whip up an MRE to report the bug to them when I can. It should be easy to demonstrate that `__package__` changes depending on whether or not the install is editable.

godlygeek: That's been fixed in scikit-build-core 0.10.6

godlygeek:
This is missing the `npm install` and `npm run-script build` steps, I think

pablogsal:
I rebased  and fixes all the other issues

pablogsal:
@godlygeek this is ready can you approve?

src/memray/commands/_attach.gdb
```diff
@@ -19,7 +19,7 @@ set scheduler-locking on
 call (int)Py_AddPendingCall(&PyCallable_Check, (void*)0)
 
 p "Checking if we can access the library"
-set $result = (int)access($libpath, 5)
+set $result = (int)access($libpath, 4)
```
godlygeek, line 22: I don't understand this change. Is `scikit-build-core` sometimes creating shared libraries that don't have the executable bit set? If so, I can't reproduce that... What circumstances does it do that under?

pyproject.toml
```diff
@@ -1,19 +1,130 @@
...
+input = "src/memray/_version.py"
+
+[project.optional-dependencies]
+test = [
+  "Cython",
+  "greenlet; python_version < '3.15'",
+  "pytest",
+  "pytest-cov",
+  "ipython",
+  "setuptools",
```
godlygeek, line 42:
```suggestion
  "scikit-build-core",
```

pyproject.toml
```diff
@@ -1,19 +1,130 @@
...
 ]
+lint = ["black", "flake8", "isort", "mypy", "check-manifest"]
+benchmark = ["asv"]
+dev = [
+  "Cython",
+  "greenlet; python_version < '3.15'",
+  "pytest",
+  "pytest-cov",
+  "ipython",
+  "setuptools",
```
godlygeek, line 64:
```suggestion
  "scikit-build-core",
```

pyproject.toml
```diff
@@ -1,19 +1,130 @@
...
+    "/Jenkinsfile",
+    "/requirements-*.txt",
+    "/.medusarc",
+    "/valgrind.supp",
+]
+wheel.packages = ["src/memray"]
+wheel.exclude = ["memray/_memray/**"]
 
 [dependency-groups]
 test=['Cython', 'coverage[toml]', "greenlet; python_version < '3.15'", 'pytest', 'pytest-cov', 'ipython', 'setuptools', 'pkgconfig', 'pytest-textual-snapshot', 'textual >= 0.43, != 0.65.2, != 0.66', 'packaging']
```
godlygeek:
```suggestion
test=['Cython', 'coverage[toml]', "greenlet; python_version < '3.15'", 'pytest', 'pytest-cov', 'ipython', 'scikit-build-core', 'pkgconfig', 'pytest-textual-snapshot', 'textual >= 0.43, != 0.65.2, != 0.66', 'packaging']
```

godlygeek (changes requested):
Just a few small things:

The `setup.py` used to run `npm install` and `npm run-script build` if `src/memray/reporters/templates/assets/*.js` didn't exist. The CMakeLists.txt never does that. Is that an intentional omission? Is the assumption that, because these files are checked into source control, we should never encounter the situation where they're missing, and so what was in the `setup.py` was dead code?

And, on a separate note: if we're removing the `MANIFEST.in`, we should also remove the `setup.py`, shouldn't we? Is there a reason to keep it?

pablogsal:
Done. I removed `setup.py`, updated ASV, and verified all checked-in JS assets in the sdist/wheel and the regeneration stuff. I added `scikit-build-core` but kept `setuptools` because the integration fixtures still use `setup.py`/`distutils`.

src/memray/commands/_attach.gdb
pablogsal: You are right: this is not a `scikit-build-core` behavior. CMake, the wheel, and a clean UV 3.14 editable install all keep the execute bit. The rebased UV and coverage jobs failed at this guard, but their logs did not record the file mode. The actual issue is that `5` checks `R_OK | X_OK`, while the guard only needs readability; I verified `dlopen` succeeds on a `0644` library. This is an unrelated forward-port fix.

godlygeek:
> kept `setuptools` because the integration fixtures still use `setup.py`/`distutils`

Ah, right, makes sense

godlygeek (approved):
LGTM. I copied the big comment about our TLS model hackery over from the setup.py into the CMakeLists.txt, and squashed a few of these commits together. I left the ones that were fixing separate issues that were merely unearthed by the build backend change un-squashed.

src/memray/commands/_attach.gdb
godlygeek: Ah, I see what caused this. https://github.com/bloomberg/memray/pull/959 fixes it. Though I've left your change in place, since you're right, we only need `R_OK` and not `R_OK | X_OK`, but I've figured out why the `x` bit wasn't (always?) getting set, and it's an easy fix.

Issue #662, opened 2024-08-08
Potential deadlock when interposing symbols


sethtroisi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I am seeing deadlock during `dl_iterate_phdr` similar to https://github.com/bloomberg/memray/commit/6fbadf72b66da53f1ac56ce2f505f79d6e2dbb2a 

By adding a hack in the same place to avoid patching `Smemray_Slib_Umemray_Ucc_Ulib.so` everything works.

```
    if ( strlen(info->dlpi_name) < 1 || 
        strstr(info->dlpi_name, "memray_") || // <- new hack
        strstr(info->dlpi_name, "/ld-linux") ||
        strstr(info->dlpi_name, "/ld-musl") ||
        strstr(info->dlpi_name, "linux-vdso.so.1")) {
      // Avoid chaos by not overwriting the symbols in the linker.
      // TODO: Don't override the symbols in our shared library!
        LOG(WARNING) << "P1 " << info->dlpi_name;
      return 0;
    }
```
I'm not sure how to guess what the `dlpi_name` will be. But `memray_` alone is dangerous because it's appearing in the path portion of the dlpi_name. Maybe checking that it's in the filename after any slashes?

The stack trace when the tests time out is with the dl being either `libunwind.so` or `libcxxabi_Slibcxxabi_Ucc_Ulibrary.so`

```
*** SIGTERM received by PID 3156388 (TID 3156388) on cpu 3 from PID 3156373; stack trace: ***
PC: @     0x7f38e0830a9a  (unknown)  memray::intercept::free()
    @     0x7f38ab409df1       1904  FailureSignalHandler()
    @     0x7f38fdd36e80  2138701600  (unknown)
    @     0x7f3865e9b3eb         32  std::__u::ios_base::~ios_base()
    @     0x7f38e082a818         80  memray::LOG::~LOG()
    @     0x7f38e082b1f7        368  memray::linker::patch_symbols()
    @     0x7f38e082abdd        400  memray::linker::phdrs_callback()
    @     0x7f38fd1c3a7d        160  dl_iterate_phdr
    @     0x7f38e082a666        352  memray::linker::SymbolPatcher::overwrite_symbols()
    @     0x7f38e085a61e        112  memray::tracking_api::Tracker::Tracker()
    @     0x7f38e085c3c2         64  memray::tracking_api::Tracker::createTracker()
[...]

PR #685, opened 2024-09-18
Avoid patching symbols in the extension module

pablogsal:
The fact that patching ourselves had not raised problems so far its
really an outstanding fact in this universe. Unfortunately seems that
with the latest toolchain + GCC there is something that causes memray to
point the d_original entry of the hooks pointing to itself, which should
never happen.

To fix this resiliently, avoid patching ourselves by getting our own
name in the extension module and then avoiding that shared object.

Closes: #662

### DIFF
--- a/news/685.bugfix.rst
+++ b/news/685.bugfix.rst
@@ -1,0 +1,1 @@
+Fix some crashes caused by interposing symbols in memray itself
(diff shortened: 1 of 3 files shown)


godlygeek (approved):
LGTM

sethtroisi:
Thanks!

Out of curiosity did you also encounter this issue? if so do you have a guess which gcc versioned caused it?

pablogsal:
> Thanks!
> 
> Out of curiosity did you also encounter this issue? if so do you have a guess which gcc versioned caused it?

I encountered this only when compiling memray with gcc 14 and python compiled with -no-pie. It may be a linker bug because that fields should never end in the PLT with a relocation.

PR #687, opened 2024-09-25
tests: Update snapshots for Textual 0.81

godlygeek:
There are no interesting changes, just some false positives.


sarahmonod (approved):
LGTM

PR #693, opened 2024-10-21
Fix symbol resolution for malloc_type functions on macOS Sequoia

pablogsal:
This commit addresses an issue with symbol resolution for malloc_type
functions on the latest macOS Sequoia. The problem arises due to changes
in how these functions are named in the dynamic symbol table of the
system libraries that are part of the linker cache like the ones that
contain the C++ runtime.

This fix ensures that Memray correctly intercepts and tracks allocations
made by malloc_type functions, fixing tracking allocations in the C++
runtime and other system libraries on macOS Sequoia.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,6 +72,9 @@ jobs:
     container:
       image: alpine
       options: --cap-add=SYS_PTRACE
+    env:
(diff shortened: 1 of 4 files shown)


godlygeek (approved):
LGTM

PR #700, opened 2024-11-27
Protect against deregistered profile functions in greenlet switches

pablogsal:
When greenlet tracking is enabled is possible that we run into a
situation where the function that recreates the Python stack in our TLS
variable after a greenlet switch is called **after** the profile
function has been deactivated. In this case, recreating the Python stack
is wrong as we are no longer tracking POP/PUSH events so when the stack
is inspected later nothing guarantees that the frames are still valid.

**Describe your changes**

**Testing performed**

**Additional context**

### DIFF
--- a/src/memray/_memray/tracking_api.cpp
+++ b/src/memray/_memray/tracking_api.cpp
@@ -1216,9 +1216,17 @@ Tracker::beginTrackingGreenlets()
 void
 Tracker::handleGreenletSwitch(PyObject* from, PyObject* to)
 {
+    // We must stop tracking the stack once our trace function is uninstalled.
+    // Otherwise, we'd keep referencing frames after they're destroyed.
+    PyThreadState* ts = PyThreadState_Get();
+    if (ts->c_profilefunc != PyTraceFunction) {
+        return;
+    }
+
     // Grab the Tracker lock, as this may need to write pushes/pops.
     std::unique_lock<std::mutex> lock(*s_mutex);
     RecursionGuard guard;
+
     PythonStackTracker::get().handleGreenletSwitch(from, to);
 }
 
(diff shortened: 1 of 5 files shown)


src/memray/_memray/tracking_api.cpp
```diff
@@ -1219,6 +1219,16 @@ Tracker::handleGreenletSwitch(PyObject* from, PyObject* to)
     // Grab the Tracker lock, as this may need to write pushes/pops.
     std::unique_lock<std::mutex> lock(*s_mutex);
     RecursionGuard guard;
+
+    // Check if the trace function is still installed in the current thread as
+    // it's possible that the thread is dying and the trace function has been
+    // uninstalled and therefore the stack will not be properly tracked so we
+    // should not recreate the current thread stack.
```
godlygeek, line 1226: This is more explanation than we need, I think. I'd just say:
```suggestion
    // We must stop tracking the stack once our trace function is uninstalled.
```
If you think it deserves more than that, maybe one more line:
```cpp
    // Otherwise, we'd keep referencing frames after they've been destroyed.
```

src/memray/_memray/tracking_api.cpp
```diff
@@ -1219,6 +1219,16 @@ Tracker::handleGreenletSwitch(PyObject* from, PyObject* to)
...
+    PyThreadState* ts = PyThreadState_Get();
+    if (ts->c_profilefunc != PyTraceFunction) {
+        return;
+    }
```
godlygeek, line 1230: It's a nit, but we could move check to the top of the function. We don't need to be holding the tracker lock or the recursion guard in order to check this.

godlygeek (commented):
Just a few minor nits

godlygeek (approved):
LGTM. I've rebased this and made all of the changes I suggested in my comments above. Check my work and feel free to land if you don't have any objections.

PR #737, opened 2025-04-03
Fix ARM Linux infinite loop in attach / live mode

godlygeek:
On ARM Linux, `char` is unsigned, unlike on all of our other platforms.
Since EOF is equal to `-1`, this means that we need to be careful to
avoid assuming that:

    EOF == static_cast<char>(EOF)

On all of our other platforms that's true, but on aarch64 Linux machines
that compares `-1 == 255` instead. And when you want to loop until EOF
but don't recognize EOF, that's an infinite loop...

Also fix a bug where we could potentially return from extension code to
Python code without holding the GIL.

Also, have CI run pytest unbuffered and in verbose mode. This way, if
a crash occurs, we should at least be able to see what test was running
when things went wrong.

Also: Prevent IPython from writing history files in our tests.
We don't need the history features, so the extra thread managing
a sqlite database just gets in our way.

Also, mark the flamegraphs in our docs as binary files.
This stops them from having textual diffs shown for them, and from
showing up in the result of `git grep`. We never want to see the
contents of these files, since they're programmatically generated.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -111,7 +111,7 @@ jobs:
           CIBW_ARCHS_LINUX: auto
           CIBW_PRERELEASE_PYTHONS: True
           CIBW_TEST_EXTRAS: test
-          CIBW_TEST_COMMAND: python -um pytest -vvv {package}/tests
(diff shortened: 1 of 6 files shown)


pablogsal (approved):
Exquisite! Real good job man ✨

Issue #740, opened 2025-04-04
Consistently choose between SI or binary units (MB or MiB, for instance)


godlygeek:
Some of our reporters use megabytes (1 MB is 1000 * 1000 bytes).

Other reporters use mebibytes (1 MiB is 1024 * 1024 bytes).

And worse still, some use a mix of the two (the line graph at the top of a flame graph uses megabytes, and the cells in the flame graph use mebibytes).

We should pick one to standardize on and be consistent about it!

ajit421:
I have a prototype solution ready for this issue. Can I go ahead and submit a PR? If possible, could anyone assign this issue to me?

godlygeek:
Sure @ajit421, feel free to submit a PR. I've assigned this issue to you.

adesidaseyi:
Hi! I have spent some time going over the source code of the reporters and found a few things:

1. The non-HTML reporters (like the Stats and Tree reporters) use the `size_fmt` function from the `memray` module, which converts raw byte values (e.g. 1024) into readable strings (e.g. "1KB") using multiples of 1024. Since it uses 1024 as the base, the technically correct unit should be binary units (i.e., "1KiB" instead of "1KB"). Alternatively, if SI units ("1KB") are preferred, the multiple should be 1000.

2. The Table and Flamegraph (tooltip) reporters use the humanFileSize function, which already uses binary units (base 1024) and correctly outputs values like "1.0 KiB"

3. The Temporal Flamegraph reporter uses Plotly.js, and it looks like the memory values shown on the y-axis and in the tooltips are controlled by Plotly’s layout settings. These seem to already follow SI units (base 1000), which Plotly seems to support out of the box.

Based on this, I think choosing **SI** units (base 1000) across all reporters could be the more consistent and user-friendly option, since most users are probably more familiar with "KB/MB/GB" than "KiB/MiB/GiB". Only a few reporters currently use binary units.

It seems whichever approach we choose would require updating some of the tests for the reporters. I'm not quite sure about this part yet since I just started looking at them.

This is my first time contributing to OSS, and I'm quite interested in the kind of work being done here, so I'm happy to hear any feedback or corrections.

adesidaseyi:
Hi @godlygeek, just wanted to follow up on this.

godlygeek:
Sorry, I got swamped by other things and let this fall through the cracks.

[...]

PR #741, opened 2025-04-06
Standardize all memory units to MiB across reporters and flame graph

ajit421:
#740  

**Describe your changes**

Standardized all memory unit representations to MiB (1 MiB = 1024 × 1024 bytes) across all reporters and flamegraph visualizations using a common bytes_to_mib() function.

**Testing performed**

Verified the output of text and HTML reports. Checked flamegraph tooltips and labels for consistent MiB usage. Ran all tests to ensure correctness.

**Additional context**
Based on discussion in issue #740, MiB was chosen for accuracy and consistency in memory profiling.
Utility function bytes_to_mib() is now reusable and can be adopted in future memory-related features.
Let me know if you'd like this logic abstracted further or applied in additional reporters.

### DIFF
--- a/src/memray/_memray.pyx
+++ b/src/memray/_memray.pyx
@@ -756,7 +756,7 @@ def print_greenlet_warning():
 
 cdef millis_to_dt(millis):
     return datetime.fromtimestamp(millis // 1000).replace(
-        microsecond=millis % 1000 * 1000).astimezone()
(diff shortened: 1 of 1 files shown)


godlygeek:
The change you made here is not correct. The function you changed isn't related to converting a number of bytes to a human readable string at all.

PR #752, opened 2025-04-13
Add Python object reference tracking

pablogsal:
This feature adds the capability to track Python objects throughout
their lifecycle in memray. The implementation introduces a reference
tracking system that records PyObject creation and destruction events,
storing them as TrackedObject records in the memray capture file with
associated stack traces. The Tracker class now accepts a
reference_tracking flag to enable this functionality and provides a
get_surviving_objects() method to retrieve live objects at the end of
tracking.

Object tracking works across threads and supports native, Python, and
hybrid stack traces, providing insights into Python object
lifetimes. This enhancement extends memray beyond memory allocation
tracking to give developers a powerful tool for understanding object
lifecycles and identifying potential reference leaks.

With this addition, memray can now correlate memory allocations with the
Python objects that trigger them, enabling more precise debugging of
memory usage patterns. The feature is particularly useful for
identifying objects that persist longer than expected or detecting
unexpected object creation patterns. By tracking both memory allocations
and object references, memray provides a more complete picture of a
program's memory behavior, helping developers optimize memory usage and
resolve leaks more effectively.

This implementation leverages the new Python reference tracing API that
was introduced in CPython 3.13.

**Describe your changes**

**Testing performed**

**Additional context**


pablogsal:
Tests require 3.13.3 at least because before there is a bug that causes some objects to be identified wrongly. cibuildwheel only has this when https://github.com/pypa/cibuildwheel/pull/2359 passes

joerick:
I just released cibuildwheel with 3.13.3 support!

godlygeek (changes requested):
Lots of comments, mostly pretty small things. The overall approach looks good to me. I have some concerns about some of the names, and how we handle the uninstallation of the hooks (and maybe even the installation?) in the free-threading world.

src/memray/_memray.pyx
```diff
@@ -365,6 +369,100 @@ cdef class AllocationRecord:
...
+        assert self._reader.get() != NULL, "Cannot get thread name without reader."
+        return self._reader.get().getThreadName(self.tid)
+
+    def stack_trace(self, max_stacks=None):
+        cache_key = ("python", max_stacks)
+        if cache_key not in self._stack_trace_cache:
+            self._stack_trace_cache[cache_key] = stack_trace(
+                self._reader.get(),
+                self.tid,
+                None,  # No allocator for objects
```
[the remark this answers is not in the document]
godlygeek: Hm, actually: do you want to record the stack associated with an object destruction, or not? The point of passing the allocator here is to allow `stack_trace` to raise an exception if you ask for the stack trace for a deallocation record. Should it also raise an exception if you ask for the stack trace of an object destroyed record? If so, passing `None` here isn't the right thing to do. And if not, we should be capturing the native stack for object destructions, but currently are not.

src/memray/_memray.pyx
```diff
@@ -732,6 +844,27 @@ cdef class Tracker:
...
+
+        Returns:
+            list: A list of objects that were alive at the end of the tracking period.
+        """
+        if sys.version_info < (3, 13, 3):
+            raise RuntimeError(
+                "Python object reference tracking requires Python 3.13.3 or later. "
+                f"Current version: {sys.version_info.major}.{sys.version_info.minor}"
+            )
+        return self._surviving_objects
```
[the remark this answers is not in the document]
pablogsal: We can easily return a tuple

tests/conftest.py
```diff
@@ -5,7 +5,7 @@
 from packaging import version
 
 SNAPSHOT_MINIMUM_VERSIONS = {
-    "textual": "3.6.0",
+    "textual": "6.1.0",
```
godlygeek, line 8: Wait, I don't understand - why do the tests now need Textual 6.1.0 if all of the snapshots generated with 3.6.0 are still passing the test suite?

tests/integration/test_record_writer.py
```diff
@@ -299,3 +303,120 @@ def test_decoding_line_numbers(tmp_path):
             ("bar", "bar b.py", 16),
             ("foo", "foo a.py", 5),
         ]
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 13), reason="Object tracking requires Python 3.13+"
+)
```
godlygeek, line 310: As above, there's no reason to skip this on <3.13, we should just allow the test record writer to write these records for all versions.
```suggestion
```

tests/integration/test_record_writer.py
```diff
@@ -299,3 +303,120 @@ def test_decoding_line_numbers(tmp_path):
...
+        track_object_lifetimes=True,
+        file_format=FileFormat.ALL_ALLOCATIONS,
+        main_tid=1,
+    )
+
+    # Write some basic setup records
+    assert writer.write_thread_record(1, "MainThread")
+    assert writer.write_code_object(1, "test_func", "test.py", b"", 1)
+    assert writer.write_frame_push(1, 1, 10, True)
+
```
godlygeek, line 330: Why bother to write these records if you make no assertions about them?
```suggestion
```

tests/integration/test_record_writer.py
```diff
@@ -299,3 +303,120 @@ def test_decoding_line_numbers(tmp_path):
...
+    ]
+
+    # We should have 4 object records
+    assert len(object_records) == 4
+
+    # Check the object records are correctly formatted
+    assert any("address=0x12345678" in record for record in object_records)
+    assert any(
+        "0x87654320" in record or "0x87654321" in record for record in object_records
+    )
```
godlygeek, line 362: Please make these make an assertion about the exact text of the record as printed by `memray parse`, like `test_write_basic_records` does it. As written, this is too generic to prove that the test record writer is working correctly. This test would pass if it always wrote created records and never wrote deleted records, for instance.

tests/integration/test_record_writer.py
```diff
@@ -299,3 +303,120 @@ def test_decoding_line_numbers(tmp_path):
...
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 13, 3),
+    reason="Python object reference tracking requires Python 3.13 or later",
+)
```
godlygeek, line 368:
```suggestion
```

tests/integration/test_record_writer.py
```diff
@@ -299,3 +303,120 @@ def test_decoding_line_numbers(tmp_path):
...
+        track_object_lifetimes=True,
+        file_format=FileFormat.AGGREGATED_ALLOCATIONS,
+        main_tid=1,
+    )
+
+    # Write some basic setup records
+    assert writer.write_thread_record(1, "MainThread")
+    assert writer.write_code_object(1, "test_func", "test.py", b"", 1)
+    assert writer.write_frame_push(1, 1, 10, True)
+
```
godlygeek, line 387:
```suggestion
```

tests/integration/test_record_writer.py
```diff
@@ -299,3 +303,120 @@ def test_decoding_line_numbers(tmp_path):
...
+    assert len(surviving_object_records) == 2
+
+    # Check the surviving object records contain the expected addresses
+    record_text = " ".join(surviving_object_records)
+    assert (
+        "0x87654320" in record_text or "0x87654321" in record_text
+    )  # 0x87654321 >> 3 = 0x10eca864
+    assert (
+        "0xaabbccd8" in record_text or "0xAABBCCDD" in record_text
+    )  # 0xAABBCCDD >> 3 shifted and rounded
```
godlygeek, line 422: As above, please make assertions about the exact text, rather than these `in` checks. You're trying to save us maintenance work by making it so that, if `memray parse` starts printing extra fields, we don't need to update this test, but we _should_ update this test if `memray parse` starts printing extra fields, exactly because we need to prove that the test record writer can write those new fields!

godlygeek (changes requested):
The `FileReader.__cinit__` has a `while True:` loop reading records from the record reader to find the high water mark. The last version of this PR I reviewed had a:
```py
                elif ret == RecordResult.RecordResultObjectRecord:
                    pass
```
in that loop so it would ignore object records. You removed that in this version, so that it breaks if it hits an object record. As far as I can see, that's wrong, and will cause it to break early - am I missing something? Is there some reason why the reader is guaranteed to never emit an object record?

tests/conftest.py
godlygeek: For posterity: one of the tests was flaky and failing due to a race. We're working around it by adding a longer sleep.

godlygeek (approved):
OK, at this point this LGTM. I've pushed some very small fixups to your work - rewording a comment, and sorting the records spit out by `memray parse` in the aggregated file mode test so that the test behaves deterministically across hash map algorithms.

Separately from that, I've added 3 commits to fix unrelated stuff:
- deleting a duplicated run of Coverage that we were making from our GitHub Actions workflow (we had coverage analysis in two different workflows)
- updating the other Coverage run to use Python 3.13 on Ubuntu 24.04, so that it sees the new tests for object tracking as covered. Previously it ran on 3.10, and so all the tests were skipped, which pushed our coverage analysis below the required 90% threshold.
- adding an extra pause to our Textual snapshot tests to work around some test flakiness.

PR #759, opened 2025-05-08
Fix temporal high water mark flame graphs

### LINKED ISSUE #757: Mismatch between temporal and non-temporal flame graph
I am having similar symptoms in my project:
* The temporal view's yellow highlight does not correspond to the curves in the graph. The explanation above that there's an spike between two keyframes from the graph seems reasonable but as I move the window around the timeline the yellow highlight seems *consistently* wrong. I guess it's possible that there's some consistent spiking (I guess large but very temporary allocations?) but there's different code running along different parts of the timeline so it seems unlikely to me.
* The total memory usage shown in the flamegraph is much lower than the numbers shown on the usage graph. For example, here I have zoomed in on an area with a memory usage consistently around 20GB and the total listed in the flamegraph is 917MB.

Screenshot 2025-04-23 at 2 57 40 PM

* The docs say that a flamegraph without `--temporal` should find the same peak usage as the temporal graph before changing the inspection window. When I produce a non-temporal graph, I see a different flamegraph which does match with the visible peak of usage. Here is the initial view of the temporal graph:

temporal

And the non-temporal one:

Screenshot 2025-04-23 at 3 04 51 PM

This kinda suggests to me that peak finding is somehow not working in temporal mode only? 

I am checking to see if I can share the profile with you. In the meantime, if it helps, I'm using memray 1.17.1 on macOS 15.4.1 (arm64) with Python 3.11.9 as `uv run python -m memray run -o output.bin --native my_module` and then `uv run python -m memray flamegraph [--temporal] --force output.bin` and then opening the resulting graph in Chrome.

_Originally posted by @jder in https://github.com/bloomberg/memray/discussions/738#discussioncomment-12926841_

godlygeek:
Closes #757 

There were two problems here, a very big one and a relatively small one.
The big one is that, when we needed to drop memory records due to
`max_memory_records` being exceeded, we correctly adjusted temporal
leaks flame graphs to properly match up the allocation records that make
up the flame graph to the memory records that make up the line chart,
but we missed making that same adjustment for temporal high water mark
flame graphs. The small one problem was an off-by-one: even for temporal
leaks flame graphs, instead of keeping allocation snapshots as of the
0th, and Nth, and 2Nth memory record, we were incorrectly keeping
snapshots as of the N-1th and 2N-1th and 3N-1th.

Update our tests to work around https://github.com/Textualize/textual/issues/5788

Also, remove a hack from our snapshot tests that is no longer necessary.


pablogsal (approved):

pablogsal:
🤦‍♂️

pablogsal (approved):

Issue #768, opened 2025-05-19
memray's wheels contain the ./memray/_memray/ directory


sarahmonod:
Memray's wheels currently contain a ./pystack/_pystack/ directory (see [memray-1.17.2-cp310-cp310-manylinux_2_12_x86_64.manylinux2010_x86_64.whl](https://inspector.pypi.io/project/memray/1.17.2/packages/4c/ff/957f855498c14fb166da39ff36e77cc97f12467fd60e414163c0ae7c1f37/memray-1.17.2-cp310-cp310-manylinux_2_12_x86_64.manylinux2010_x86_64.whl/) for instance). This contains some of Memray's source code. This is harmless, but not necessary or useful, and we should remove it by telling setuptools that memray/_memray/ is not a package.

lkollar:
Fixed in #773. Confirmed by checking wheel contents of the latest release as well.

PR #773, opened 2025-05-22
Wheelfix

abstractedfox:
Addresses #768 

Addresses the issue by being more specific about which file types should be included. This was tested by comparing the contents of the wheel before and after the change. Notably, trying to use exclude_package_data to specifically exclude memray/_memray didn't seem to work in this case.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -333,6 +333,7 @@ setup(
         compiler_directives=COMPILER_DIRECTIVES,
     ),
     include_package_data=True,
+    exclude_package_data={"memray": ["_memray/*"]},
     install_requires=install_requires,
     extras_require={
         "test": test_requires,


godlygeek:
Huh, wait, there's something weird going on here... When I test your change, indeed I don't get the `memray/_memray/` directory in the built wheel, _but_ I also don't get it when I test _without_ your change! Indeed, 7f6ee7c56c17cda9b9d16705980c68743ac7513c was supposed to have removed the `memray/_memray` directory last year, and as far as I can see that's working - _except_ that our published wheels still contain that directory! This must mean that cibuildwheel is building wheels that include that directory even though it's not included when I build locally on my laptop.

I have no idea why that would be...

godlygeek:
> Notably, trying to use exclude_package_data to specifically exclude memray/_memray didn't seem to work in this case.

_And_, I think that the caching is why you couldn't get `exclude_package_data` to work. After blowing away the cache, it's enough to just do:
```py
exclude_package_data={"memray._memray": ["*"]},
```

godlygeek (approved):
I've updated this to just use `exclude_package_data` now that I got it working, but I've left the commit crediting you. Thanks for the help with this, @abstractedfox!

abstractedfox:
@godlygeek Thanks for the explanation too!

PR #774, opened 2025-05-27
Standardize byte amounts to SI units across all reporters

adesidaseyi:
*Issue number of the reported bug or feature request: #740*

**Describe your changes**
Ensure byte amounts in reporters are correctly represented using SI units.
This eliminates inconsistencies in values displayed across different reporters.

Use proper representation of SI unit symbol for kilobytes as 'kB' and not 'KB', for consistency with our decision to use SI unit representation across all reporters.

Also, set Jinja filesizeformat filter to use SI units for `Peak memory usage` value in the base HTML template for Flamegraph and Tree reporters.

Build JavaScript assets to reflect the change to SI units in Flamegraph and Table reporters.

**Testing performed**
Updated tests for TUI reporters (Textual snapshots), Tree reporter, Stats reporter, Summary reporter, and Table reporter to reflect the switch to SI unit formatting for byte sizes. The `size_fmt` function was modified to use SI units for byte amounts. After reviewing the rendered output, no issues were found with this change, so tests and snapshots were updated accordingly.

**Additional context**


adesidaseyi:
Thanks so much for accepting my PR @godlygeek and @pablogsal !
It means a lot! 🙏

PR #777, opened 2025-06-06
Fix a race condition under Textual 3.3.0

godlygeek:
Textual 3.3.0 introduced a race condition that leads to errors if an app's message queue is first accessed from a different thread. See https://github.com/Textualize/textual/issues/5845

This potentially affects the live mode TUI, which starts a background thread early on in app startup to monitor the memory usage reports coming in from the application being profiled and update the screen. We can work around this by starting our update thread a bit later in the startup process, at a point where the message queue will already have been accessed by the thread running the app.

### DIFF
--- a/src/memray/reporters/tui.py
+++ b/src/memray/reporters/tui.py
@@ -768,8 +768,6 @@ class TUIApp(App[None]):
         super().__init__()
 
     def on_mount(self) -> None:
-        self._update_thread.start()
-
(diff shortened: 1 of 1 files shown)


godlygeek:
While this _theoretically_ could affect real runs of Memray, I think that the race is so hard to provoke that this isn't worth a news entry, so I've marked it as `skip news`. @pablogsal let me know if you disagree.

pablogsal (approved):

Issue #470, opened 2023-09-27
--no-web Packaging Option


Delengowski:
### Is there an existing proposal for this?

- [X] I have searched the existing proposals

### Is your feature request related to a problem?

I have an issue utilizing the flamegraph and table output options when using memray on an airgapped system. Since the system does not have internet access, connecting to the CDNs to get the javascript is not possible and the web pages aren't fully loaded.

### Describe the solution you'd like

I would like it if memray would in a `--no-web` packaging option that would include the necessary html/css/js files as part of memray and have the the generated html to point accordingly.

### Alternatives you considered

Putting the files on the air gapped system and changing the links the html to point to the local file system path.

Delengowski:
I'd like to take the task of making a PR for this but am a little lost on how to best go about funneling it through to ultimately make say a conda package that has this option

godlygeek:
Honestly, I'm not sure what work is required here either. The biggest part of any PR here is likely figuring out exactly what files are fetched dynamically by our flame graphs as they exist today (using the developer tools of the browser should be able to get us this), and then figuring out what would be involved in vendoring a copy of those into the Memray repo. Once we have a vendored copy of each of those dependencies, it should "just" be a matter of exposing a switch that changes whether our Jinja templates include URLs linking to external resources or includes the contents of those resources directly.

If we find that some of these dependencies are difficult to vendor, or have licenses that are incompatible with us vendoring them instead of loading them dynamically, this might be a nonstarter.

vitorespindola:
Flamegraph report:

Bootstrap - MIT License - https://github.com/twbs/bootstrap/blob/main/LICENSE
D3 Flamegraph - Apache License 2.0 - https://github.com/spiermar/d3-flame-graph/blob/master/LICENSE
JQuery - MIT License - jquery.org/license
Lodash - MIT License - https://lodash.com/license
plotly.js - MIT license - https://github.com/plotly/plotly.js/blob/master/LICENSE
popper.js(Now Floating UI) - MIT License - https://github.com/floating-ui/floating-ui/blob/v2.x/LICENSE.md
D3 - ISC License - https://github.com/d3/d3/blob/main/LICENSE

kanadn:
I have a prototype solution ready for this issue. Can I go ahead and submit a PR? If possible, could anyone assign this issue to me?

pablogsal:
@godlygeek WDYT?

godlygeek:
[...]

PR #790, opened 2025-07-07
Add --no-web to flamegraph reporters

pablogsal:
Closes: #470


Delengowski:
Thank you taking this on Pablo. I was just revisiting it to tackle while on vacation from work, and was looking at how other projects like Plotly Dash do this.

So the plan is to just vendor all the JavaScript script but conditionally use it. I thought the project may have wanted to just make a separate wheel build option to vendor if specified. This seems easier.

godlygeek (approved):
I've pushed two small fixups, but otherwise this LGTM.

This is much, much simpler than I was afraid this feature would be!

PR #796, opened 2025-07-26
Remove setuptools from rpath_extension test

pablogsal:
Some of our tests fail because setuptools in not installed
in the virtual environment and we cannot build extensions.

**Describe your changes**

**Testing performed**

**Additional context**

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -61,6 +61,9 @@ jobs:
           - os: ubuntu-latest
           - wheel_type: manylinux_aarch64
             os: ubuntu-24.04-arm
+          - wheel_type: manylinux_i686
+            os: ubuntu-latest
+            cibw_archs_linux: auto32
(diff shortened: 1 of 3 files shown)


godlygeek:
> Some of our tests fail because setuptools in not installed
> in the virtual environment and we cannot build extensions.

It doesn't seem like switching to `distutils` is a reasonable fix for this, now that `distutils` is no longer part of the standard library.

Our test dependencies have `setuptools; python_version >= '3.12'` in them. If you hit a failure with a version older than `3.12`, maybe all we ought to do is update that to be version-independent (in both `setup.py` and `requirements-test.txt`) instead of switching back to `distutils` instead of `setuptools`.

pablogsal:
The test were failing in old versions and all our other extensions were using distutils anyway so that's why I chose this particular fix. If you prefer to change the version limit in setup tools we can go that route I don't feel very strongly about it

PR #817, opened 2025-08-20
Add supprt for new c23 deallocators

### LINKED ISSUE #725: Intercept new `free_sized` and `free_aligned_sized` functions

naamanu:
closes #725 

This commit introduces 2 new functions to add support for `free_sized` and `free_aligned_sized` deallocators introduced in c23.

Tests ensure that `free_sized` and `free_aligned_sized` are skipped if unavailable otherwise we get a `FREE` record recorded for them.


godlygeek (changes requested):
This looks like a good start, but I see a lot of things here that don't look right.

godlygeek:
Unfortunately, there's not much we can do here until major libc implementations begin to support these C23 deallocators. I'm going to close this PR, since there's no sense having it sit open while it's blocked and untestable. Once the C23 deallocators are available in glibc, we can reopen this PR, but I don't want to leave the PR open waiting for something we have no control over to happen.

Thanks for the work you put in here, @naamanu!

godlygeek:
Hm. The GitHub Actions don't want to run on this reopened PR. I'll open a new PR instead.

PR #826, opened 2025-09-05
Expose the record writer for validating arbitrary files

godlygeek:
Add a `RecordWriterTestHarness` class for testing and validating Memray record files by allowing direct writing of test records. This provides a way to create controlled test files with known contents for validating the file format and reader functionality. The implementation supports writing all record types including memory snapshots, allocations, frames, and thread information, making it possible to test edge cases and complex scenarios that would be difficult to trigger in normal operation.

### DIFF
--- a/src/memray/_memray/record_reader.cpp
+++ b/src/memray/_memray/record_reader.cpp
@@ -1201,7 +1201,7 @@ RecordReader::dumpAllRecordsFromAllAllocationsFile()
                     Py_RETURN_NONE;
                 }
 
-                printf("time=%ld memory=%" PRIxPTR "\n", record.ms_since_epoch, record.rss);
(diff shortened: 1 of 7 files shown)


pablogsal (approved):
Exquisite

src/memray/_memray/record_reader.cpp
```diff
@@ -1100,10 +1100,11 @@ RecordReader::dumpAllRecordsFromAllAllocationsFile()
                             "<unknown allocator " + std::to_string((int)record.allocator) + ">";
                     allocator = unknownAllocator.c_str();
                 }
-                printf("address=%p size=%zd allocator=%s\n",
+                printf("address=%p size=%zd allocator=%s native_frame_id=%zd\n",
```
pablogsal, line 1103: Oh, good catch!

PR #832, opened 2025-09-29
Fix enum value collision between AllocatorType and ObjectTrackingEvent

pablogsal:
The ObjectTrackingEvent enum values were colliding with AllocatorType
values. This caused a bug where OBJECT_DESTROYED (value 2) was
incorrectly matching PYOBJECT_MALLOC in the deallocation check.

This would incorrectly treat pymalloc allocations as deallocations when
checking stack traces. The bug went undetected because we lacked tests
for Python allocations with trace_python_allocators enabled.

Fixed by changing ObjectTrackingEvent values to 10 and 20 to avoid
collision. Added test coverage for pymalloc allocations with Python
stack traces to prevent regressions.

**Describe your changes**

**Testing performed**

**Additional context**


godlygeek (approved):
I pushed three tiny changes:
- a fix to the grammar in the news entry
- a change to the enumerators' values to ensure they don't conflict (the `10` you chose still conflicted; `ALIGNED_ALLOC == 10 == `OBJECT_CREATED`!) - I just picked a different convention; one enum's enumerators count up from 1 and the other's count down from negative 1, and we'll never get conflicts.
- I added a comment explaining that they need to not conflict.

Issue #839, opened 2025-10-22
Tutorials Update for exercise 1


jcarnaxide:
### Is there an existing proposal for this?

- [x] I have searched the existing proposals

### Is your feature request related to a problem?

When running tutorial exercise 1, the flamegraph that is currently generated is not a great visual way to explore the flamegraph. Specifically the tree has only 1 path, and its hard to show case that the x-axis of the flamegraphe represents size of memory allocated.

### Describe the solution you'd like

An updated tutorial example where we can see distinctly differently sized branches in the tree of the flamegraph.

PR #840, opened 2025-10-23
Tutorial exercise 1 uses example that is more conducive to learning

jcarnaxide:
#839 

Updated the values as well as layout of code in exercise 1 of the hands-on tutorials. Now the allocations appear as 3 separate branches in the flamegraphs, each of different relative size. This allows for a better example to illustrate how flamegraphs are meant to be read (specifically showing the x-axis represents relative memory allocation amount.)

I ran this new example locally, with/without solution implemented. New flamegraphs look good. Tests fail/pass the way expected. I also included an updated image of the generated flamegraph with the initial codeset (as intended to be viewed while following the tutorial).


jcarnaxide:
@godlygeek I believe this PR is suitable for review. Are there any other steps I should follow to indicate this?

PR #859, opened 2025-11-28
Implement fast unwinding using ghost unwind

pablogsal:


godlygeek:
@pablogsal - what's the state on this one? Should I try rebasing it and review it, or was there more work you wanted to do on it?

pablogsal:
I'm unconvinced this is worth the cost. I have benchmarked this and althogh I can see speedups similar to the original ones I measured there are still edge cases. I may explore this again in the future but for now let's close it

PR #871, opened 2026-01-22
tests: Reduce test flakiness on macOS

godlygeek:
On Linux systems, we use `/proc/$pid/syscall` to wait until the tracked process is blocked in the `connect` or `accept` call we want to interrupt. There's no analogous mechanism on macOS, so we just sleep there. On the GitHub Actions macOS runners, sometimes this delay is not long enough. Increase it to try to reduce test flakiness.

### DIFF
--- a/tests/integration/test_main.py
+++ b/tests/integration/test_main.py
@@ -82,8 +82,10 @@ def track_and_wait(output_dir, sleep_after=100):
 
 
 def _wait_until_process_blocks(pid: int) -> None:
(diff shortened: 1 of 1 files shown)


pablogsal (approved):

godlygeek:
> The smart solutions in tests often blow up

Says the man who wrote the test code that leverages `/proc/$pid/syscall` on Linux 😆 

OK. Yeah, I think the low tech hack has the best ROI for now. We can revisit if this doesn't improve the situation.

PR #892, opened 2026-03-30
Deduplicate rows in table reporter with identical displayed fields

### LINKED ISSUE #857: Duplicate entries in `memray table --leaks` reporter.
### Current Behavior

I am testing Memray's capabilities to identify memory leaks in python programs that use C extensions. I have a program that uses a library, which  has a known memory leak coming from an extension code, and I am running Memray in several configurations to find the one shows the leak most prominently.

When I run my program with `memray --native` mode for 1 hr, my profile file  is about 2GB without using aggregated mode or  6MB in aggregated mode (yay for this feature!). 

When I create a table profile, it has about 25k rows, yet many of the rows are duplicates. I have converted the html report to a CSV, file and then did further aggregation to remove duplicates by adding together allocations that happened in the same place.

This reduced the number of rows to ~200; when I removed the Thread_ID column and aggregated the entries again, I ended up with only 50 rows, and had a meaningful information about my leak.  

Here are sample entries from the table that demonstrate duplication 

```
Thread_ID,Size,Allocator,Allocations,Location
0x1,192,malloc,1,operator new(unsigned long) at <unknown>:0
0x1,56,malloc,1,operator new(unsigned long) at <unknown>:0
...
0x1,944,malloc,1,_PyMem_RawMalloc at Objects/obmalloc.c:99
0x1,944,malloc,1,_PyMem_RawMalloc at Objects/obmalloc.c:99
...
0x25 (fn_api_status_handler),328,realloc,1,upb_Arena_InitSlow at <unknown>:0
0x38 (Thread-30 (_run)),328,realloc,1,upb_Arena_InitSlow at <unknown>:0
0x1c (Thread-6 (_run)),328,realloc,1,upb_Arena_InitSlow at <unknown>:0
...
0x25 (fn_api_status_handler),71,malloc,1,<unknown> at <unknown>:0
0x25 (fn_api_status_handler),87,malloc,1,<unknown> at <unknown>:0
...
```

In my case the application is creating several ephemeral threads, and thread ID info is not meaningful, it would be nice to have an option to exclude ThreadID from the report. But even that aside , it seems that we should be adding together allocations that are happening at the same location, increasing the Size and Total allocations instead.

I have also tried using `--trace-python-allocation`, which didn't have a meaningful impact.

I also verified that duplicate entries appear in table report when not using `--native` mode. It happens but at a smaller ratio: post-processing resulted in 3x reduction in number of rows.  

### Expected Behavior

In Table report, there should be only 1 row for each (Thread_ID, Size, Allocator, Location) tuple. 

Bonus: provide the option to exclude ThreadID from the report.

### Steps To Reproduce

My setup is somewhat involved, I am profiling a data processing pipeline, also discussed in https://github.com/bloomberg/memray/discussions/852. 
I am happy to attach the collected profiles if it helps. If we must have a repro for a meaningful investigation, let me know.

### Memray Version

1.19.1

### Python Version

3.10

### Operating System

Linux

### Anything else?

_No response_

abhiojha8:
*Issue number of the reported bug or feature request: #857*

**Describe your changes**

The table reporter creates one row per `AllocationRecord` but only displays the top stack frame (`max_stacks=1`). Records with different full call stacks that map to the same displayed top frame appeared as duplicate rows in the HTML table.

This change aggregates records in `TableReporter.from_snapshot()` by their displayed key `(tid, allocator, top_frame)`, summing `size` and `n_allocations`. No data is lost — the totals are preserved exactly.

Additionally, this adds a `--split-threads` option to the `table` command, matching the `flamegraph` reporter's existing flag. By default threads are merged; `--split-threads` keeps them separate.

Closes: #857

**Testing performed**

- All 630 tests pass (including 7 new unit tests and 1 updated integration test)
- Verified on the `exercise_3/lru_cache.py` tutorial: 21 leaked allocation records that previously produced 21 rows (with 18 visual duplicates) now correctly produce 3 deduplicated rows with identical size/allocation totals
- Pre-commit hooks all pass (ruff, black, isort, prettier, rstcheck, etc.)


tvalentyn:
seeing a lint error:

src/memray/reporters/table.py:39: error: Missing type arguments for generic type "tuple" [type-arg]

godlygeek:
Sorry for the delay. I started reviewing this months ago, but got completely sidetracked by something that's wrong - or at least surprising - about the existing C++ aggregation. The Cython `get_high_watermark_allocation_records` takes a `merge_threads=True`, which is plumbed down to the C++ `SnapshotAllocationAggregator::getSnapshotAllocations`, but while that does merge allocations with identical stacks into a single allocation record, that merged allocation record keeps the TID of the first seen allocation with that stack. This makes it so that the fix in this PR didn't fully work (duplicates could still show up in the table, because we could have different TIDs across allocations even with `merge_threads=True`, and so we could get two stacks with the same top frame but different TIDs).

Unfortunately, it also turned out to be annoying to fix that. We can't just blank out the TID at the point when the C++ aggregation happens, because it's used for the logic to skip initial frames on the main thread, above where Memray started tracking.

Coming back to this again now, I've just decided to make the minimal change to get the table reporter to not have duplicates, which is to pass the `merge_threads=True` flag down to the reporter as well, and have it ignore the TIDs when re-aggregating if that flag is passed. Not ideal, and a bit wasteful, but anything else requires way too many changes to justify making as part of this PR, unfortunately.

Other than that, I spotted some small things that were wrong and fixed those:

- The TID column was still shown in the HTML report even when threads were merged. I've updated it to drop that column when `merge_threads` is true.
- The type annotation for the aggregation dict used just `tuple` without any generic args. I've fixed it to `Tuple[str, str, str]` to correctly declare the type of the key for the dict (and used `Tuple` instead of `tuple` for consistency with the rest of the code base. We should switch everything over at some point now that we've dropped 3.8 support...)
- I updated the news entry to mention that we no longer split allocations by thread by default in the table reporter

sarahmonod (approved):

godlygeek:
Thanks for the contribution, @abhiojha8!

PR #893, opened 2026-04-01
Vendor Textual

lkollar:
Since this vendors the latest version of Textual which needs Python 3.9, this also drops support for 3.7 and 3.8 in Memray. If you think we should keep compatibility with those, we can downgrade Textual instead.


godlygeek:
I don't like needing to monkeypatch `pytest_textual_snapshot`, and I don't like needing to modify all of the vendored modules to do `from memray._vendor.textual import ...` instead of just `from textual import ...`

I suggest a different path: let's make sure that Textual only gets imported when you're running `memray tree` or `memray run --live` or `memray attach`, and not when you're doing `memray run`, and not in the process you're attaching to. If we can make it so that the `import textual` only happens in a different process than the user's code runs in, then we don't need to worry about conflicting versions of textual or of its dependencies, and we can play a simpler trick: only when we're about to `import textual`, we can add an extra directory to `sys.path` (or maybe `sys.meta_path`) to force the import to find our vendored copy instead of looking in the normal paths and finding the version (if any) that the user installed.

That would let us get rid of all of the import rewriting, and all of the monkeypatching, and makes it so that updating the vendored version of the library is as easy as unzipping the wheel in the right spot and adding its declared dependencies to our declared dependencies.

Does that strategy make sense, or is there some reason that wouldn't work as nicely as I'm thinking?

PR #898, opened 2026-04-03
Account for huge page allocations on Linux

pablogsal:


src/memray/_memray/tracking_api.cpp
```diff
@@ -54,6 +54,31 @@ starts_with(const std::string& haystack, const std::string_view& needle)
...
+        size_t* value_in_kb)
+{
+    size_t line_start = 0;
+    while (line_start < proc_status.size()) {
+        size_t line_end = proc_status.find('\n', line_start);
+        if (line_end == std::string::npos) {
+            line_end = proc_status.size();
+        }
+
+        std::string line = proc_status.substr(line_start, line_end - line_start);
```
godlygeek: Bit of a nitpick, but we should probably use a `string_view` here instead of a `string` to avoid an allocation for each line:

```suggestion
        std::string_view line = std::string_view(proc_status).substr(line_start, line_end - line_start);
```

We'd need to update `starts_with` to take the haystack as a `string_view` too.

I think that's an improvement, do you agree?

godlygeek:
Also, whether or not you want to switch to `string_view`, please amend your commit to add the DCO `Signed-off-by`, @pablogsal

pablogsal:
> Also, whether or not you want to switch to `string_view`, please amend your commit to add the DCO `Signed-off-by`, @pablogsal

Both done.I pulled the `/proc/self/status` parsing into a small helper and switched the line handling over to `std::string_view`, so we don’t allocate a new `std::string` for every line while scanning for `VmRSS` and `HugetlbPages`.  I also added a small deterministic test for the new parsing path so we explicitly cover folding `HugetlbPages` into RSS, and I amended the commit with the DCO sign-off too.

godlygeek (approved):
LGTM

PR #899, opened 2026-04-05
Add speedscope export timestamp fallback

pablogsal:
Build the speedscope export around sampled stacks and make the ordering
source explicit instead of relying on the reader's iteration order.

When a capture has exact allocation timestamps, record them in the
non-aggregated file format, expose them through the reader metadata, and
sort each aggregated speedscope sample by the earliest allocation time
that contributed to it. This keeps the sampled profile aligned with the
real execution order without changing the existing high-water-mark and
leak aggregation logic.

When exact timestamps are not available, switch the speedscope path to
temporal allocation records and recover a stable approximation from the
memory snapshots we already store. Leak exports keep only intervals that
survive to the end of the run and order them by the first snapshot where
they appear. High-water-mark exports evaluate intervals at the peak
snapshot and order them by the first snapshot that contributes to that
peak.

Keep this behavior speedscope-specific and reject `--aggregate` with
`--allocation-timestamps`, because aggregated captures discard the event
ordering that the export depends on. The patch also bumps the file
format/header metadata and adds command, reader/writer, and reporter
tests for both the exact-timestamp and fallback paths.

**Describe your changes**

**Testing performed**

**Additional context**

### DIFF
--- a/news/899.feature.rst
+++ b/news/899.feature.rst
@@ -1,0 +1,1 @@
+Add a ``speedscope`` output format for ``memray transform``.
--- a/src/memray/_ipython/flamegraph.py
+++ b/src/memray/_ipython/flamegraph.py
@@ -20,6 +20,11 @@ from memray import FileReader
 from memray.commands.common import warn_if_not_enough_symbols
 from memray.reporters.flamegraph import FlameGraphReporter
 
+_typed_cell_magic = cast(
+    Callable[[Callable[..., Any]], Callable[..., Any]],
+    cell_magic,
+)
+
 TEMPLATE = """\
 from memray import Tracker, FileFormat
 with Tracker(
@@ -133,7 +138,7 @@ def argument_parser() -> argparse.ArgumentParser:
 
 @magics_class
 class FlamegraphMagics(Magics):
-    @cell_magic  # type: ignore
+    @_typed_cell_magic
     def memray_flamegraph(self, line: str, cell: str) -> None:
         """Memory profile the code in the cell and display a flame graph."""
         if self.shell is None:
--- a/src/memray/_memray/snapshot.cpp
+++ b/src/memray/_memray/snapshot.cpp
@@ -111,9 +111,11 @@ Interval::rightIntersects(const Interval& other) const
 void
 SnapshotAllocationAggregator::addAllocation(const Allocation& allocation)
 {
+    SnapshotAllocation sequenced_allocation{allocation, d_index};
+
     switch (hooks::allocatorKind(allocation.allocator)) {
         case hooks::AllocatorKind::SIMPLE_ALLOCATOR: {
-            d_ptr_to_allocation[allocation.address] = allocation;
+            d_ptr_to_allocation[allocation.address] = sequenced_allocation;
             break;
         }
         case hooks::AllocatorKind::SIMPLE_DEALLOCATOR: {
@@ -124,7 +126,7 @@ SnapshotAllocationAggregator::addAllocation(const Allocation& allocation)
             break;
         }
         case hooks::AllocatorKind::RANGED_ALLOCATOR: {
-            d_interval_tree.addInterval(allocation.address, allocation.size, allocation);
+            d_interval_tree.addInterval(allocation.address, allocation.size, sequenced_allocation);
             break;
         }
         case hooks::AllocatorKind::RANGED_DEALLOCATOR: {
@@ -141,32 +143,38 @@ SnapshotAllocationAggregator::getSnapshotAllocations(bool merge_threads)
     reduced_snapshot_map_t stack_to_allocation{};
 
     for (const auto& it : d_ptr_to_allocation) {
-        const Allocation& record = it.second;
+        const auto& snapshot_allocation = it.second;
+        const Allocation& record = snapshot_allocation.allocation;
         const thread_id_t thread_id = merge_threads ? NO_THREAD_INFO : record.tid;
         auto loc_key = LocationKey{record.frame_index, record.native_frame_id, thread_id};
         auto alloc_it = stack_to_allocation.find(loc_key);
         if (alloc_it == stack_to_allocation.end()) {
-            stack_to_allocation.insert(alloc_it, std::pair(loc_key, record));
+            stack_to_allocation.insert(alloc_it, std::pair(loc_key, snapshot_allocation));
         } else {
-            alloc_it->second.size += record.size;
-            alloc_it->second.n_allocations += 1;
+            alloc_it->second.allocation.size += record.size;
+            alloc_it->second.allocation.n_allocations += 1;
+            alloc_it->second.sequence_number =
+                    std::min(alloc_it->second.sequence_number, snapshot_allocation.sequence_number);
         }
     }
 
     // Process ranged allocations. As there can be partial deallocations in mmap'd regions,
     // we update the allocation to reflect the actual size at the peak, based on the lengths
     // of the ranges in the interval tree.
-    for (const auto& [range, allocation] : d_interval_tree) {
+    for (const auto& [range, snapshot_allocation] : d_interval_tree) {
+        const Allocation& allocation = snapshot_allocation.allocation;
         const thread_id_t thread_id = merge_threads ? NO_THREAD_INFO : allocation.tid;
         auto loc_key = LocationKey{allocation.frame_index, allocation.native_frame_id, thread_id};
         auto alloc_it = stack_to_allocation.find(loc_key);
         if (alloc_it == stack_to_allocation.end()) {
-            Allocation new_alloc = allocation;
-            new_alloc.size = range.size();
+            SnapshotAllocation new_alloc = snapshot_allocation;
+            new_alloc.allocation.size = range.size();
             stack_to_allocation.insert(alloc_it, std::pair(loc_key, new_alloc));
         } else {
-            alloc_it->second.size += range.size();
-            alloc_it->second.n_allocations += 1;
+            alloc_it->second.allocation.size += range.size();
+            alloc_it->second.allocation.n_allocations += 1;
+            alloc_it->second.sequence_number =
+                    std::min(alloc_it->second.sequence_number, snapshot_allocation.sequence_number);
         }
     }
 
@@ -181,6 +189,8 @@ TemporaryAllocationsAggregator::TemporaryAllocationsAggregator(size_t max_items)
 void
 TemporaryAllocationsAggregator::addAllocation(const Allocation& allocation)
 {
+    SnapshotAllocation sequenced_allocation{allocation, d_index};
+
     hooks::AllocatorKind kind = hooks::allocatorKind(allocation.allocator);
     auto it = d_current_allocations.find(allocation.tid);
     switch (kind) {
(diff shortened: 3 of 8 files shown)


src/memray/_memray/record_writer.cpp
```diff
@@ -415,15 +425,24 @@ StreamingRecordWriter::writeThreadSpecificRecord(thread_id_t tid, const Allocati
...
+            && !writeVarint(record.size)))
+    {
+        return false;
+    }
+    if (d_header.has_allocation_timestamps) {
+        if (!writeVarint(record.timestamp_us - d_last.allocation_timestamp_us)) {
+            return false;
+        }
+        d_last.allocation_timestamp_us = record.timestamp_us;
+    }
```
godlygeek: Why use `writeVarint` directly instead of `writeIntegralDelta`? Is it just the mismatch between signed and unsigned? If so, I think we should just use a signed timestamp instead. Sure, we're using a monotonic clock and so the delta will never be negative, but that just cuts our range in half. 

Looks like the range is still almost 300 millenia even if we cut it in half:
```py
>>> 2**63 / 1000000 / 60 / 60 / 24 / 365
292471.208677536
```
so we probably don't need to care about discarding that last bit, and it makes our lives easier to be able to reuse the existing delta helpers.

src/memray/_memray/record_writer.cpp
```diff
@@ -415,15 +425,24 @@ StreamingRecordWriter::writeThreadSpecificRecord(thread_id_t tid, const Allocati
...
+        || (pointer_cache_index == -1 && !writeIntegralDelta(&d_last.data_pointer, record.address >> 3))
+        || (allocator_id >= 8 && !writeSimpleType(record.allocator))
+        || (d_header.native_traces
+            && hooks::allocatorKind(record.allocator) != hooks::AllocatorKind::SIMPLE_DEALLOCATOR
+            && !writeIntegralDelta(&d_last.native_frame_id, record.native_frame_id))
+        || (hooks::allocatorKind(record.allocator) != hooks::AllocatorKind::SIMPLE_DEALLOCATOR
+            && !writeVarint(record.size)))
+    {
+        return false;
+    }
```
godlygeek: To keep the diff smaller, I'd prefer:
```suggestion
    bool success = writeSimpleType(token)
           && (pointer_cache_index != -1
               || writeIntegralDelta(&d_last.data_pointer, record.address >> 3))
           && (allocator_id < 8 || writeSimpleType(record.allocator))
           && (!d_header.native_traces
               || hooks::allocatorKind(record.allocator) == hooks::AllocatorKind::SIMPLE_DEALLOCATOR
               || writeIntegralDelta(&d_last.native_frame_id, record.native_frame_id))
           && (hooks::allocatorKind(record.allocator) == hooks::AllocatorKind::SIMPLE_DEALLOCATOR
               || writeVarint(record.size));
    if (!success) {
        return false;
    }
```

tests/integration/test_api.py
```diff
@@ -105,3 +105,33 @@ def test_aggregated_capture_with_socket_destination():
...
+    with Tracker(result_file, allocation_timestamps=True):
+        allocator.valloc(1234)
+        allocator.free()
+
+    with FileReader(result_file) as reader:
+        records = list(filter_relevant_allocations(reader.get_allocation_records()))
+        assert reader.metadata.has_allocation_timestamps is True
+
+    assert len(records) == 2
+    assert records[0].timestamp_us > 0
```
godlygeek: Why greater than zero? This is a timestamp since the tracking session started, so it could technically be zero since there's no sleeps, right? We might want to put in a `time.sleep(0.01)` before both the `valloc` and the `free` so that both of the assertions can be `>` instead of `>=`

src/memray/_memray/tracking_api.h
```diff
@@ -442,8 +444,10 @@ class Tracker
     const bool d_follow_fork;
     const bool d_trace_python_allocators;
     const bool d_reference_tracking;
+    const bool d_allocation_timestamps;
     linker::SymbolPatcher d_patcher;
     std::unique_ptr<BackgroundThread> d_background_thread;
+    const std::chrono::steady_clock::time_point d_monotonic_start{std::chrono::steady_clock::now()};
```
godlygeek: Is there any way that we can synchronize this to refer to the exact same moment in time as the `start_time` in the metadata? I think the answer is "no", but if we could it would be very nice. As it is, we can _sort of_ say what wall clock time an event happened at, but not exactly - even setting aside clock adjustments due to NTP or whatever, or DST transitions, we'll always be off by some fixed number of microseconds due to the delta between when we get the wall clock timestamp and when we get the monotonic clock timestamp.

Like I said, I don't think there's anything we can do about that, but if there was it'd be cool...

src/memray/_memray/record_writer.cpp
```diff
@@ -181,23 +183,27 @@ createRecordWriter(
         bool native_traces,
         FileFormat file_format,
         bool trace_python_allocators,
-        bool track_object_lifetimes)
+        bool track_object_lifetimes,
+        bool has_allocation_timestamps)
 {
+    has_allocation_timestamps = file_format == FileFormat::ALL_ALLOCATIONS && has_allocation_timestamps;
```
godlygeek: There's no need for this, it's simpler to just do...

```suggestion
```

src/memray/_memray/record_writer.cpp
```diff
@@ -181,23 +183,27 @@ createRecordWriter(
...
+                    has_allocation_timestamps);
         case FileFormat::AGGREGATED_ALLOCATIONS:
             return std::make_unique<AggregatingRecordWriter>(
                     std::move(sink),
                     command_line,
                     native_traces,
                     trace_python_allocators,
-                    track_object_lifetimes);
+                    track_object_lifetimes,
+                    has_allocation_timestamps);
```
godlygeek: and...
```suggestion
                    false);
```

But actually, since we're hardcoding it to false for `AggregatingRecordWriter`, why have it take this parameter at all?

godlygeek (commented):
I'm still looking through this one. I've reviewed the changes to add the allocation timestamp to every allocation record, I haven't yet really reviewed the new report. I've got a few comments on the code from the first commit.

But mostly, I'm a bit confused by why we need the per-allocation timestamp. What is it that this gives us beyond what we can already get from our temporal format? Is it just to get finer granularity rather than sticking us with the default 10ms granularity?

src/memray/_memray/record_writer.cpp
godlygeek: Ah, it's probably less because of the range, and more because avoiding the zigzag encoding means that deltas from 64 to 127 microseconds can be written in one byte instead of two, and that's likely going to be a fairly common window... OK, this is fine. We could always factor out a `writeNonNegativeIntegralDelta` in the future if this becomes something we need in more than one spot.

src/memray/_memray/record_writer.cpp
godlygeek: Meh, nevermind. clang-format wants to reindent that anyway, so it doesn't keep the diff smaller at all (even though most lines only change due to whitespace). Not worth the trouble, I guess.

godlygeek:
Hm. An interesting observation, with my current favorite stress test:

```sh
memray run --trace-python-allocators --allocation-timestamps --no-compress -fo output.bin -m test test_set
```

82.50% of all records have a delta of 0 microseconds from the previous
17.36% of all records have a delta of 1 microsecond from the previous
 0.14% of all records have a delta more than 1 microsecond from the previous
 
The uncompressed capture file is 184,012,800 bytes and contains 42,419,372 allocation records, and deltas of 0 or 1 are varint encoded as a single byte per allocation record, so 19% of the space in the uncompressed capture file (~35 megabytes) is spent encoding "the time hasn't changed since the last record", and an additional 4% of the capture file (~7 megabytes) is spent encoding "the time changed by 1us since the last record".

That seems very inefficient, and makes me think we want a different encoding for timestamps. I'm thinking about creating a new `CLOCK_ADVANCED` record type with a single flag bit. If the flag is 0 it means the clock advanced by 1us, otherwise a varint is emitted saying how many microseconds time advanced by. That means it will take 0 bytes to represent that the time is the same as the previous record and 1 byte to represent that the time has advanced by 1us.

godlygeek:
@pablogsal I'm still not understanding why we need to add the fine grained timestamps for the speedscope report. It seems like the only thing they're used for is to tell when the first time we saw an allocation with a given stack was, and then the superposition of stacks in the speedscope flamegraph winds up ordered left to right by the first time we saw an allocation at each of those stacks. Right?

But... we don't need timestamps for that! The capture file is written in allocation order, so we know whether we saw an allocation with stack X before an allocation with stack Y without needing to know exactly when those allocations occurred. We could get the exact same behavior just replacing the timestamp with a sequence number.

I feel like either I'm badly misreading something or you implemented a really overengineered way to explicitly record a piece of information that we already had implicitly.

Starting with your original commit 31e477b74018c0f5214e4e8d167bbcdf4e0a49d4 (just to make sure I haven't screwed anything up while working on it), I applied this patch:

```diff
diff --git a/src/memray/_memray/tracking_api.cpp b/src/memray/_memray/tracking_api.cpp
index 3dc1692..9b581bf 100644
--- a/src/memray/_memray/tracking_api.cpp
+++ b/src/memray/_memray/tracking_api.cpp
@@ -1495,6 +1495,7 @@ Tracker::getTracker()
 uint64_t
 Tracker::currentTimestampUs() const
 {
+    std::this_thread::sleep_for(std::chrono::microseconds(1));
     return std::chrono::duration_cast<std::chrono::microseconds>(
                    std::chrono::steady_clock::now() - d_monotonic_start)
             .count();
```
That makes it so that we still record the fine grained timestamps, but also guarantees that every allocation will have a unique timestamp, so that the order is deterministic (we never have two allocations rounded to the same timestamp because of our quantum).

Then, I ran:
```sh
memray run -fo output.bin --allocation-timestamps --trace-python-allocators -m test test_set
```
to generate a report with tons of data, and with allocation timestamps in the report.

I generated a speedscope report with:
```sh
memray transform speedscope -f -o memray-speedscope-output.speedscope.json.uniq_micros output.bin
```

Then I applied _this_ patch, which completely ignores the timestamps encoded in the capture file in favor of just using a sequence number:
```diff
diff --git a/src/memray/_memray/record_reader.cpp b/src/memray/_memray/record_reader.cpp
index 64eecd4..b29aa2a 100644
--- a/src/memray/_memray/record_reader.cpp
+++ b/src/memray/_memray/record_reader.cpp
@@ -356,7 +356,7 @@ RecordReader::processAllocationRecord(const AllocationRecord& record)
         d_latest_allocation.native_segment_generation = 0;
     }
     d_latest_allocation.n_allocations = 1;
-    d_latest_allocation.timestamp_us = record.timestamp_us;
+    d_latest_allocation.timestamp_us++;
     return true;
 }
```
and then I regenerated that report again:
```sh
memray transform speedscope -f -o memray-speedscope-output.speedscope.json.uniq_seqnum output.bin
```
and... the two files are byte-for-byte identical.

The only difference the timestamp makes over a sequence number is screwing up the order, by making it arbitrary which order two allocations seen in the same microsecond show up in!

Am I missing something?

pablogsal:
Yep, you’re totally right 😅 My fault. I’m used to profiling formats that rely on timestamps, so I went straight down that path. I’ve removed all the timestamp machinery and the temporal fallback. Snapshot aggregation now keeps a local sequence number and uses it to preserve ordering. I also rebased everything into one clean commit. Thanks for catching it!

godlygeek (approved):
Oh man, this is so much easier to review now 😅 

I could nitpick a bit about some of the naming, but transforms tend to be low maintenance so I'm just not going to sweat it.

Looks good to me. Thanks @pablogsal!

PR #907, opened 2026-04-25
Improve error messages when run command fails to find target

swaroski:
Issue number of the reported bug or feature request: #512
This PR improves the user experience when memray run fails to execute the target script or module. Key changes include:
   - Modified validate_target_file to catch FileNotFoundError and PermissionError, raising a clean MemrayCommandError instead of allowing a raw traceback.
   - Updated _run_tracker to catch ImportError when running with the -m flag, providing a clear error message if the module cannot be found.
   - Changed the "Successfully generated profile results" message logic from a finally block to an else block. This ensures that the success message is only
     printed if the application actually starts and finishes tracking successfully, rather than printing it even when the startup fails.

  Testing performed
   * Added unit tests to tests/unit/test_cli.py specifically targeting missing files and missing modules.
   * Verified that memray run non_existent.py returns exit code 1 and prints a human-readable error.
   * Verified that memray run -m non_existent_module returns exit code 1 and no longer prints the misleading success message.
   * Ran the full test suite via pytest tests to ensure no regressions were introduced to existing tracking logic.

### DIFF
--- a/src/memray/commands/run.py
+++ b/src/memray/commands/run.py
@@ -62,7 +62,13 @@ def _run_tracker(
                     sys.path[0] = os.getcwd()
                 # run_module will replace argv[0] with the script's path
                 sys.argv = ["", *args.script_args]
-                runpy.run_module(args.script, run_name="__main__", alter_sys=True)
+                try:
+                    runpy.run_module(args.script, run_name="__main__", alter_sys=True)
+                except ImportError as e:
+                    if e.name in (None, args.script):
+                        post_run_message = None  # Initial import failed
+                        raise MemrayCommandError(f"memray: {e}", exit_code=1)
+                    raise
             elif args.run_as_cmd:
                 if _should_modify_sys_path():
                     sys.path[0] = ""
@@ -314,6 +320,8 @@ class RunCommand:
                 "Only valid Python files or commands can be executed under memray",
                 exit_code=1,
             )
+        except (FileNotFoundError, PermissionError) as error:
+            raise MemrayCommandError(str(error), exit_code=1)
 
     def run(self, args: argparse.Namespace, parser: argparse.ArgumentParser) -> None:
         if args.no_compress:
--- a/tests/unit/test_cli.py
+++ b/tests/unit/test_cli.py
@@ -919,3 +919,26 @@ class TestTransformSubCommand:
         assert namespace.output == "output.html"
         assert namespace.force is True
         assert namespace.format == "gprof2dot"
+
+
+class TestRunCommandReal:
+    def test_run_non_existent_file(self, capsys):
+        # GIVEN
+        # WHEN
+        exit_code = main(["run", "non_existent_file.py"])
+
+        # THEN
+        assert exit_code == 1
+        captured = capsys.readouterr()
+        assert "No such file or directory: 'non_existent_file.py'" in captured.err
+
+    def test_run_non_existent_module(self, capsys):
+        # GIVEN
+        # WHEN
+        exit_code = main(["run", "-m", "non_existent_module"])
+
+        # THEN
+        assert exit_code == 1
+        captured = capsys.readouterr()
+        assert "No module named non_existent_module" in captured.err
+        assert "[memray] Successfully generated profile results." not in captured.out


godlygeek:
Looks like CI is broken on main... I'll come back to this after fixing that...

src/memray/commands/run.py
```diff
@@ -75,7 +78,9 @@ def _run_tracker(
                     )
                 sys.argv = [args.script, *args.script_args]
                 runpy.run_path(args.script, run_name="__main__")
-        finally:
+        except Exception:
+            raise
```
godlygeek: This has no effect at all, you can just remove these lines.
```suggestion
```
There's no difference between catching then immediately reraising an exception versus not catching it at all.

src/memray/commands/run.py
```diff
@@ -62,7 +62,10 @@ def _run_tracker(
                     sys.path[0] = os.getcwd()
                 # run_module will replace argv[0] with the script's path
                 sys.argv = ["", *args.script_args]
-                runpy.run_module(args.script, run_name="__main__", alter_sys=True)
+                try:
+                    runpy.run_module(args.script, run_name="__main__", alter_sys=True)
+                except ImportError as e:
+                    raise MemrayCommandError(str(e), exit_code=1)
```
godlygeek: Let's only do this if the module that failed to import was in fact `args.script`, and not something else imported by `args.script`. Also, let's add a `memray:` prefix to the message. Both of these match what the interpreter itself does:
```shell-session
$ python -m foo
/bin/python3.14: No module named foo
$ echo 'import bar' >foo.py
$ python -m foo
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/private/tmp/foo.py", line 1, in <module>
    import bar
ModuleNotFoundError: No module named 'bar'
```

```suggestion
                    if e.name == args.script:
                        raise MemrayCommandError(f"memray: {e}", exit_code=1)
                    raise
```

src/memray/commands/run.py
```diff
@@ -75,7 +78,9 @@ def _run_tracker(
                     )
                 sys.argv = [args.script, *args.script_args]
                 runpy.run_path(args.script, run_name="__main__")
-        finally:
+        except Exception:
+            raise
+        else:
```
godlygeek: I don't like this change. It makes it so that we don't print the "[memray] Successfully generated profile results." message not just if we failed to run the user's script, but also if the user's script ended with an exception. We can't tell those two cases apart in general, and so I think we should still print this message. We did successfully generate a memory profile, even if the user's script failed.

src/memray/commands/run.py
```diff
@@ -314,6 +319,8 @@ def validate_target_file(self, args: argparse.Namespace) -> None:
                 "Only valid Python files or commands can be executed under memray",
                 exit_code=1,
             )
+        except (FileNotFoundError, PermissionError) as error:
+            raise MemrayCommandError(str(error), exit_code=1)
```
godlygeek, line 324: I like this change, very good idea.

godlygeek (changes requested):
I'd like some changes here before landing this.

Also, can you please update the `Signed-off-by:` line in your commit message to use your full name? I believe that would be:

```
Signed-off-by: Suyog Bhise <bhise.suyog@gmail.com>
```

You can edit your most recent commit message with `git commit --amend`

godlygeek:
Also, please take a look at https://github.com/bloomberg/memray/blob/main/CONTRIBUTING.md and in particular https://github.com/bloomberg/memray/blob/main/CONTRIBUTING.md#linting-your-code - it would be helpful to make sure that your proposed changes pass the style lints before pushing them.

swaroski:
@godlygeek addressed review feedback: filtered ImportError to
  args.script with memray: prefix, removed dead except/raise, restored finally so success message still prints, fixed
  Signed-off-by. Ready for another look.

src/memray/commands/run.py
```diff
@@ -62,7 +62,12 @@ def _run_tracker(
                     sys.path[0] = os.getcwd()
                 # run_module will replace argv[0] with the script's path
                 sys.argv = ["", *args.script_args]
-                runpy.run_module(args.script, run_name="__main__", alter_sys=True)
+                try:
+                    runpy.run_module(args.script, run_name="__main__", alter_sys=True)
+                except ImportError as e:
+                    if e.name == args.script:
```
godlygeek: Looks like `runpy.run_module` sets `e.name` to `None` if it fails to find the module. I think that might be a CPython bug? Either way, it's easy to work around.

Also, while we're at it, let's set `post_run_message = None` if we hit this path - then we won't print the instructions for generating the report if we failed to import the top level module.


```suggestion
                    if e.name in (None, args.script):
                        post_run_message = None  # Initial import failed
```

godlygeek: I've opened https://github.com/python/cpython/issues/149117 to report this `runpy.run_module` behavior to the CPython folks and see if they agree it's a bug.

godlygeek (approved):
LGTM now, let's see if the test suite passes...

PR #908, opened 2026-04-27
Fix greenlet integration with greenlet 3.4.0

godlygeek:
Our attempts to handle ancient versions of greenlet are causing us to try to import the `settrace` function from `greenlet` rather than from `greenlet._greenlet`, which is failing because we're trying to import it before greenlet's own `__init__.py` has imported it.

Drop support for greenlet versions older than 1.0, which is more than 5 years out of date at this point, to fix this issue.

### DIFF
--- a/news/908.bugfix.rst
+++ b/news/908.bugfix.rst
@@ -1,0 +1,2 @@
+Fix integration with the ``greenlet`` package when greenlet 3.4.0 is installed
+and when tracing is started before the greenlet module is imported.
--- a/news/908.removal.rst
+++ b/news/908.removal.rst
@@ -1,0 +1,1 @@
+Drop integration with 0.x versions of the ``greenlet`` module (which are from 2020 or earlier).
--- a/src/memray/_memray/hooks.cpp
+++ b/src/memray/_memray/hooks.cpp
@@ -380,9 +380,7 @@ dlopen(const char* filename, int flag) noexcept
     }
     if (ret) {
         tracking_api::Tracker::invalidate_module_cache();
-        if (filename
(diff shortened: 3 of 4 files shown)


godlygeek:
Specifically, the full failure mode here is:

- The user starts Memray tracking
- The user does `import greenlet`
- The `greenlet` module does `from ._greenlet import ...`
- The interpreter dlopen's the `_greenlet` shared library
- Memray's dlopen hook fires and sets a flag to tell our Python profile function to try to install our greenlet trace function the next time a Python instruction is run
- The `greenlet._greenlet` of greenlet 3.4.0 does a `PyImport_ImportModule("atexit")` from it's initialization function
- Because `atexit` is written in Python, our Python profile function runs
- We try to access `sys.modules["greenlet._greenlet"]` but fail because it's not yet added to sys.modules
- We fall back to using `sys.modules["greenlet"]` for greenlet 0.x compatibility
- We try to access `greenlet.settrace`, which fails because it hasn't yet been imported by `greenlet` from `greenlet._greenlet`

The `PyImport_ImportModule("atexit")` was reverted in greenlet 3.5.0, but removing the fallback to import `greenlet.settrace` when `greenlet._greenlet.settrace` can't be imported fixes the issue for greenlet 3.4.0 and hardens us against the possibility of the greenlet module running other Python code in the future, or allowing a context switch to another thread (what we were doing was probably broken in free-threading builds as well).

lkollar (approved):
Would it be worth mentioning that in the news fragment that this drops greenlet support before 1.0? LGTM otherwise.

godlygeek:
> Would it be worth mentioning that in the news fragment that this drops greenlet support before 1.0? LGTM otherwise.

Good point, done.

Issue #747, opened 2025-04-11
Add more logging around error conditions


pablogsal:
Check https://github.com/bloomberg/memray/issues/731 for some context. The objective is to know what is going on when different low-level calls fail from the logs. We should be looking at every "error path" around and check if it makes sense to add a log line there.

swaroski:
Follow-up to #911 — a few additional silent / bare error paths I'd like to address in a small follow-up PR:

1. **`FileSink::seek` — `munmap` failure is silent.** When sliding the mmap window forward, a failed `munmap` returns `false` with no log line, so the user only sees the eventual generic write failure.
2. **`SocketSink` — `listen()` failure has no errno.** `throw IoError{"Encountered error in listen call"}` drops `errno`. The neighboring `socket`/`setsockopt`/`bind`/`accept` paths all log + include `strerror`; `listen` is the odd one out.
3. **`BackgroundThread` — `/proc/self/status` open failure has no errno.** `throw IoError{"Failed to open /proc/self/status"}` (tracking_api.cpp). EACCES vs ENOENT vs EMFILE all surface identically.
4. **`Tracker` ctor — `Failed to write output header` exception text has no errno.** #911 added detailed `LOG(ERROR)` inside `FileSink`, which is great, but the Python-side `RuntimeError` message users see in tracebacks (e.g. #731) is still bare. Appending `strerror(errno)` to the throw makes the traceback self-contained for users who don't have `MEMRAY_LOG_LEVEL` cranked up.

Plan: same shape as #911 — log + include `strerror` at each site. Happy to adjust scope if any of these aren't worth it.

godlygeek:
I'd rather not add logs for something unless we can think of a way that it could plausibly happen. Otherwise it's just more code for us to maintain with no chance that any user ever benefits from it.

Skimming through these...

1. I can't think of any way `munmap` could fail other than a bug in our code or memory trashing in the user's application. I don't think we should bother adding specific logs for this one.
2. This one could happen if another process is already listening on the port. Appending strerror to the listen() failure message does seem worthwhile.
3. Users could see this one if they're a mount namespace without /proc mounted, or possibly if they start the process as root and then drop privileges to a regular user uid. Or I suppose if they hit the limit on the number of open files. Since it is reachable, I'd be OK with adding strerror to it.
[...]

Issue #731, opened 2025-03-31
Failed to write output header


jaredhaertel:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

I have a file `hello.py` that just contains `print("hello world")`. I get the following error
```
(test) jhaertel@orca code % memray run hello.py
Writing profile results into memray-hello.py.75618.bin
Traceback (most recent call last):
  File "/u/jhaertel/.conda/envs/test/bin/memray", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/u/jhaertel/.conda/envs/test/lib/python3.13/site-packages/memray/commands/__init__.py", line 138, in main
    arg_values.entrypoint(arg_values, parser)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/u/jhaertel/.conda/envs/test/lib/python3.13/site-packages/memray/commands/run.py", line 339, in run
    _run_with_file_output(args)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/u/jhaertel/.conda/envs/test/lib/python3.13/site-packages/memray/commands/run.py", line 188, in _run_with_file_output
    _run_tracker(
    ~~~~~~~~~~~~^
        destination=destination,
        ^^^^^^^^^^^^^^^^^^^^^^^^
        args=args,
        ^^^^^^^^^^
        post_run_message=example_report_generation_message,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/u/jhaertel/.conda/envs/test/lib/python3.13/site-packages/memray/commands/run.py", line 57, in _run_tracker
    with tracker:
         ^^^^^^^
  File "src/memray/_memray.pyx", line 697, in memray._memray.Tracker.__enter__
  File "src/memray/_memray.pyx", line 720, in memray._memray.Tracker.__enter__
RuntimeError: Failed to write output header
```
This occurs with any python file I attempt.

### Expected Behavior

I expected it to run and create a binary file.

### Steps To Reproduce

1. In this virtual environment, I just have python and memray.
2. Run memray on `hello.py` that just contains `print("hello world")`.

### Memray Version

1.16.0

### Python Version

3.12, 3.13

### Operating System

macOS

### Anything else?

This is on MacOS 12.6.2 on a Intel i7-6700K CPU.

[...]

PR #911, opened 2026-04-27
Add error logging
  around FileSink and header serialization failures

swaroski:
Refs #747, #731

  ## Summary
  - `FileSink::seek` and `FileSink::grow` now log the failing syscall (`mmap`, `munmap`, `lseek`, `posix_fallocate`),
  the errno string, and a hint to try writing the capture file to a different location such as `/tmp`.
  - `RecordWriter::writeHeaderCommon` now logs when header serialization fails, pointing readers at the preceding
  sink-level log entries for the underlying cause.
  - Previously, filesystems that do not support shared writable `mmap` or `posix_fallocate` (issue #731) surfaced only
  as a generic `RuntimeError: Failed to write output header` with no diagnostic information.

  ## Test plan
  - [x] Extension rebuilds cleanly.
  - [x] `tests/unit/test_tracker.py` passes against the rebuilt extension.
  - [ ] CI

### DIFF
--- a/src/memray/_memray/sink.cpp
+++ b/src/memray/_memray/sink.cpp
@@ -3,6 +3,7 @@
 
 #include <cerrno>
 #include <cstdio>
+#include <sstream>
 
 #include <arpa/inet.h>
 #include <fcntl.h>
@@ -13,6 +14,7 @@
 #include <utility>
 
 #include "exceptions.h"
+#include "logging.h"
 #include "lz4_stream.h"
 #include "sink.h"
 
@@ -134,6 +136,14 @@ FileSink::seek(off_t offset, int whence)
     //       though not to write beyond the end.
     d_buffer = static_cast<char*>(mmap(d_buffer, BUFFER_SIZE, PROT_WRITE, MAP_SHARED, d_fd, offset));
     if (d_buffer == MAP_FAILED) {
+        std::ostringstream msg;
+        msg << "Failed to mmap " << BUFFER_SIZE << " bytes of output file " << d_filename
+            << " at offset " << offset << ": " << strerror(errno);
+        if (offset == 0) {
+            msg << ". The destination filesystem may not support shared writable mmap;"
+                   " try writing the capture file to a different location (e.g. /tmp).";
+        }
+        LOG(ERROR) << msg.str();
         d_buffer = nullptr;
         return false;
     }
@@ -163,6 +173,19 @@ FileSink::grow(size_t needed)
     } while (rc == EINTR);
 
     if (rc != 0) {
+        std::ostringstream msg;
+        msg << "Failed to grow output file " << d_filename << " by " << delta
+            << " bytes: " << strerror(rc);
+        if (d_fileSize == 0) {
+#ifdef __APPLE__
+            msg << ". The destination filesystem may not support F_PREALLOCATE;"
+                   " try writing the capture file to a different location (e.g. /tmp).";
+#else
+            msg << ". The destination filesystem may not support posix_fallocate;"
+                   " try writing the capture file to a different location (e.g. /tmp).";
+#endif
+        }
+        LOG(ERROR) << msg.str();
         errno = rc;
         return false;
     }
--- a/news/747.feature.rst
+++ b/news/747.feature.rst
@@ -1,0 +1,1 @@
+Log the underlying error (and a hint to try a different output location) when ``mmap`` or ``posix_fallocate`` fails inside ``FileSink``, and when the capture file header can't be serialized. Previously these failures surfaced only as a generic ``Failed to write output header`` ``RuntimeError`` with no information about the cause.


src/memray/_memray/sink.cpp
```diff
@@ -111,6 +112,7 @@ FileSink::seek(off_t offset, int whence)
     // when we grow the file, and don't move it when we write, so users can't
     // possibly know what the offset is.
     if (whence != SEEK_SET && whence != SEEK_END) {
+        LOG(ERROR) << "FileSink::seek called with unsupported whence " << whence;
```
godlygeek: This case would be a programming error if it ever happened. We don't need to give a nice user-facing error message for something that can only happen if we screwed up (or the user trashed program memory).

src/memray/_memray/sink.cpp
```diff
@@ -121,19 +123,27 @@ FileSink::seek(off_t offset, int whence)
     }
 
     if (offset < 0) {
+        LOG(ERROR) << "FileSink::seek got a negative absolute offset for fd " << d_fd
+                   << ": " << strerror(errno);
```
godlygeek: Same for this one.

src/memray/_memray/sink.cpp
```diff
@@ -121,19 +123,27 @@ FileSink::seek(off_t offset, int whence)
...
         errno = EINVAL;
         return false;
     }
 
     // Free our existing buffer, if any
     if (d_buffer && 0 != munmap(d_buffer, BUFFER_SIZE)) {
+        LOG(ERROR) << "Failed to unmap previous output buffer: " << strerror(errno);
```
godlygeek: And this one.

src/memray/_memray/sink.cpp
```diff
@@ -121,19 +123,27 @@ FileSink::seek(off_t offset, int whence)
...
         return false;
     }
 
     // Note: It is OK to map beyond the end of the file,
     //       though not to write beyond the end.
     d_buffer = static_cast<char*>(mmap(d_buffer, BUFFER_SIZE, PROT_WRITE, MAP_SHARED, d_fd, offset));
     if (d_buffer == MAP_FAILED) {
+        LOG(ERROR) << "Failed to mmap output file (fd " << d_fd << ", offset " << offset
+                   << ", length " << BUFFER_SIZE
+                   << "): " << strerror(errno)
```
godlygeek: Logging the fd isn't useful to the end user. We'd want to do something like:
```suggestion
        LOG(ERROR) << "Failed to mmap " << BUFFER_SIZE << " bytes of output file "
                   << d_filename << " at offset " << offset << ": " << strerror(errno)
```

src/memray/_memray/sink.cpp
```diff
@@ -121,19 +123,27 @@ FileSink::seek(off_t offset, int whence)
...
+                   << ". The destination filesystem may not support shared writable mmap;"
+                   << " try writing the capture file to a different location (e.g. /tmp).";
```
godlygeek: This only makes sense to say if the offset is 0. For a non-zero offset, we successfully created at least one shared writable mmap previously, so it must not be a lack of filesystem support that has gone wrong.

src/memray/_memray/sink.cpp
```diff
@@ -163,6 +173,10 @@ FileSink::grow(size_t needed)
     } while (rc == EINTR);
 
     if (rc != 0) {
+        LOG(ERROR) << "Failed to grow output file (fd " << d_fd << ", offset " << d_fileSize
+                   << ", delta " << delta << "): " << strerror(rc)
```
godlygeek:
```suggestion
        LOG(ERROR) << "Failed to grow output file " << d_filename << " by "
                   << delta << " bytes: " << strerror(rc)
```

src/memray/_memray/sink.cpp
```diff
@@ -163,6 +173,10 @@ FileSink::grow(size_t needed)
...
+                   << ". The destination filesystem may not support posix_fallocate;"
```
godlygeek: As above, this would only make sense to say if `d_fileSize` is 0. Otherwise we've called `posix_fallocate` successfully at least once. Also, note that we've got our own emulation of `posix_fallocate` for macOS... we probably shouldn't mention `posix_fallocate` in an error message at all on macOS. We should probably mention `F_PREALLOCATE` instead.

src/memray/_memray/record_writer.cpp
```diff
@@ -491,6 +492,9 @@ RecordWriter::writeHeaderCommon(const HeaderRecord& header)
         or !writeSimpleType(header.python_allocator) or !writeSimpleType(header.trace_python_allocators)
         or !writeSimpleType(header.track_object_lifetimes))
     {
+        LOG(ERROR) << "Failed to serialize capture file header. The destination sink"
+                   << " rejected one of the header fields; see preceding log entries"
+                   << " (e.g. mmap or posix_fallocate failures) for the underlying cause.";
```
godlygeek: I don't think this log adds any value. We're already going to log an error message that says "Failed to write output header" due to:
```cpp
    if (!d_writer->writeHeader(false)) {
        throw IoError{"Failed to write output header"};
    }
```
I think that's informative enough and this doesn't add any real value over the messages logged by the sink and the one logged by the tracker.

godlygeek (changes requested):

swaroski:
@godlygeek addressed: dropped programming-error LOGs
  (whence/negative-offset/munmap), reworded mmap+grow logs to use d_filename and bytes wording, gated filesystem-support
   hint on first-call only, added macOS F_PREALLOCATE wording, removed redundant header-serialize LOG, added
  Signed-off-by.

src/memray/_memray/sink.cpp
```diff
@@ -163,6 +173,19 @@ FileSink::grow(size_t needed)
     } while (rc == EINTR);
 
     if (rc != 0) {
+        auto msg = std::ostringstream{};
```
godlygeek: Simpler and easier to read to do:
```suggestion
        std::ostringstream msg;
```

src/memray/_memray/sink.cpp
```diff
@@ -134,6 +136,14 @@ FileSink::seek(off_t offset, int whence)
     //       though not to write beyond the end.
     d_buffer = static_cast<char*>(mmap(d_buffer, BUFFER_SIZE, PROT_WRITE, MAP_SHARED, d_fd, offset));
     if (d_buffer == MAP_FAILED) {
+        auto msg = std::ostringstream{};
```
godlygeek: Simpler and easier to read to do:
```suggestion
        std::ostringstream msg;
```

godlygeek (approved):
LGTM

PR #912, opened 2026-04-29
Intercept C23 free_sized() and free_aligned_sized()

### LINKED ISSUE #725: Intercept new `free_sized` and `free_aligned_sized` functions

swaroski:
Closes: #725

  Adds hooks for C23 sized deallocators on glibc Linux. Both record as FREE per spec. Weak forward decls + runtime symbol resolution make manylinux
   wheels (built on old glibc) work on newer glibc systems.

  ## Test plan
  - [x] New test `test_free_sized_extension` passes on glibc 2.42+, skips on older
  - [x] Existing extension tests pass (no regression)
  - [x] pre-commit clean

### DIFF
--- a/news/725.feature.rst
+++ b/news/725.feature.rst
@@ -1,0 +1,1 @@
+Intercept the C23 ``free_sized()`` and ``free_aligned_sized()`` deallocators on Linux. Calls to either are now recorded as ``FREE`` allocations, matching the behavior of plain ``free()``.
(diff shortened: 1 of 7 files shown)


swaroski:
Fixed preprocessor guard for non-glibc platforms (Alpine/musl/macOS). `__GLIBC_PREREQ` undefined there caused
  tokenization error even with short-circuit.

godlygeek:
I'm going to close this in favor of resurrecting #817, which came first and is closer to what I want the final state to be. Thanks for the heads up that glibc released a version that finally supports these deallocators, though!

PR #914, opened 2026-05-02
Append strerror to listen, /proc/self/status, and header-write errors

swaroski:
## Summary

Follow-up to #911. Append `strerror(errno)` to three remaining bare error sites so the cause is visible in tracebacks without raising `MEMRAY_LOG_LEVEL`:

- `SocketSink::open` — `listen()` failure (e.g. port already bound).
- `Tracker::BackgroundThread` — `/proc/self/status` open failure (mount namespace without `/proc`, post-privilege-drop, `EMFILE`).
- `Tracker` ctor — `Failed to write output header` `RuntimeError` (e.g. `ENOSPC`, `EFBIG`, `EPIPE`).

Per maintainer guidance, the `FileSink::seek` `munmap` path is left unchanged.

## Test plan

- [ ] Builds clean with `pip install -e . --no-build-isolation`
- [ ] `pytest tests/` passes
- [ ] Manual: bind port, run `SocketSink` against same port — `IoError` contains `Address already in use`
- [ ] Manual: write capture to tiny tmpfs — `RuntimeError` traceback contains `No space left on device`


news/747.feature.rst
```diff
@@ -1 +1 @@
-Log the underlying error (and a hint to try a different output location) when ``mmap`` or ``posix_fallocate`` fails inside ``FileSink``, and when the capture file header can't be serialized. Previously these failures surfaced only as a generic ``Failed to write output header`` ``RuntimeError`` with no information about the cause.
+Improve error reporting for capture-file failures: log the underlying ``errno`` (and a hint to try a different output location) when ``mmap`` or ``posix_fallocate`` fails inside ``FileSink``; append ``strerror`` to the ``listen`` failure in ``SocketSink``, to the ``/proc/self/status`` open failure in the background thread, and to the ``Failed to write output header`` ``RuntimeError`` so the cause is visible in Python tracebacks even without ``MEMRAY_LOG_LEVEL`` raised.
```
godlygeek: This seems much too specific for a changelog entry. Let's just say something like

```suggestion
Improve error reporting after a failure to record a memory capture by logging the reported error string and including it in the exception message.
```

src/memray/_memray/tracking_api.cpp
```diff
@@ -876,7 +878,9 @@ Tracker::Tracker(
 
     d_writer->setMainTidAndSkippedFrames(thread_id(), computeMainTidSkip());
     if (!d_writer->writeHeader(false)) {
-        throw IoError{"Failed to write output header"};
+        // Save errno before string ops below can clobber it.
+        int saved_errno = errno;
+        throw IoError{"Failed to write output header: " + std::string(strerror(saved_errno))};
     }
```
godlygeek: Saving it isn't necessary, because the `strerror` happens before any of the `std::string` operations - it's specified to be the first operation that will happen on this line.
```suggestion
        throw IoError{"Failed to write output header: " + std::string(strerror(errno))};
}
```

src/memray/_memray/tracking_api.cpp
```diff
@@ -944,7 +948,8 @@ Tracker::BackgroundThread::BackgroundThread(
 #ifdef __linux__
     d_procs_statm.open("/proc/self/status");
     if (!d_procs_statm) {
-        throw IoError{"Failed to open /proc/self/status"};
+        int saved_errno = errno;
+        throw IoError{"Failed to open /proc/self/status: " + std::string(strerror(saved_errno))};
```
godlygeek: As above:
```suggestion
        throw IoError{"Failed to open /proc/self/status: " + std::string(strerror(errno))};
```

src/memray/_memray/sink.cpp
```diff
@@ -379,8 +379,10 @@ SocketSink::open()
     }
 
     if (listen(sockfd, 1) == -1) {
+        int saved_errno = errno;
         ::close(sockfd);
-        throw IoError{"Encountered error in listen call"};
+        LOG(ERROR) << "Encountered error in 'listen' call: " << strerror(saved_errno);
+        throw IoError{"Failed to listen on socket: " + std::string(strerror(saved_errno))};
```
godlygeek: I think it's slightly cleaner to do:
```suggestion
        std::string reason = strerror(errno);
        ::close(sockfd);
        LOG(ERROR) << "Encountered error in 'listen' call: " << reason;
        throw IoError{"Failed to listen on socket: " + reason};
```
That way it's only converted to a string once instead of twice, and it's a bit less verbose as well.

src/memray/_memray/tracking_api.cpp
```diff
@@ -2,6 +2,8 @@
 #include <Python.h>
 
 #include <cassert>
+#include <cerrno>
+#include <cstring>
```
godlygeek: I think you included this for `strerror`, but `cstring` doesn't provide `strerror`, it provides `std::strerror`, which isn't what you're calling.

`strerror` comes from `stdio.h`.


```suggestion
#include <stdio.h>
```

godlygeek (approved):
LGTM, thanks!

PR #926, opened 2026-05-18
Removing requirements-*.txt files in favor of dep groups

### LINKED ISSUE #919: Switch from requirements-*.txt files to to dependency groups
This project has been managing its development dependencies using `requirements-*.txt` files in the repository root directory. Switch to using [dependency groups](https://packaging.python.org/en/latest/specifications/dependency-groups/) instead.

grantlouisherman:
*Issue number of the reported bug or feature request: #919

**Describe your changes**
I removed requirements-*.txt in favor or using dep groups

**Testing performed**
Automated testing


pyproject.toml
```diff
@@ -9,6 +9,11 @@ requires = [
 
 build-backend = 'setuptools.build_meta'
 
+[dependency-groups]
+test=['Cython', 'coverage[toml]', "greenlet; python_version < '3.14'", 'pytest', 'pytest-cov', 'ipython', 'setuptools', 'pkgconfig', 'pytest-textual-snapshot', 'textual >= 0.43, != 0.65.2, != 0.66', 'packaging']
+docs=['mypy', 'bump2version','towncrier', 'pre-commit', {include-group = "test"}]
```
grantlouisherman: In the original there was a requirement from extra to docs, but they essentially have the same dependencies so I just made them the same. Obviously this is no DRY but I dont think there is a good way to do variables in pyproject tomls?

godlygeek: I think you misread something 😅 

This should be:

```suggestion
docs=['IPython', 'sphinx','sphinx-autobuild', 'sphinx-argparse', 'furo']
```

pyproject.toml
```diff
@@ -9,6 +9,11 @@ requires = [
...
+extra=['mypy', 'bump2version','towncrier', 'pre-commit', {include-group = "test"]
```
godlygeek: While we're at it, let's fix:
```suggestion
extra=['mypy', 'bump2version', 'towncrier', 'pre-commit', {include-group = "docs"]
```

pyproject.toml
grantlouisherman: Yeah wow Im so sorry

grantlouisherman:
@godlygeek should be good now .... I hope

Dockerfile
```diff
@@ -39,16 +39,15 @@ ENV PATH="${VIRTUAL_ENV}/bin:/usr/lib/ccache:${PATH}" \
...
-    -r /tmp/requirements-docs.txt \
-    cython \
-    pkgconfig \
-    setuptools \
-    wheel
+COPY pyproject.toml /tmp/pyproject.toml
+
+RUN $PYTHON -m pip install --upgrade pip \
+    && $PYTHON -m pip install \
+        --group /tmp/pyproject.toml:extra \
```
godlygeek, line 47: Looks like this incorrectly dropped the test dependencies (`extra` pulls in `docs`, but neither `extra` nor `docs` pulls in `test`)
```suggestion
        --group /tmp/pyproject.toml:test \
        --group /tmp/pyproject.toml:extra \
```

README.md
```diff
@@ -106,7 +106,8 @@ cd memray
 python3 -m venv ../memray-env/  # just an example, put this wherever you want
 source ../memray-env/bin/activate
 python3 -m pip install --upgrade pip
-python3 -m pip install -e . -r requirements-test.txt -r requirements-extra.txt
+python3 -m pip install --group extra
+python3 -m pip install -e .
```
godlygeek: This can be done in one call, and it's better to do it that way because it guarantees that the dependencies of the project itself and of the extra requirements are resolved together and so definitely mutually compatible.

Also, we still need to install the test dependencies here.

```suggestion
python3 -m pip install -e . --group test --group extra
```

godlygeek (approved):
Thanks for the contribution, @grantlouisherman !

PR #927, opened 2026-05-18
Support Python 3.15

### LINKED ISSUE #920: Support Python 3.15
Test with Python 3.15, fix any incompatibilities, and update our trove classifiers to declare Python 3.15 support.

ikrommyd:
Closes #920 

**Describe your changes**
We switch to `enum class` instead of `enum` because cython uses `IntFlag` for `enum` and that messes up negative values in 3.15.
We also remove the `".llvm." in symbol` check  in `hybrid_stack_trace` as that no longer appears to be the case.
We add 3.15 in CI and we have to use the `cibuildwheel` release candidate to get 3.15.

**Testing performed**
Tested locally with 3.14 and 3.15 and the tests pass. We'll see what the CI says

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -181,6 +181,10 @@ jobs:
           rmdir dist/{sdist,*-wheels}
           rm -r dist/tests
           ls -R dist
+      - name: Avoid publishing Python 3.15 wheels
+        run: |
+          rm -f dist/*cp315*
+          ls -R dist
       - uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b # release/v1
         with:
           skip-existing: true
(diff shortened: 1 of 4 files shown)


da-woods:
I do intend to fix the negative enum issue in Cython before 3.15 - I just haven't decided exactly how (or done it) yet.

The other thing you could do rather than changing the numbers is change the enum to `cpdef enum class`. This creates a c++ enum class instead (which is probably fine here) and creates an `IntEnum` rather than an `IntFlag` which should avoid the issue.

ikrommyd:
@da-woods I tried switching to enum class and tests are fine locally. We'll see what CI says.

ikrommyd:
Rebased to main. Had to fix conflicts in workflows.

godlygeek (approved):
LGTM, thanks Iason!

ikrommyd:
You’re welcome! And thanks for fixing the merge conflicts. I was gonna fix this at the airport today but you got there first.

Issue #765, opened 2025-05-19
Update `memray stats` to show the top 5 non-stdlib modules with the most allocations


godlygeek:
Update the `memray stats` reporter to associate each allocation with a module (the module of the closest Python frame to the allocation that's not part of the standard library). Print some statistics about which modules are responsible for the most allocations.

ivonastojanovic:
I'll give it a go :)

godlygeek:
Closed by #938

Issue #766, opened 2025-05-19
Update `memray flamegraph` to allow filtering by module


godlygeek:
Build on top of the work done in #765 to allow filtering flame graphs by module, so that only stacks containing frames for the chosen module are shown.

grantlouisherman:
Hey @godlygeek  could I take a look at this?

sarahmonod:
All yours @grantlouisherman ! Let us know if you can still do this even though the PyCon US 2026 sprints are over.

grantlouisherman:
@sarahmonod how do the sprints relate to this? Just my ignorance

PR #938, opened 2026-05-19
Add per-module allocation statistics to memray stats

ivonastojanovic:
*Issue number of the reported bug or feature request: #765*

**Describe your changes**
Display the top N modules responsible for the most allocations in the memray stats reporter. For each allocation, walk the call stack to locate the nearest non-stdlib Python frame and extract its top-level module name. Aggregate allocation counts and allocated bytes by module, and include the top N results in both terminal and JSON output.

**Testing performed**
Unit tests + running `memray stats` on the example below:
```
"""Test script for memray stats module attribution."""

import json

import psutil
import rich.text
import yaml


def allocate_with_yaml():
    docs = [yaml.dump({"key": i, "nested": {"values": list(range(20))}}) for i in range(2000)]
    return [yaml.safe_load(d) for d in docs]


def allocate_with_psutil():
    procs = []
    for _ in range(100):
        procs.append(psutil.Process().memory_info())
        procs.append(psutil.Process().cpu_times())
    return procs


def allocate_with_rich():
    texts = [rich.text.Text(f"Hello world {i} " * 50) for i in range(3000)]
    return texts


def allocate_with_json():
    data = [json.dumps({"key": i, "values": list(range(50))}) for i in range(5000)]
    return [json.loads(d) for d in data]


def main():
    a = allocate_with_yaml()
    b = allocate_with_psutil()
    c = allocate_with_rich()
    d = allocate_with_json()
    print(f"Done: {len(a)} yaml, {len(b)} psutil, {len(c)} rich, {len(d)} json")


if __name__ == "__main__":
    main()

```

Output:
```
...

🥇 Top 5 allocating modules (by size):
        - yaml: 225.162MB total | 690815 allocations
        - test_modules: 37.624MB total | 43279 allocations
        - rich: 2.642MB total | 3000 allocations
        - psutil: 1.016MB total | 780 allocations
        - src: 352.635kB total | 199 allocations

🥇 Top 5 allocating modules (by number of allocations):
        - yaml: 225.162MB total | 690815 allocations
        - test_modules: 37.624MB total | 43279 allocations
        - rich: 2.642MB total | 3000 allocations
        - psutil: 1.016MB total | 780 allocations
        - src: 352.635kB total | 199 allocations
```

JSON Output:
```
...
],
"top_allocations_by_module": [
    {
      "module": "yaml",
      "num_allocations": 690815,
      "total_bytes": 225161828
    },
    {
      "module": "test_modules",
      "num_allocations": 43279,
      "total_bytes": 37623961
    },
    {
      "module": "rich",
      "num_allocations": 3000,
      "total_bytes": 2641500
    },
    {
      "module": "psutil",
      "num_allocations": 780,
      "total_bytes": 1016183
    },
    {
      "module": "src",
      "num_allocations": 199,
      "total_bytes": 352635
    }
  ],
  "top_allocations_by_module_by_count": [
    {
      "module": "yaml",
      "num_allocations": 690815,
      "total_bytes": 225161828
    },
    {
      "module": "test_modules",
      "num_allocations": 43279,
      "total_bytes": 37623961
    },
    {
      "module": "rich",
      "num_allocations": 3000,
      "total_bytes": 2641500
    },
    {
      "module": "psutil",
      "num_allocations": 780,
      "total_bytes": 1016183
    },
    {
      "module": "src",
      "num_allocations": 199,
      "total_bytes": 352635
    }
  ],
  "metadata": {
  ...
```

**Additional context**

### DIFF
--- a/src/memray/_memray.pyx
+++ b/src/memray/_memray.pyx
@@ -1493,6 +1493,8 @@ def compute_statistics(
     report_progress=False,
     num_largest=5,
 ):
+    from memray.reporters.module_tools import ModuleResolver
+
     cdef shared_ptr[RecordReader] reader_sp = make_shared[RecordReader](
         unique_ptr[FileSource](new FileSource(file_name))
     )
@@ -1512,15 +1514,28 @@ def compute_statistics(
         total=total,
         report_progress=report_progress,
     )
+
+    # Aggregate allocation counts and bytes per distinct call stack using a C++
+    # integer map keyed by frame_index. Many allocations share the same stack,
+    # so this avoids duplicating work.
+    cdef unordered_map[size_t, pair[size_t, size_t]] stats_by_stack
+    cdef pair[size_t, size_t]* stack_entry
+
     with progress_indicator:
         while True:
             PyErr_CheckSignals()
             ret = reader.nextRecord()
             if ret == RecordResult.RecordResultAllocationRecord:
+                allocation = reader.getLatestAllocation()
                 aggregator.addAllocation(
-                    reader.getLatestAllocation(),
-                    reader.getLatestPythonLocationId(reader.getLatestAllocation()),
+                    allocation,
+                    reader.getLatestPythonLocationId(allocation),
                 )
+
+                if not isDeallocator(allocation.allocator):
+                    stack_entry = &stats_by_stack[allocation.frame_index]
+                    stack_entry.first += 1
+                    stack_entry.second += allocation.size
                 progress_indicator.update(1)
             elif ret == RecordResult.RecordResultMemoryRecord:
                 pass
@@ -1554,6 +1569,34 @@ def compute_statistics(
         for count_and_loc in aggregator.topLocationsByCount(num_largest)
     ]
 
+    # Resolve each distinct call stack to a module name, adding its aggregated
+    # counts into the per-module totals.
+    module_resolver = ModuleResolver()
+    module_stats = {}  # module -> [num_allocations, total_bytes]
+    cdef pair[size_t, pair[size_t, size_t]] stack_stats
+    for stack_stats in stats_by_stack:
+        stack = reader.Py_GetStackFrame(stack_stats.first)
+        module_name = module_resolver.get_module_for_stack(stack)
+        entry = module_stats.setdefault(module_name, [0, 0])
+        entry[0] += stack_stats.second.first
+        entry[1] += stack_stats.second.second
+
+    top_modules_by_count = sorted(
+        module_stats.items(), key=lambda x: x[1][0], reverse=True
+    )[:num_largest]
+    top_modules_by_allocation_count = [
+        (name, count, total_bytes)
+        for name, (count, total_bytes) in top_modules_by_count
+    ]
+
+    top_modules_by_size = sorted(
+        module_stats.items(), key=lambda x: x[1][1], reverse=True
+    )[:num_largest]
+    top_modules_by_allocation_size = [
+        (name, count, total_bytes)
+        for name, (count, total_bytes) in top_modules_by_size
+    ]
+
     # And we're done!
     cdef uint64_t peak_memory = aggregator.peakBytesAllocated()
     return Stats(
@@ -1565,6 +1608,8 @@ def compute_statistics(
         allocation_count_by_allocator=allocation_count_by_allocator,
         top_locations_by_size=top_locations_by_size,
         top_locations_by_count=top_locations_by_count,
+        top_modules_by_allocation_size=top_modules_by_allocation_size,
+        top_modules_by_allocation_count=top_modules_by_allocation_count,
     )
 
 
--- a/src/memray/reporters/module_tools.py
+++ b/src/memray/reporters/module_tools.py
@@ -1,0 +1,134 @@
+"""Utilities for extracting module names from file paths."""
+
+import site
+import sys
+import sysconfig
+from pathlib import Path
+from typing import Dict
+from typing import Iterable
+from typing import List
+from typing import Optional
+from typing import Tuple
+
+from typing_extensions import TypedDict
+
+
+class PathInfo(TypedDict):
+    stdlib: Optional[Path]
+    site_packages: List[Path]
+    sys_path: List[Path]
+
+
+def get_python_path_info() -> PathInfo:
+    """Get information about Python's search paths.
+
+    Returns:
+        dict: Dictionary containing stdlib path, site-packages paths, and sys.path entries.
+    """
+    libdest = sysconfig.get_config_var("LIBDEST")
+    stdlib: Optional[Path] = Path(libdest) if libdest else None
+
+    # Get site-packages directories
+    site_packages: List[Path] = [Path(p) for p in site.getsitepackages()]
+
+    # Get user site-packages
+    user_site = site.getusersitepackages()
+    if Path(user_site).exists():
+        site_packages.append(Path(user_site))
+
+    return {
+        "stdlib": stdlib,
+        "site_packages": site_packages,
+        "sys_path": [Path(p) for p in sys.path if p],
+    }
+
+
+def _is_relative_to(path: Path, base: Path) -> bool:
+    try:
+        path.relative_to(base)
+        return True
+    except ValueError:
+        return False
+
+
+def _path_to_module(path: Path) -> str:
+    if path.is_absolute():
+        raise ValueError(f"Expected a relative path, got: {path}")
+
+    if path.name == "__init__.py":
+        path = path.parent
+    elif path.suffix == ".py":
+        path = path.with_suffix("")
+
+    return ".".join(path.parts)
+
+
+class ModuleResolver:
+    """Resolve file paths to module names, caching results per file name.
+
+    Holds the Python search path info and a cache so that the (relatively
+    expensive) classification is only computed once per file name.
+    """
+
+    def __init__(self, path_info: Optional[PathInfo] = None) -> None:
+        self._path_info: PathInfo = (
+            path_info if path_info is not None else get_python_path_info()
+        )
+        self._cache: Dict[str, Tuple[str, str]] = {}
+
+    def extract_module_name_and_type(self, filename: str) -> Tuple[str, str]:
+        """Extract Python module name and type from a file path.
+
+        Returns:
+            tuple: (module_name, module_type) where module_type is one of:
+                   'stdlib', 'site-packages', 'project', or 'unknown'
+        """
+        try:
+            return self._cache[filename]
+        except KeyError:
+            result = self._extract_module_name_and_type_impl(filename)
+            self._cache[filename] = result
+            return result
+
+    def _extract_module_name_and_type_impl(self, filename: str) -> Tuple[str, str]:
+        if not filename:
+            return ("unknown", "unknown")
+
+        if filename.startswith("<frozen "):
+            return (filename[len("<frozen ") : -1], "stdlib")
+
+        file_path = Path(filename)
+
+        for site_pkg in self._path_info["site_packages"]:
+            if _is_relative_to(file_path, site_pkg):
+                return (
+                    _path_to_module(file_path.relative_to(site_pkg)),
+                    "site-packages",
+                )
+
+        stdlib = self._path_info["stdlib"]
+        if stdlib and _is_relative_to(file_path, stdlib):
+            return (_path_to_module(file_path.relative_to(stdlib)), "stdlib")
+
+        for path_entry in self._path_info["sys_path"]:
+            if _is_relative_to(file_path, path_entry):
+                return (_path_to_module(file_path.relative_to(path_entry)), "project")
+
+        # Fallback: use just the filename, not the full absolute path
+        return (_path_to_module(Path(file_path.name)), "unknown")
+
+    def get_module_for_stack(self, stack: Iterable[Tuple[str, str, int]]) -> str:
+        """Find the top-level module of the closest non-stdlib frame in a stack.
+
+        Walks frames from leaf to root, returning the first non-stdlib module's
+        top-level package name. Returns "<root>" if we reach Memray's own calls
+        into the user's code, or if all frames are stdlib or the stack's empty.
+        """
+        for frame in stack:
+            module_name, module_type = self.extract_module_name_and_type(frame[1])
+            if module_type != "stdlib":
+                top_level = module_name.split(".")[0]
+                if top_level == "memray":
+                    return "<root>"
+                return top_level
+        return "<root>"
--- a/tests/unit/test_stats_reporter.py
+++ b/tests/unit/test_stats_reporter.py
@@ -121,8 +121,53 @@ def fake_stats():
             (("fake_func2", "fake.py", 10), 3 * 2**10),
             (("__main__", "fake.py", 15), 4),
         ],
+        top_modules_by_allocation_size=[
+            ("fake", 20, sum(mem_allocation_list)),
+        ],
+        top_modules_by_allocation_count=[
+            ("fake", 20, sum(mem_allocation_list)),
+        ],
     )
     return s
+
+
+@pytest.fixture(scope="module")
+def empty_module_stats():
+    """Stats with no module attribution (top_modules_by_allocation_size is empty)."""
+    return Stats(
+        metadata=Metadata(
+            start_time=datetime(2023, 1, 1, 1),
+            end_time=datetime(2023, 1, 1, 2),
+            total_allocations=20,
+            total_frames=4,
+            peak_memory=2500000,
+            command_line="python -c 'import json; json.dumps(range(100))'",
+            pid=123456,
+            python_allocator="pymalloc",
+            has_native_traces=False,
+            trace_python_allocators=True,
+            file_format=FileFormat.ALL_ALLOCATIONS,
+            main_thread_id=0x1,
+        ),
+        total_num_allocations=20,
+        total_memory_allocated=3341500,
+        peak_memory_allocated=2500000,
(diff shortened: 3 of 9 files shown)


grantlouisherman:
Hey @ivonastojanovic do you have a timeline of when this might land? I started looking at https://github.com/bloomberg/memray/issues/766 and I can just rebase onto this but its always easier if this is merged into main. Obviously no rush.

src/memray/reporters/module_tools.py
```diff
@@ -0,0 +1,127 @@
...
+    Returns:
+        dict: Dictionary containing stdlib path, site-packages paths, and sys.path entries.
+    """
+    stdlib: Optional[Path] = None
+    # Get standard library path from os module location
+    try:
+        if hasattr(os, "__file__") and os.__file__:
+            stdlib = Path(os.__file__).parent
+    except (AttributeError, OSError):  # pragma: no cover
+        pass
```
godlygeek: I think this should probably be:
```suggestion
    stdlib = sysconfig.get_config_var("LIBDEST")
```

src/memray/reporters/module_tools.py
```diff
@@ -0,0 +1,127 @@
...
+
+    # Get site-packages directories
+    site_packages: List[Path] = []
+    try:
+        site_packages.extend(Path(p) for p in site.getsitepackages())
+    except (AttributeError, OSError):  # pragma: no cover
```
godlygeek: Why would either of these exceptions ever occur? I can't think of any way either of these could happen...

src/memray/reporters/module_tools.py
```diff
@@ -0,0 +1,127 @@
...
+        pass
+
+    # Get user site-packages
+    try:
+        user_site = site.getusersitepackages()
+        if user_site and Path(user_site).exists():
+            site_packages.append(Path(user_site))
+    except (AttributeError, OSError):  # pragma: no cover
```
godlygeek: As above, when would these exceptions happen?

src/memray/reporters/module_tools.py
```diff
@@ -0,0 +1,127 @@
...
+    site_packages: List[Path] = []
+    try:
+        site_packages.extend(Path(p) for p in site.getsitepackages())
+    except (AttributeError, OSError):  # pragma: no cover
+        pass
+
+    # Get user site-packages
+    try:
+        user_site = site.getusersitepackages()
+        if user_site and Path(user_site).exists():
```
godlygeek: I don't believe it's possible for `site.getusersitepackages()` to return something falsey, so I don't think the `user_site and` is needed here...

src/memray/reporters/module_tools.py
```diff
@@ -0,0 +1,127 @@
...
+
+def _is_relative_to(path: Path, base: Path) -> bool:
+    try:
+        path.relative_to(base)
+        return True
+    except ValueError:
+        return False
+
+
+def extract_module_name(filename: str, path_info: PathInfo) -> Tuple[str, str]:
```
godlygeek:
```suggestion
def extract_module_name_and_type(filename: str, path_info: PathInfo) -> Tuple[str, str]:
```

src/memray/reporters/module_tools.py
```diff
@@ -0,0 +1,127 @@
...
+
+    Returns:
+        tuple: (module_name, module_type) where module_type is one of:
+               'stdlib', 'site-packages', 'project', or 'other'
+    """
+    if not filename:
+        return ("unknown", "other")
+
+    if filename.startswith("<frozen "):
+        return ("stdlib", "stdlib")
```
godlygeek: I'd think we should do
```suggestion
        return (filename[len("<frozen "):-1], "stdlib")
```

src/memray/reporters/module_tools.py
```diff
@@ -0,0 +1,127 @@
...
+
+    try:
+        file_path = Path(filename)
+    except (ValueError, OSError):
```
godlygeek: Can the `Path` constructor raise `ValueError` or `OSError`? I can't think of how... it doesn't make any syscalls, so I _really_ wouldn't expect to see `OSError`...

src/memray/reporters/module_tools.py
```diff
@@ -0,0 +1,127 @@
...
+        return (str(filename), "other")
```
godlygeek: Why `str(filename)`? Isn't it already a str?

src/memray/reporters/module_tools.py
```diff
@@ -0,0 +1,127 @@
...
+    except ValueError:
+        return False
+
+
+def extract_module_name(filename: str, path_info: PathInfo) -> Tuple[str, str]:
+    """Extract Python module name and type from file path.
+
+    Returns:
+        tuple: (module_name, module_type) where module_type is one of:
+               'stdlib', 'site-packages', 'project', or 'other'
```
godlygeek: "other" would be better as "unknown", I think

src/memray/reporters/module_tools.py
```diff
@@ -0,0 +1,127 @@
...
+        return (str(filename), "other")
+
+    for site_pkg in path_info["site_packages"]:
+        if _is_relative_to(file_path, site_pkg):
+            return (_path_to_module(file_path.relative_to(site_pkg)), "site-packages")
+
+    if path_info["stdlib"] and _is_relative_to(file_path, path_info["stdlib"]):
+        return (_path_to_module(file_path.relative_to(path_info["stdlib"])), "stdlib")
+
+    if not str(file_path).startswith(("<", "[")):
```
godlygeek: What's the case where we see `[` as the first character?

src/memray/reporters/module_tools.py
```diff
@@ -0,0 +1,127 @@
...
+
+def _path_to_module(path: Path) -> str:
+    if path.suffix == ".py":
+        path = path.with_suffix("")
+
+    parts = [p for p in path.parts if p != path.root and p != path.drive]
+
+    # Handle __init__ files - they represent the package itself
+    if parts and parts[-1] == "__init__":
+        parts = parts[:-1]
```
godlygeek: I think it'd be slightly cleaner to do this as:

```suggestion
    if path.name == "__init__.py":
        path = path.parent
    elif path.suffix == ".py":
        path = path.with_suffix("")

    parts = [p for p in path.parts if p != path.root and p != path.drive]
```

src/memray/reporters/module_tools.py
```diff
@@ -0,0 +1,127 @@
...
+
+    # Fallback: use just the filename, not the full absolute path
+    return (_path_to_module(Path(file_path.name)), "other")
+
+
+def _path_to_module(path: Path) -> str:
+    if path.suffix == ".py":
+        path = path.with_suffix("")
+
+    parts = [p for p in path.parts if p != path.root and p != path.drive]
```
godlygeek: We always expect a relative path, right? Maybe we should just raise a type error if `path.is_absolute()`, and then drop this `if p != path.root and p != path.drive` logic.

src/memray/reporters/module_tools.py
```diff
@@ -0,0 +1,127 @@
...
+
+    # Handle __init__ files - they represent the package itself
+    if parts and parts[-1] == "__init__":
+        parts = parts[:-1]
+
+    return ".".join(parts) if parts else path.stem
```
godlygeek: When would the `if parts` ever be false?

src/memray/reporters/module_tools.py
```diff
@@ -0,0 +1,127 @@
...
+    """Find the top-level module of the closest non-stdlib frame in a stack.
+
+    Walks frames from leaf to root, returning the first non-stdlib module's
+    top-level package name. Returns None if every frame is stdlib or the
+    stack is empty.
+    """
+    for frame in stack:
+        module_name, module_type = extract_module_name(frame[1], path_info)
+        if module_type != "stdlib":
+            top_level = module_name.split(".")[0] if module_name else module_name
```
godlygeek: Is `module_name` able to be falsey? I'd expect `"<unknown>"` or something like that at worst...

src/memray/_memray.pyx
```diff
@@ -1512,15 +1515,35 @@ def compute_statistics(
...
-                    reader.getLatestPythonLocationId(reader.getLatestAllocation()),
+                    allocation,
+                    reader.getLatestPythonLocationId(allocation),
                 )
+
+                if (AllocatorType(<int>allocation.allocator)
+                        not in (AllocatorType.FREE, AllocatorType.MUNMAP,
+                                AllocatorType.PYMALLOC_FREE,
+                                ObjectTrackingEvent.OBJECT_DESTROYED)
+                        and allocation.frame_index != 0):
```
godlygeek: Do we actually need to check any of this? I _think_ that `allocation.frame_index != 0` implies `AllocatorType(<int>allocation.allocator) not in (AllocatorType.FREE, AllocatorType.MUNMAP, AllocatorType.PYMALLOC_FREE, ObjectTrackingEvent.OBJECT_DESTROYED)` - if we have a non-zero `frame_index` then we have a stack to look at, and if we have a stack to look at it must not be a deallocation, because we don't record stacks for deallocations.

And indeed, if I change this to `if True:` the tests all still pass - so I'm _pretty_ sure we don't need this `if` at all, we just need to make sure to correctly handle the degenerate case where `Py_GetStackFrame` returns `[]` - and it seems like we do.

src/memray/_memray.pyx
```diff
@@ -1512,15 +1515,35 @@ def compute_statistics(
...
+                    stack = reader.Py_GetStackFrame(allocation.frame_index)
+                    module_name = get_module_for_stack(stack, path_info)
+                    if module_name is not None:
+                        entry = module_stats.get(module_name)
+                        if entry is None:
+                            module_stats[module_name] = [1, allocation.size]
+                        else:
+                            entry[0] += 1
+                            entry[1] += allocation.size
```
godlygeek: What you have is fine, but it'd be slightly cleaner to do:

```suggestion
                        entry = module_stats.setdefault(module_name, [0, 0])
                        entry[0] += 1
                        entry[1] += allocation.size
```
`dict.setdefault` basically behaves like
```py
def setdefault(dict, key, default_val):
    if key not in dict:
        dict[key] = default_val
    return dict[key]
```

src/memray/_memray.pyx
```diff
@@ -1554,6 +1577,14 @@ def compute_statistics(
         for count_and_loc in aggregator.topLocationsByCount(num_largest)
     ]
 
+    top_modules = sorted(
+        module_stats.items(), key=lambda x: x[1][0], reverse=True
```
godlygeek, line 1585: I've thought about this a little bit: I'm pretty sure that if we have to choose only one, it's probably more useful to show the top modules by number of bytes allocated, rather than the top modules by number of allocations. Or we could show both things. But if we pick only one, I think it should be bytes.

That's my fault, I should have thought it through on the issue and been more specific.

tests/unit/test_stats_reporter.py
```diff
@@ -121,10 +121,51 @@ def fake_stats():
...
+        top_allocations_by_module=[
+            ("fake", 20, sum(mem_allocation_list)),
+        ],
     )
     return s
 
 
+@pytest.fixture(scope="module")
+def empty_module_stats():
+    """Stats where all allocations come from stdlib, so no modules are attributed."""
```
godlygeek: Hm. I think that this shows that our fallback is wrong. I think if all of the allocations come from the stdlib, we should report the module as `<root>` or `__main__` or `<string>` or something like that... _something_ called the stdlib, after all. We might have trouble figuring out what to call it, but I don't think we should ignore it entirely...

godlygeek (changes requested):

godlygeek:
This looks pretty good. I have a lot of comments, but they're almost all pretty minor stuff.

src/memray/reporters/module_tools.py
ivonastojanovic: Oh you're right, checked [here](https://docs.python.org/3/library/site.html#module-site), will remove the exceptions.

src/memray/_memray.pyx
ivonastojanovic: Oh you're totally right, we don't need this check then

src/memray/_memray.pyx
ivonastojanovic: Yeah, that makes sense, I agree that bytes allocated is probably the more useful default if we have to choose only one. Allocation count can still be valuable though, so showing both might be the best option overall.

godlygeek: I'd be fine with showing both

src/memray/reporters/module_tools.py
ivonastojanovic: Ah, actually we can probably remove this check entirely since filenames starting with `<` won’t satisfy the `sys_path` loop condition anyway.

tests/unit/test_stats_reporter.py
ivonastojanovic: What do you think about `__main__` then?

src/memray/_memray.pyx
ivonastojanovic: @godlygeek Updated the tests, we can’t remove the `if allocation.frame_index != 0` check because deallocations are now being counted under `__main__`.

tests/unit/test_stats_reporter.py
godlygeek: `__main__` is misleading since it may not be from the main module. I'm gonna flip this to do `<root>` instead, which matches what we use for the root of our flame graphs.

src/memray/_memray.pyx
godlygeek: Ah, actually, we do want to explicitly check for deallocations rather than empty stacks, so that allocations with no Python stack can be attributed to the tracking root as well instead of being skipped (due to having `allocation.frame_index == 0`)

godlygeek:
I pushed some small fixes. At this point this PR looks semantically correct to me, but it's very slow when working with large numbers of allocations, even after I attempted to add some caching to it. I'm thinking we're going to need to push some of what's currently happening in the Python layer into the Cython or C++ layer to get this to be fast enough...

Running:
```sh
memray run --trace-python-allocators --no-compress -fo output.bin -m test test_set
```
to create a capture file and then analyzing it with
```sh
time memray stats -f --json output.bin
```
takes 8 seconds on the main branch for me, and 25 seconds on this PR branch even after adding some caching, so `memray stats` is spending around 2/3rds of its time attributing allocations to the closest non-stdlib module with what's in this PR now.

godlygeek:
Granted, that's a particularly stressful stress test - since the `python -m test` test suite is part of the stdlib, it has to walk further up the stack than might normally be expected to find non-stdlib modules to attribute allocations to.

godlygeek:
I have a few different ideas about how to make this fast enough. Roughly from easiest to hardest:

1. Add a cache to `extract_module_name_and_type` so that it doesn't repeatedly compute the answer for the same file name (this is probably easiest to do by factoring `extract_module_name_and_type` and `get_module_for_stack` into methods of a class that holds the cache and the path info as member variables)
2. Expose a new `RecordReader::getCallerStackId()` that takes a frame index representing a call stack, and returns the frame index representing the call stack with all but the most recent frame. Then, change `compute_statistics` to iterate over each Python stack one frame at a time, stopping when we reach the first non-stdlib frame (saving us the work of creating Python objects representing frames we never need to look at because they're below the topmost non-stdlib frame, and more importantly allowing us to memoize the results for intermediate stacks and avoid recomputing the answer for each unique call stack that calls into stdlib code from the same initial call stack)
3. Move everything from `module_tools.py` into `cdef` functions in `_memray.pyx` and try to get a bit of speedup by avoiding `str` and `pathlib.Path` operations in favor of faster `std::string` or `std::filesystem` operations

Or of course, we could just accept that it makes things take 3x as long, and put it behind a flag so people need to opt into the slower but more informative behavior.

I lean towards trying both (1) and (2) though, and seeing how far those two changes get us. I'm betting that they should be enough...

godlygeek:
> I have a few different ideas about how to make this fast enough. Roughly from easiest to hardest:
>
> 1\. Add a cache to `extract_module_name_and_type` so that it doesn't repeatedly compute the answer for the same file name (this is probably easiest to do by factoring `extract_module_name_and_type` and `get_module_for_stack` into methods of a class that holds the cache and the path info as member variables)

Turns out that was enough all by itself. Without these changes my favorite test case takes ~8.0 seconds on my laptop, and with this new cache it's at ~8.45 seconds, so about 5% slower. That's a price I'm willing to pay for this feature.

godlygeek (approved):
LGTM at this point, but I've changed this pretty heavily so I'd prefer someone else's eyes on it before I land it.

There's one outstanding issue I see here: the stdlib and site-packages detection is based on the environment that the reporter runs in, not on the environment that the tracked process runs in. The best way to address that is going to be recording the results from `get_python_path_info` in the capture file and reusing those captured paths in the reporter, but I'm going to push that off to a separate PR since this one's already kind of dragging and I'm happy to make that change myself (unless someone wants to beat me to it!)

src/memray/_memray.pyx
```diff
@@ -1517,15 +1514,30 @@ def compute_statistics(
         total=total,
         report_progress=report_progress,
     )
+
+    module_resolver = ModuleResolver()
```
sarahmonod: I find it a bit confusing to declare this 70+ lines before it is first used. Looking at the constructor's implementation, I don't see a reason to do it at that point in particular.

sarahmonod (approved):

src/memray/_memray.pyx
godlygeek: Ah, that's thanks to some refactoring - it used to be used in the `while True` loop, but after the change to add some caching it's now used in a second loop lower down. I'll move it.

PR #939, opened 2026-05-19
feat: add light/dark theme toggle for HTML reports

### LINKED ISSUE #650: Dark Mode for Flame graph reports

- [X] I have searched the existing proposals

I have an issue when looking at Flame graphs, especially in a dark room, and would like to reduce white light to my eyes.

I would love if generated Flame graphs had some sort of toggle at the top of the HTML file which could switch between Light/Dark mode.

Something else to consider would be to attempt to read the machine default value for Light/Dark mode, and initializing Flame graphs to this value.

There are some plugins out there that attempt to auto convert web pages out there that don't support Dark Mode. I am not sure that I have tried one of these with a flame graph, but I could give it a whirl if we think it's worth it.

Do-yoon:
## Summary
Adds a light/dark theme toggle to the flamegraph and table HTML reports.
Closes #650.

## What's included
- Toggle button (🌙/☀️) in the navbar
- Initial theme follows the OS `prefers-color-scheme`
- Selection persisted via `localStorage`
- `data-theme` is set before first paint to avoid FOUC
- CSS variables for light and dark palettes (dark uses a black base)
- Bootstrap overrides for navbar, modals, forms, tables, DataTables pagination, search input, dropdowns and scrollbars
- Recolored flame graph bar fills for the dark theme
- News entry (`news/650.feature.rst`)

## Test plan
- [x] Generate flamegraph report and verify both themes
- [x] Generate table report and verify both themes (incl. Previous/Next, search)
- [x] Verify initial paint follows OS dark/light preference
- [x] Verify localStorage persistence across reloads


sarahmonod:
This PR definitely deserves a NEWS entry. @Do-yoon do you know how to use towncrier? https://github.com/bloomberg/pystack/pull/305 has an example if you want one.

godlygeek:
@pradyunsg I'm going to defer to you for reviewing this one, if you don't mind doing it.

Do-yoon:
@sarahmonod 

Thanks for pointing me to towncrier — I hadn't used it before, so I really appreciate the pystack#305 example.

Before I push, I just want to make sure I've got the convention right: I'm planning to add a single fragment at `news/650.feature.rst` containing a one-line description of the feature. Does that look correct to you, or should I use a different filename/category?

godlygeek:
Yes, that's the correct filename

Do-yoon:
I've added the news fragment at `news/650.feature.rst` and pushed it as a new commit, so this should be ready for another look whenever you get the chance. Let me know if anything else is needed.

Do-yoon:
@godlygeek
@pradyunsg

Rebased onto the latest main to resolve a merge conflict.

The conflict was only in `src/memray/reporters/templates/assets/table.js`, which is a generated webpack bundle — both this PR (dark theme changes in `assets/common.js`) and #857 (row deduplication in assets/table.js) had regenerated it, so it couldn't be merged textually. There were no conflicts at the source level.

To resolve it, I took main's version of the bundle and regenerated it with `npm ci && npm run build` on top of the rebased sources, so the new bundle includes both the deduplication change from #857 and the theme support from this PR. The other bundles didn't need regenerating since #857 doesn't touch common.js.

pradyunsg (approved):
LGTM, as a bi-state theme mode.

godlygeek:
Thanks for the contribution, @Do-yoon!

PR #940, opened 2026-05-20
Add diagnostic message for missing permissions on attach

### LINKED ISSUE #925: Permission error in _attach.lldb leaves target process in a bad state

- [x] I have searched the existing issues

### Current Behavior

When trying to run memray as root against a process not owned by root, the dlopen can fail and leave the debugger in a bad state.  There's a call to dlerror but that doesn't abort the block...

lldb shows a lot of information with the useful error in the middle, but also doesn't clear the breakpoint when this happens

```
Executable module set to "/apps/python3.11.15/bin/python3.11".
Architecture set to: x86_64-pc-linux-gnu.
(lldb) expr char $libpath[]="/root/.shiv/memray_bffd7ac1e7e719c989f912fcc43b5e8f3867ec1328a5efdd12122c6958934853/site-packages/memray/_inject.abi3.so"
(lldb) expr int $port=30420
(lldb) expr void* $rtld_default=(void*)0
(lldb) expr int $rtld_now=2
(lldb) command source -s 0 '/root/.shiv/memray_bffd7ac1e7e719c989f912fcc43b5e8f3867ec1328a5efdd12122c6958934853/site-packages/memray/commands/_attach.lldb'
Executing commands in '/root/.shiv/memray_bffd7ac1e7e719c989f912fcc43b5e8f3867ec1328a5efdd12122c6958934853/site-packages/memray/commands/_attach.lldb'.
(lldb) p "MEMRAY: Attached to process."
...
(lldb)  expr auto $dll = ((void*(*)(const char*, int))$dlopen)($libpath, $rtld_now)
(lldb)  p ((char*(*)(void))$dlerror)()
(char *) $7 = 0x00005ea2e14cbd00 "/root/.shiv/memray_bffd7ac1e7e719c989f912fcc43b5e8f3867ec1328a5efdd12122c6958934853/site-packages/memray/_inject.abi3.so: cannot open shared object file: Permission denied"
(lldb)  expr auto $spawn = $dlsym($dll, "memray_spawn_client")
(lldb)  p ((int(*)(int))$spawn)($port)?"FAILURE":"SUCCESS"
error: Aborting reading of commands after command #7: 'p ((int(*)(int))$spawn)($port)?"FAILURE":"SUCCESS"' failed with error: Execution was interrupted, reason: signal SIGSEGV: invalid address (fault address: 0x0).
The process has been returned to the state before expression evaluation.
Process 9347 resuming
Process 9347 stopped
* thread #1, name = 'gunicorn', stop reason = breakpoint 1.
    frame #0: 0x0000718e33ca9790 libc.so.6`malloc
libc.so.6`malloc:
->  0x718e33ca9790 <+0>: endbr64
    0x718e33ca9794 <+4>: pushq  %rbp
    0x718e33ca9795 <+5>: movq   %rsp, %rbp
    0x718e33ca9798 <+8>: pushq  %r13
```

gdb doesn't have nearly as much info, but I think it's the same (segfault while in one of the functions with a breakpoint).

```
$2 = "MEMRAY: Checking if process is Python 3.7+."
$3 = {void *(size_t)} 0x718e341d95a0 <PyMem_Malloc>
$4 = {void *(size_t, size_t)} 0x718e341d96b0 <PyMem_Calloc>
$5 = {void *(void *, size_t)} 0x718e341e3040 <PyMem_Realloc>
$6 = {void (void *)} 0x718e341d9a10 <PyMem_Free>
$7 = "MEMRAY: Process is Python 3.7+."
$8 = 0
Breakpoint 1 at 0x718e33ca9798
Breakpoint 2 at 0x718e33caac80
Breakpoint 3 at 0x718e33ca9e48
Breakpoint 4 at 0x718e33ca9ad0
Breakpoint 5 at 0x718e341d95a0: PyMem_Malloc. (16 locations)
Breakpoint 6 at 0x718e341d96b0: PyMem_Calloc. (2 locations)
Breakpoint 7 at 0x718e341e2e1d: PyMem_Realloc. (5 locations)
Breakpoint 8 at 0x718e341d9a10: PyMem_Free. (17 locations)
Breakpoint 9 at 0x718e3422a3e1: PyErr_CheckSignals. (11 locations)
Breakpoint 10 at 0x718e3421b8f0: file Objects/object.c, line 1538.
Thread 1 "gunicorn" hit Breakpoint 10, PyCallable_Check (x=x@entry=0x0) at Objects/object.c:1538
1538    Objects/object.c: No such file or directory.
#0  PyCallable_Check (x=x@entry=0x0) at Objects/object.c:1538
#1  0x0000718e3413e0bf in make_pending_calls (interp=0x718e345798b8 <_PyRuntime+58936>) at Python/ceval.c:709
#2  0x0000718e3410807e in eval_frame_handle_pending (tstate=<optimized out>) at Python/ceval.c:1190
#3  _PyEval_EvalFrameDefault (tstate=<optimized out>, frame=<optimized out>, throwflag=<optimized out>) at Python/ceval.c:1748
#4  0x0000718e3421ba96 in _PyEval_EvalFrame (throwflag=0, frame=0x718e34710718, tstate=0x718e34593c38 <_PyRuntime+166328>) at ./Include/internal/pycore_ceval.h:73
#5  _PyEval_Vector (kwnames=<optimized out>, argcount=2, args=0x718e29734a98, locals=0x0, func=0x718e32a363e0, tstate=0x718e34593c38 <_PyRuntime+166328>) at Python/ceval.c:6434
#6  _PyFunction_Vectorcall (func=0x718e32a363e0, stack=0x718e29734a98, nargsf=<optimized out>, kwnames=<optimized out>) at Objects/call.c:393
#7  0x0000718e29a78531 in ?? () from /apps/perfbench-backend/lib/python3.11/site-packages/uvloop/loop.cpython-311-x86_64-linux-gnu.so
#8  0x0000000000000000 in ?? ()
Thread 1 "gunicorn" received signal SIGSEGV, Segmentation fault.
0x00007ffcf3a91f6f in ?? ()
```

### Expected Behavior

1. short, concise error ("permission denied: <filename>")
2. clear breakpoints once one is hit

### Steps To Reproduce

1. run target process as regular user
2. run memray as root (making sure its files are not readable to regular users)

### Memray Version

1.19.x ?

### Python Version

3.11

### Operating System

Linux

### Anything else?

_No response_

tnytown:
Resolves #925.

**Describe your changes**
When the debugger calls `dlopen()` on the memray library in the target
process, it does so from the context of the target process' user. This
is ordinarily not a problem; but if memray was installed by another user
to a location inaccessible from the target user, the `dlopen()` fails,
causing the subsequent `dlsym()` and `memray_spawn_client()` calls to
fail as well. #925 is an instance of this problem where memray was
installed by the root user.

This changeset adjusts the attach scripts to short-circuit on `dlopen()`
failure before attempting to run `memray_spawn_client()`. When failure
happens, we match upon the `dlerror()` output and print a diagnostic
message.

**Testing performed**
Installed and ran memray as root, attaching to a target process running
as another user. The diagnostic message surfaced correctly. Tested with
the following configuration:

```
Linux nixos 6.18.19 #1-NixOS SMP Thu Mar 19 15:08:51 UTC 2026 aarch64 GNU/Linux
GNU gdb (GDB) 17.1
lldb version 18.1.8
Python 3.13.12 (main, Mar 20 2026, 00:36:06) [Clang 22.1.1 ]
```

**Additional context**
Future work might involve gracefully handling this case by copying the
required shared library and other resources to a permissive temporary
directory.

### DIFF
--- a/src/memray/commands/_attach.lldb
+++ b/src/memray/commands/_attach.lldb
@@ -9,7 +9,10 @@ p ((void(*)(void*))PyMem_Free)
 
 p "MEMRAY: Process is Python 3.7+."
 
-# When adding new breakpoints, also update _attach.gdb
+# When adding new breakpoints, also update _attach.gdb - but note that
+# PyErr_CheckSignals and PyCallable_Check are intentionally excluded,
+# as for some reason including them caused lldb on Linux to stop at
+# breakpoints that had already been deleted.
 breakpoint set -b malloc -b calloc -b realloc -b free -b PyMem_Malloc -b PyMem_Calloc -b PyMem_Realloc -b PyMem_Free
 
 # Set commands to execute when breakpoint is reached
@@ -21,7 +24,8 @@ expr auto $dlerror = $dlsym($rtld_default, "dlerror")
 expr auto $dll = ((void*(*)(const char*, int))$dlopen)($libpath, $rtld_now)
 p ((char*(*)(void))$dlerror)()
 expr auto $spawn = $dlsym($dll, "memray_spawn_client")
-p ((int(*)(int))$spawn)($port)?"FAILURE":"SUCCESS"
+p ((char*(*)(void))$dlerror)()
+p (!$dll || !$spawn || ((int(*)(int))$spawn)($port)) ? "FAILURE" : "SUCCESS"
 DONE
 
 continue
--- a/news/940.bugfix.rst
+++ b/news/940.bugfix.rst
@@ -1,0 +1,1 @@
+Fix a bug that crashed target processes when ``memray attach`` fails. The underlying permissions issue is now accounted for and a troubleshooting message prints when it arises.
--- a/src/memray/commands/_attach.gdb
+++ b/src/memray/commands/_attach.gdb
@@ -18,6 +18,13 @@ p "MEMRAY: Process is Python 3.7+."
 set scheduler-locking on
 call (int)Py_AddPendingCall(&PyCallable_Check, (void*)0)
 
+p "Checking if we can access the library"
+set $result = (int)access($libpath, 5)
+if $result != 0
+    p "cannot open shared object file"
+    quit 1
+end
+
 # When updating this list, also update the "commands" call below,
 # and the breakpoints hardcoded for lldb in attach.py
 b malloc
--- a/src/memray/commands/attach.py
+++ b/src/memray/commands/attach.py
@@ -231,6 +231,13 @@ def debugger_inject(debugger: str, pid: int, port: int, verbose: bool) -> str | 
         if "MEMRAY: Process is Python 3.7+." not in output:
             return "The process does not seem to be running Python 3.7 or newer."
 
+    if "cannot open shared object file" in output:
+        return (
+            "The target process couldn't open the memray shared library.\n"
+            "This can occur if the memray installation is owned by a different user "
+            "and is inaccessible to the target process."
+        )
+
     return "An unexpected error occurred. Run with --verbose to debug the failure."
 
 


godlygeek:
Hm. I'm still getting a crash even with these changes. What I've done is:

In one window, start a program that loops forever making small allocations:

```shell-session
$ cat >crash_test.py <<"EOF"
import mmap
import os
import threading
import time

print(f"{os.getpid() = }")


def thread_body():
    global x
    while True:
        mmap.mmap(-1, 1024 * 1024)
        x = "x" * 1024 * 1024
        time.sleep(1)

t1 = threading.Thread(target=thread_body)
t1.start()
EOF

$ python3.14 crash_test.py
os.getpid() = 18422
```

Then in another window:
- Grab a root shell with `sudo -s -H`
- Make a temp directory with `mktemp -d`
- Change to that directory with e.g. `cd /tmp/tmp.YCeNHMoYt3`
- Make a virtualenv with `python3.14 -m venv .venv`
- Activate the virtualenv with `source .venv/bin/activate`
- Install memray with `pip install /path/to/clone/of/memray`
- Attach to the user process with `memray attach --verbose --method=gdb 18422`

I end up with:
```
../../gdb/dummy-frame.c:211: internal-error: dummy_frame_pop: Assertion `dp != NULL' failed.
A problem internal to GDB has been detected,
further debugging may prove unreliable.
----- Backtrace -----
0x53e556 ???
0x53e5a6 ???
0xb8ca6e ???
0xb8cde4 ???
0xd7787d ???
0x642abe ???
0x7fa07f ???
0x7f8385 ???
0x76049b ???
0x760748 ???
0x46582c ???
0x560afa ???
0x76620f ???
0x4dbbc7 ???
0x4db949 ???
0x75f16a ???
0x92f12a ???
0x59592e ???
0x59aab2 ???
0xb2e469 ???
0x5a75bd ???
0x5a7c1e ???
0x501d27 ???
0x501ec7 ???
0x76a0a1 ???
0xb2db71 ???
0x5aa223 ???
0x772f19 ???
0x58b637 ???
0x58b797 ???
0x58b80f ???
0x8a636e ???
0x8a6526 ???
0x8a79a9 ???
0x8a7c00 ???
0x8a7c9f ???
0x414381 ???
0x7891eb63a7e4 ???
0x41425a ???
0xffffffffffffffff ???
---------------------
../../gdb/dummy-frame.c:211: internal-error: dummy_frame_pop: Assertion `dp != NULL' failed.
A problem internal to GDB has been detected,
further debugging may prove unreliable.
Quit this debugging session? (y or n) [answered Y; input not from terminal]

This is a bug, please report it.  For instructions, see:
<https://bbgithub.dev.bloomberg.com/gnu/gdb/issues>.


An unexpected error occurred. Run with --verbose to debug the failure.
```

If I add some extra `p` calls to the gdb script to tell where that crash is happening, it happens _inside_ the `dlopen` call. None of the new error handling added in this PR seems to fire, because it seems like we crash before ever reaching any of it...

src/memray/commands/_attach.lldb
```diff
@@ -10,7 +10,7 @@ p ((void(*)(void*))PyMem_Free)
 p "MEMRAY: Process is Python 3.7+."
 
 # When adding new breakpoints, also update _attach.gdb
-breakpoint set -b malloc -b calloc -b realloc -b free -b PyMem_Malloc -b PyMem_Calloc -b PyMem_Realloc -b PyMem_Free
+breakpoint set -b malloc -b calloc -b realloc -b free -b PyMem_Malloc -b PyMem_Calloc -b PyMem_Realloc -b PyMem_Free -b PyErr_CheckSignals -b PyCallable_Check
```
godlygeek: These 2 were intentionally omitted  from this script. Per 52b3e1bffbeb193bcf0ad04603ab9020c1d586e1:

    Add 2 new breakpoints to our gdb script, on `PyCallable_Check` and
    `PyError_CheckSignals`. Also, use `Py_AddPendingCall` to schedule a call
    to `PyCallable_Check` on the main thread, in the hopes of triggering one
    of our breakpoints to happen sooner.

    For now, apply this only to our gdb script, and not to our lldb script.
    Making the analogous changes to the lldb script results in lldb on Linux
    stopping at an already-deleted breakpoint, and I haven't figured out
    a workaround or fix for that yet.

Granted I should have left a comment to that effect, and it's lucky that I remembered that 😅

godlygeek:
I've pushed a commit with an alternative approach to the gdb script that seems to work for me. Can you check if it works for you, @tnytown / @thatch?

src/memray/commands/_attach.lldb
tnytown: Ouch! I'll get rid of this; would you rather I rebase to drop the commit or attach another commit reverting?

tnytown:
@godlygeek cff28f1 works for me on `aarch64-linux` / GDB 17.1 ... very odd that this happens in the first place, also it's fortunate that gdb supports branching in its script language :) would you like me to update the lldb script to match the `access` and branch approach?

src/memray/commands/_attach.lldb
godlygeek: No preference, I plan to squash merge this either way.

godlygeek:
No, I'm fine with leaving the lldb one as you have it. It seems to be working, and these scripts are already subtly different in lots of ways, so I'm not worried about one more difference.

godlygeek (approved):
I've reverted the changes to the breakpoints in the LLDB script, and added a comment explaining why the two you added are missing instead. And I've gone with my fix for the gdb script instead of the one you proposed.

After those changes, this looks good to me - thanks for the contribution, @tnytown!

PR #943, opened 2026-05-24
Track Apache Arrow mimalloc allocations

pablogsal:
Hook the `arrow_mi_malloc_aligned`, `arrow_mi_realloc_aligned`, and `arrow_mi_free` symbols introduced in Apache Arrow 25 so Memray can attribute Arrow logical allocations instead of seeing only mimalloc mmap arenas.

Scope: Linux Arrow 25+ builds using mimalloc with interposable `arrow_mi_*` wrapper symbols. Official PyArrow 25 Linux wheels satisfy these requirements.

The hooks are re-resolved when libraries are loaded after Tracker initialization. To avoid crossing independent statically linked mimalloc instances, Memray patches only the ELF object that defines the selected wrapper; if multiple providers are loaded, later providers are left untracked rather than routed through the wrong allocator.

Tests cover loading Arrow before tracking, during tracking, and between Tracker sessions, plus isolation between multiple providers.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -57,7 +57,7 @@ jobs:
           echo /home/runner/node_modules/.bin >>"$GITHUB_PATH"
       - name: Install Python dependencies
         run: |
-          ${PYTHON} -m pip install --upgrade pip scikit-build-core cython pkgconfig
+          ${PYTHON} -m pip install --upgrade pip scikit-build-core cython pkgconfig "pyarrow>=25.0"
           make test-install
       - name: Disable ptrace security restrictions
         run: |
--- a/src/memray/_memray/elf_shenanigans.cpp
+++ b/src/memray/_memray/elf_shenanigans.cpp
@@ -41,9 +41,17 @@ patch_symbol(
         const Hook& hook,
         typename Hook::signature_t intercept,
         const char* symname,
+        const SymbolTable& symbols,
         Addr addr,
         bool restore_original)
 {
+    const bool is_arrow_mimalloc_hook = strncmp(symname, "arrow_mi_", 9) == 0;
+    if (is_arrow_mimalloc_hook
+        && (!hook || symbols.getSymbolAddress(symname) != reinterpret_cast<Addr>(hook.d_original)))
+    {
+        return;
+    }
+
     // Make sure that we can read and write to the page where the address that we are trying to
     // patch;
     if (unprotect_page(addr) < 0) {
@@ -89,6 +97,7 @@ overwrite_elf_table(
                 MEMRAY_ORIG(hookname),                                                                  \
                 &intercept::hookname,                                                                   \
                 symname,                                                                                \
+                symbols,                                                                                \
                 symbol_addr,                                                                            \
                 restore_original);                                                                      \
         continue;                                                                                       \
--- a/src/memray/_memray/hooks.cpp
+++ b/src/memray/_memray/hooks.cpp
@@ -418,6 +418,59 @@ aligned_alloc(size_t alignment, size_t size) noexcept
     return ret;
 }
 
+// Apache Arrow mimalloc wrapper hooks (apache/arrow#41128). These wrap
+// mimalloc allocations that are otherwise opaque to memray because mimalloc
+// is statically linked into libarrow and serves allocations out of mmap
+// arenas. Tracked under the existing aligned-alloc / realloc / free families.
+void*
+arrow_mi_malloc_aligned(size_t size, size_t alignment) noexcept
+{
+    assert(MEMRAY_ORIG(arrow_mi_malloc_aligned));
+
+    void* ret;
+    {
+        tracking_api::RecursionGuard guard;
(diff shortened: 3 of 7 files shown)


src/memray/_memray/tracking_api.cpp
```diff
@@ -898,6 +898,11 @@ Tracker::~Tracker()
 {
     RecursionGuard guard;
     tracking_api::Tracker::deactivate();
+#ifdef MEMRAY_HAS_GHOST_STACK
+    if (d_fast_unwind) {
+        ghost_stack_dump_stats();
+    }
+#endif
```
godlygeek: Looks like this was a rebase mistake

godlygeek:
Looks like https://github.com/apache/arrow/pull/41128 has landed. Should we move forward with this, @pablogsal ?

src/memray/_memray/tracking_api.cpp
pablogsal: Fixed in 0ce0c24: the unrelated ghost-stack block was removed while rebasing this PR onto current main.

pablogsal:
Ready for rock and roll

.github/workflows/coverage.yml
```diff
@@ -57,7 +57,7 @@ jobs:
           echo /home/runner/node_modules/.bin >>"$GITHUB_PATH"
       - name: Install Python dependencies
         run: |
-          ${PYTHON} -m pip install --upgrade pip scikit-build-core cython pkgconfig
+          ${PYTHON} -m pip install --upgrade pip scikit-build-core cython pkgconfig "pyarrow==25.0.*"
```
godlygeek: Why `==25.0.*` instead of something like `>=25.0`? Why would we not want to upgrade to 26.x once it comes out?

src/memray/_memray/elf_shenanigans.cpp
```diff
@@ -41,9 +41,17 @@ patch_symbol(
         const Hook& hook,
         typename Hook::signature_t intercept,
         const char* symname,
+        const SymbolTable& symbols,
         Addr addr,
         bool restore_original)
 {
+    const bool is_arrow_mimalloc_hook = strncmp(symname, "arrow_mi_", 9) == 0;
+    if (is_arrow_mimalloc_hook
+        && (!hook || symbols.getSymbolAddress(symname) != reinterpret_cast<Addr>(hook.d_original)))
```
godlygeek, line 50: AFAICS, `!hook` can never be true. The one caller of this passes `MEMRAY_ORIG(hookname)` as `hook`, and it first calls `MEMRAY_ORIG(hookname).d_symbol` so that must not be null.

src/memray/_memray/hooks.cpp
```diff
@@ -102,6 +102,19 @@ ensureAllHooksAreValid()
 #undef FOR_EACH_HOOKED_FUNCTION
 }
 
+void
+ensureArrowHooksAreValid()
+{
+#if defined(__linux__)
```
godlygeek, line 108: Why can't we interpose these on macOS?

.github/workflows/coverage.yml
pablogsal: Yep, good point. This should be `>=25.0.0` so we also test newer Arrow versions.

src/memray/_memray/elf_shenanigans.cpp
pablogsal: Unless I am missing something, `!hook` checks `d_original`, not `d_symbol`. `d_symbol` is always present, but `d_original` is null while the weak Arrow symbol is unresolved. Without this check we can install an interceptor with a null original.

src/memray/_memray/hooks.cpp
pablogsal: I checked the released macOS wheels. These calls are direct branches inside `libarrow`, not dyld stubs, so Memray has nothing to patch. We need an Arrow linker change first, then Mach-O resolution support in Memray. This PR is Linux-only.

src/memray/_memray/elf_shenanigans.cpp
godlygeek: Ah, I missed the operator overloading that was happening here:
https://github.com/bloomberg/memray/blob/ebf3b278b4e688cea2fba8180a4145c46716cf25/src/memray/_memray/hooks.h#L107-L110

I read this as checking whether a non-null pointer had been passed, but it's checking whether the passed object is or isn't truthy, which is based on whether its `d_original` member is non-null. OK, nevermind me, I just misread this.

godlygeek (approved):
LGTM. I've fixed the coverage to require `arrow>=25.0`, and I tweaked the wording on the news item to something I think users will understand better (from "Track logical allocations" to "Track individual allocations", to make it clearer that what we were seeing previously was the allocations of the arenas and not the individual allocations from those arenas).

Other than that, this LGTM. Thanks @pablogsal!

godlygeek:
I do imagine we might eventually want to be able to interpose these on macOS as well, but if Arrow doesn't support that yet, that's a problem for another day.

Issue #748, opened 2025-04-11
Update the C++ FileSink to fall back to write() if memory mapping isn't supported


godlygeek:
For the sake of performance, memray writes to capture files by mapping the capture file into memory and then writing to it as an in-memory array that is eventually synchronized to the filesystem. (See `FileSink` in sink.cpp / sink.h)

This is fast, but some (uncommon and non-standard) filesystems don't support it.

We should detect if the operations that we need (`posix_fallocate` and `mmap`) aren't supported by the filesystem that the capture file is being written to, and if so log a warning and fall back to a slower but more portable path that simply uses `write()` to append each new chunk of data to the capture file. In order to test this, we'll also need to add a way to force the slower path to be used even if `posix_fallocate` and `mmap` are supported. This should probably just be a `MEMRAY_NO_MMAP=1` environment variable, or something like that.

XueSongTap:
Can I work on this ?

godlygeek:
Sure!

XueSongTap:
# FileSink Fallback Implementation Proposal

## Approach
I propose adding a fallback mechanism when memory mapping isn't supported:

1. **Add detection logic** to check if `posix_fallocate` and `mmap` are supported
2. **Implement environment variable** (`MEMRAY_NO_MMAP=1`) to force fallback path
3. **Create dual write paths** - keep existing mmap path but add write() alternative

## Key Changes

```cpp
class FileSink {
private:
    bool d_useMmap{true}; // Flag to determine which path to use
    // ...
};
```

The constructor would check both environment variable and filesystem support:

```cpp
d_useMmap = !getenv_bool("MEMRAY_NO_MMAP") && testMmapSupport();
```

The `writeAll` method would branch based on this flag:

```cpp
bool FileSink::writeAll(const char* data, size_t length)
{
    return d_useMmap ? writeWithMmap(data, length) : writeWithWrite(data, length);
}
```

## Questions

I'm unsure about naming the fallback write method. Options:
- `writeWithWrite` (too repetitive?)
- `writeFallback`
- `writeTraditional`

@godlygeek  Do you have any preference on the naming convention?

samridh10exe:
@godlygeek Hi, I noticed this issue and the discussion above.
It looks like there is no open PR linked yet, and the last activity was in April.

I am interested in picking this up. My plan is:

Add a FileSink flag that selects mmap vs a write-based fallback

[...]

PR #945, opened 2026-06-08
Issue #748: adding a fallback for file sinks

grantlouisherman:
*Issue number of the reported bug or feature request: #748*

**Describe your changes**
I have added a backup for creating files when mmap or other linux based kernel functions are not available

**Testing performed**
Added two unit tests to test to make sure that if the env variable is set then we fallback to using just `::write`


src/memray/_memray/sink.cpp
```diff
@@ -127,8 +171,33 @@ FileSink::seek(off_t offset, int whence)
         return false;
     }
 
+    // d_fd's kernel file offset is the single source of truth for the current
+    // logical write position, so reposition it first. This is cheap and
+    // should never fail for a regular file.
```
godlygeek, line 176: This assumption is incorrect. There definitely are filesystems for which `seek` isn't supported.

See https://github.com/awslabs/mountpoint-s3/blob/main/doc/SEMANTICS.md for example, for which

> All writes must be sequential: writes after seeking to any offset other than the end of the previous write will fail.

And, we don't actually check that we're writing to a regular file... With `-f`, Memray might be writing to a fifo...

godlygeek (changes requested):
I hate to do this to you, but while reviewing this, I've realized that I really don't want us to automatically fall back to the slower write path, because it will be _so much_ slower.

For a quick test using Python 3.14's own test suite, `memray run --trace-python-allocators --no-compress -fo output.bin -m test test_set` takes about 8 seconds on my machine when `MEMRAY_NO_MMAP=1` isn't set, but about 48 seconds when it is, so this setting is making things take around 6x as long to run, at least for allocation-heavy workloads where Memray has a lot of stuff to write.

I'm thinking that it doesn't make sense for us to automatically fall back to something that makes Memray so much drastically slower, and that instead we should expose a command line argument letting users opt into just using `write()`. I think `FileSink` should just fail with an error message telling people to either write the output file to a different disk or to pass the new command line argument to opt into a slower and less robust method of writing. And I think then we should introduce a new sink class (maybe called `BufferedFileSink` off the top of my head) where `writeAll` just writes to a buffer instead of immediately calling `write()`, and that buffer is only flushed once it fills.

That's already what `SocketSink` does, so we might be able to factor a common base class out that we could share between `SocketSink` and the new `BufferedFileSink`.

I think that approach will be cleaner, will make the tradeoff clearer to users, and will let us get much better performance as well (rearchitecting the existing `FileSink` to buffer when not using mmap but not when using mmap seems ugly and likely to bite us in the future).

Like I said, I hate to do this to you, but I didn't realize how bad this approach was going to be for performance until trying it out.

Would you like to try switching to the approach of plumbing through a different type of sink based on a command line flag, @grantlouisherman, or would you like me to take over driving this?

grantlouisherman:
Hey @godlygeek totally no problem if it’s not useful. Yeah please take it over.

godlygeek:
I wouldn't say "not useful", I'd say it was very useful in proving that my idea about how to solve this was a bad idea 😆 

You delivered exactly what I asked for, but once I had a chance to play with it I could see that what I asked for isn't actually what I needed.

So genuinely, even though I won't be landing this, thanks for doing it - this PR was useful for showing me that I was wrong about how to solve this.

PR #962, opened 2026-07-14
Allow opting out of memory-mapped capture file IO

### LINKED ISSUE #748: Update the C++ FileSink to fall back to write() if memory mapping isn't supported

godlygeek:
Add a `--buffered-file-io` option to `memray run` (and a `buffered=True` argument to `memray.FileDestination`) that writes the capture file using buffered `write` calls instead of the default memory-mapped IO. This is useful on filesystems that don't support shared writable `mmap`, though it's slower and less robust to crashes.

Closes #748

### DIFF
--- a/src/memray/_memray/sink.cpp
+++ b/src/memray/_memray/sink.cpp
@@ -269,23 +269,21 @@ FileSink::~FileSink()
     }
 }
 
-SocketSink::SocketSink(std::string host, uint16_t port)
-: d_host(std::move(host))
-, d_port(port)
+BufferedSink::BufferedSink(size_t bufferSize)
+: BUFFER_SIZE(bufferSize)
 , d_buffer(new char[BUFFER_SIZE])
 , d_bufferNeedle(d_buffer.get())
 {
-    open();
 }
 
 size_t
-SocketSink::freeSpaceInBuffer()
+BufferedSink::freeSpaceInBuffer()
 {
     return BUFFER_SIZE - (d_bufferNeedle - d_buffer.get());
 }
 
 bool
-SocketSink::writeAll(const char* data, size_t length)
+BufferedSink::writeAll(const char* data, size_t length)
 {
     while (freeSpaceInBuffer() < length) {
         size_t toWrite = freeSpaceInBuffer();
@@ -304,19 +302,27 @@ SocketSink::writeAll(const char* data, size_t length)
 }
 
 bool
-SocketSink::flush()
-{
-    return _flush();
-}
-
-bool
-SocketSink::_flush()
+BufferedSink::flush()
 {
     const char* data = d_buffer.get();
     size_t length = d_bufferNeedle - data;
 
     d_bufferNeedle = d_buffer.get();
 
+    return writeBufferedData(data, length);
+}
+
+SocketSink::SocketSink(std::string host, uint16_t port)
+: BufferedSink(PIPE_BUF)
+, d_host(std::move(host))
+, d_port(port)
+{
+    open();
+}
+
+bool
+SocketSink::writeBufferedData(const char* data, size_t length)
+{
     while (length) {
         ssize_t ret = ::send(d_socket_fd, data, length, 0);
         if (ret < 0 && errno != EINTR) {
@@ -348,7 +354,7 @@ SocketSink::cloneInChildProcess()
 SocketSink::~SocketSink()
 {
     if (d_socket_open) {
-        _flush();
+        flush();
         ::close(d_socket_fd);
         d_socket_open = false;
     }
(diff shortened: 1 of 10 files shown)


src/memray/_memray/sink.cpp
```diff
@@ -408,6 +422,75 @@ SocketSink::open()
...
+BufferedFileSink::cloneInChildProcess()
+{
+    std::string file_name = d_fileNameStem + "." + std::to_string(::getpid());
+    return std::make_unique<BufferedFileSink>(file_name, true, d_compress);
+}
+
+BufferedFileSink::~BufferedFileSink()
+{
+    if (d_fd != -1) {
+        flush();
```
lkollar, line 485: Could we log an error if `flush` fails? It could help notice an issue if something like `ENOSPC` occurs as in that case we will compress a truncated file and it could be difficult to figure out what happened.

lkollar (approved):
Just one note, LGTM otherwise.

src/memray/_memray/sink.cpp
godlygeek: This is actually trickier than you'd think. Thinking it through, there's two possibilities: either (1) some earlier write failed first, or (2) the write done by  the destructor was the first to fail.

In case 1 where there were previous failed writes, we should have logged when they occurred, so logging again in the destructor is going to cause duplication and noise, unless we somehow flag when a previous write failed and only log on the first failed write (though that'd require some refactoring because currently the failure logging is done by the tracker, not by the sink itself). Either way, in this case, we don't want to log an error from the destructor.

In case 2 the first write to fail is the one in the `BufferedFileSink` destructor, meaning the tracker finished tracking normally with no indication anything was wrong, and then the `Tracker` destructor fired to end the tracking session. The `Tracker` destructor does this:

https://github.com/bloomberg/memray/blob/8033ae6d628dcb442cf3323b0fa1aeda5f977550/src/memray/_memray/tracking_api.cpp#L932-L933

The last line of that is what destroys the sink (the writer owns the sink), but before destroying the writer and the sink it tries to overwrite the header in the file with updated statistics. That calls into:

https://github.com/bloomberg/memray/blob/8033ae6d628dcb442cf3323b0fa1aeda5f977550/src/memray/_memray/record_writer.cpp#L467-L476

and that call to `BufferedFileSink::seek` will do a flush before seeking if there's any unflushed data:

https://github.com/bloomberg/memray/blob/8033ae6d628dcb442cf3323b0fa1aeda5f977550/src/memray/_memray/sink.cpp#L458-L470

so after that `seek()` finishes there's no buffered data (and per our case 2 premise, when `seek()` called `flush()` it succeeded). Then `writeHeader(true)` goes on to write a new header, which would get buffered - but now it's overwriting data that's already in the file, and so it can't fail with an ENOSPC - the space was already allocated.

I think we could get EIO, ENETDOWN, ENETUNREACH, ENXIO... but I don't think there's any way we could get ENOSPC here without some earlier call failing with ENOSPC first. I think you'd pretty much need to physically unplug the disk we're writing to, or for the network to go down while writing to a network disk, in order for the flush done by seek to succeed but the one done by the destructor to fail a moment later.

About compressing a truncated file: if we've run out of space and so the file is truncated, that most likely will result in `compressFile` failing to write the compressed file out (since it tries to spool it to a temp file on the same disk!) and that will report a reasonably useful "Failed to compress input file" error message to at least signal that something has gone off the rails, even if it doesn't mention that the problem was disk space...

All of which is to say, I'm not going to make this change. It's trickier than you'd think, and adding an error message here without duplicating something we're already reporting to the user elsewhere looks like it would need a pretty big overhaul of our error reporting strategy, for a case that's gonna be really tough to trigger without pausing the process at a breakpoint, unplugging something, and resuming the process.

godlygeek: I'll give you a chance to respond before I merge this - let me know if my logic is unconvincing 😀

lkollar: Makes sense. Sorry for the pain, I was hoping this would be a much simpler change 😅

PR #972, opened 2026-07-22
Avoid duplicating an error message

godlygeek:
When you run `memray --live-remote` with a script that runs indefinitely and then connect to the server from another terminal, the tracker currently writes two different error messages about the broken socket if you quit the GUI. First, it complains when it fails to write some allocation record, and then it complains again when its RSS-tracking background thread fails to write a periodic checkpoint a few milliseconds later.

Make the background thread recognize when we've already encountered a fatal error and stopped tracking, to avoid having it duplicate an error we already displayed.


pradyunsg (approved):

pradyunsg:
It _might_ be worth adding a test, but that's probably also non-trivial for this situation.

godlygeek:
It's a bit tough to add an automated test (testing for the absence of something is hard, doubly so when it's technically timing based), but I did test it manually. And I don't expect that this is something we're likely to regress on.

PR #982, opened 2026-08-03
Record module search paths while tracking

godlygeek:
This allows the `stats` reporter to use the search paths that were active when the tracked process ran, rather than the search paths that are used by the reporter process, for determining a module name and type from its file path. This fixes an issue where we would categorize a module incorrectly if the reporter was run on a different machine or container or from a different virtual environment.


sarahmonod (approved):
LGTM!

ivonastojanovic (approved):
LGTM! Thanks @godlygeek!

PR #983, opened 2026-08-04
Allow control over flame graph source inclusion

godlygeek:
Provide a new command line argument for `memray flamegraph` to control what source code lines, if any, are included in the flame graph.

With `--confidential-files=all`, all files are treated as confidential and so no source code lines are included in the report (only function names, file names, and line numbers, not the text of any source line).

With `--confidential-files=none`, no files are treated as confidential and so lines from any file may be included in the report.

With `--confidential-files=default` (the default) Memray uses heuristics to guess whether each file is likely to contain secrets. It includes source lines from .py files and from any file that is world-readable or executable, and excludes lines from other files.


src/memray/reporters/frame_tools.py
```diff
@@ -80,3 +84,44 @@ def is_frame_from_import_system(frame: StackFrame) -> bool:
...
+    #
+    # We assume that Python source files won't contain keys, certificates, or
+    # other tokens, but other files might. We assume that a file with a `.py`
+    # extension is a Python source file, and that an executable file is likely
+    # to be a Python source file too. We assume that any world-readable file
+    # will not contain secrets.
+    #
+    # These heuristics may not always be correct, and flame graphs may wind up
+    # having secrets embedded in them. Don't share a Memray flame graph with
+    # someone unless you trust them.
```
pradyunsg: Why not make this a docstring?

pradyunsg (approved):
LGTM, with a trivial comment.

PR #986, opened 2026-08-05
ci: Publish Python 3.15 wheels

godlygeek:


tests/integration/test_tracking.py
```diff
@@ -128,15 +128,6 @@ def test_simple_cpp_allocation_tracking(tmp_path):
...
     allocator_func, allocator_type, domain, tmp_path
 ):
-    # TODO: Remove this entire check once we stop using 3.15.0b2
-    if (
-        sys.version_info >= (3, 15)
-        and sys.version_info <= (3, 15, 0, "beta", 2)
-        and domain == PymallocDomain.PYMALLOC_OBJECT
-        and allocator_func == "realloc"
-    ):
-        pytest.skip("Skipped for https://github.com/python/cpython/issues/151297")
```
sarahmonod, line 138: Is it worth making this check more generic? As in, something like this?

```suggestion
    if (
        sys.version_info[2:] <= (0, "beta", 2)
        and domain == PymallocDomain.PYMALLOC_OBJECT
        and allocator_func == "realloc"
    ):
        pytest.skip("Skipped as the ABI is not yet frozen")
```

sarahmonod: Has nothing to do with ABI stability after all, ignore me

sarahmonod (approved):
LGTM!

PR #988, opened 2026-08-06
Add support for the new C23 deallocators

### LINKED ISSUE #725: Intercept new `free_sized` and `free_aligned_sized` functions

godlygeek:
Record calls to the `free_sized` and `free_aligned_sized` deallocators introduced in C23.

Instead of delegating to the original functions, we delegate to the `free` function for both. This is a legal implementation of these new deallocators, and simplifies our implementation since these functions may not always be available.

Closes #725 

This is much of the code from #817, but updated to interpose the C23 deallocators even when Memray itself wasn't built with C23, and to test these changes through the `memray._test_utils` extension module instead of adding something to one of the hand written extension modules in our test suite.


sarahmonod (approved):

godlygeek:
Thanks for the contribution, @naamanu!

PR #989, opened 2026-08-07
ci: Build wheels on macOS 26

godlygeek:


godlygeek:
I tried this seeing if it would make CI run any faster (since the macOS wheel builds are the slowest thing CI does), but it seems like it doesn't make much of a difference.

PR #990, opened 2026-08-07
Guide users towards overwriting files

### LINKED ISSUE #828: Overwriting output files if they already exist

- [x] I have searched the existing proposals

If I put something like:

```
with memray.Tracker("out.bin"):
    .... # do thing
```

in a script, the first time I run it `memray` will glady write out to `out.bin`. The second time I run it, `memray` complains about `out.bin` already existing. So I need to throw in a `Path("out.bin").unlink(error_ok=True)` to have my script run the first and subsequent times.

This is also true on CLI invocations like `memray run -o out.bin ...` and `memray flamegraph -o report.html ...`

If I specify an output file that already exists, I would expect it to "just" overwrite it! This is how many other "write profiling/debugging results to disk" utilities work, generally on the assumption that you run it over and over again.

"I have accidentally stated an existing file" seems like a rare occurence compared to "I would like to write to the same path with fresh data". To me anyways.

Having some flags would also do the trick for me though it seems like "learning about the flag and using it" is its own little set of weirdness

godlygeek:
Users may not know that the `--force` option exists (or its equivalent `overwrite=True` API argument for `memray.Tracker`). Try to point them to it in our error messages to help them out.

Closes #828

### DIFF
--- a/src/memray/_memray/sink.cpp
+++ b/src/memray/_memray/sink.cpp
@@ -137,6 +137,12 @@ FileSink::FileSink(const std::string& file_name, bool overwrite, bool compress)
         d_fd = ::open(file_name.c_str(), flags, 0644);
     } while (d_fd < 0 && errno == EINTR);
     if (d_fd < 0) {
+        if (errno == EEXIST) {
+            throw IoError{
+                    "Output file " + file_name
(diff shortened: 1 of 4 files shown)


sarahmonod (approved):
LGTM

PR #994, opened 2026-08-10
Ensure user code never sees our 3rd party imports

godlygeek:
Import `jinja2`, `rich`, and `textual` lazily so that they aren't loaded into `sys.modules` before running user code in `memray run` or `memray attach`.


godlygeek:
@lkollar This lays the groundwork for what I was describing in https://github.com/bloomberg/memray/pull/893#issuecomment-4963318239 (and is something I've wanted to clean up for a long time, anyway...)

PR #1000, opened 2026-08-18
Use PyEval_SetProfileAllThreads on Python 3.12

### LINKED ISSUE #999: Segmentation fault on python 3.12

- [x] I have searched the existing issues

### Current Behavior

Experienced segfault using `Memray.Tracker` in one of my tests (python 3.12). The segfaults were flaky, only happening ever 5th or so test run. Using GDB i was able to retrieve this stacktrace:
```bash
(gdb) bt
#0  0x00007ffff76da7a5 in Py_XDECREF.lto_priv.91 () from /lib64/libpython3.12.so.1.0
#1  0x00007ffff766697c in _PyEval_SetProfile () from /lib64/libpython3.12.so.1.0
#2  0x00007fffe33eabad in memray::compat::setprofileAllThreads(int (*)(_object*, _frame*, int, _object*), _object*) [clone .constprop.0] (func=func@entry=0x0, arg=0x0) at /project/src/memray/_memray/compat.cpp:17
#3  0x00007fffe3375590 in memray::tracking_api::PythonStackTracker::removeProfileHooks () at /project/src/memray/_memray/tracking_api.cpp:924
#4  memray::tracking_api::Tracker::~Tracker (this=<optimized out>, this=<optimized out>) at /project/src/memray/_memray/tracking_api.cpp:924
#5  0x00007fffe33a7935 in std::default_delete<memray::tracking_api::Tracker>::operator() (this=<optimized out>, __ptr=0x55555b142940) at /opt/rh/gcc-toolset-14/root/usr/include/c++/14/bits/unique_ptr.h:87
#6  std::default_delete<memray::tracking_api::Tracker>::operator() (this=<optimized out>, __ptr=0x55555b142940) at /opt/rh/gcc-toolset-14/root/usr/include/c++/14/bits/unique_ptr.h:87
#7  std::__uniq_ptr_impl<memray::tracking_api::Tracker, std::default_delete<memray::tracking_api::Tracker> >::reset (this=<optimized out>, __p=0x0) at /opt/rh/gcc-toolset-14/root/usr/include/c++/14/bits/unique_ptr.h:205
#8  std::unique_ptr<memray::tracking_api::Tracker, std::default_delete<memray::tracking_api::Tracker> >::reset (this=<optimized out>, __p=0x0) at /opt/rh/gcc-toolset-14/root/usr/include/c++/14/bits/unique_ptr.h:504
#9  memray::tracking_api::Tracker::destroyTracker () at /project/src/memray/_memray/tracking_api.cpp:1508
#10 __pyx_pf_6memray_7_memray_7Tracker_4__exit__ (__pyx_v_exc_type=<optimized out>, __pyx_v_exc_value=<optimized out>, __pyx_v_exc_traceback=<optimized out>, __pyx_v_self=0x7fff0cf76260) at /tmp/tmp8b2puaxm/build/_memray.cpp:22523
#11 __pyx_pw_6memray_7_memray_7Tracker_5__exit__ (__pyx_v_self=0x7fff0cf76260, __pyx_args=<optimized out>, __pyx_nargs=<optimized out>, __pyx_kwds=<optimized out>) at /tmp/tmp8b2puaxm/build/_memray.cpp:22404
```
Digging around  found this [PR](https://github.com/bloomberg/memray/pull/622/changes#diff-90f88377015b1982adcf9d23e50c159aaeadfe87b5bb84cc6236bf9bc1f612e6R9-R40) where the guard clauses for using CPython's `PyEval_SetProfileAllThreads`  is set at `python>=3.13`. This API was however introduced in `python 3.12`.

I assume the guard close being set to this python version was an oversight and suggest updating it to allow `python 3.12`.

### Expected Behavior

No spurious segmentation faults.

### Steps To Reproduce

Environment: memray 1.20.0, CPython 3.12.x (RHEL 8 system build, LTO), Linux x86_64.

Reproducing should only require a pool of threads that repeatedly start()/join() trivial targets, a few hundred idle threads to lengthen the thread-state list, and a loop that enters and exits a Tracker as fast as possible. This is a data race and probably platform specific, I am still not able to recreate this macOS/arm64 with python 3.12

### Memray Version

1.20.0

### Python Version

3.12

### Operating System

Linux

### Anything else?

_No response_

eqbech:
resolves: #999

**Describe your changes**
Updated to now use `SetProfileAllThreads` for `python>=3.12`. As this is when the method became available in CPython.

- https://docs.python.org/3/whatsnew/3.12.html
- https://github.com/python/cpython/issues/93503

### DIFF
--- a/news/1000.bugfix.rst
+++ b/news/1000.bugfix.rst
@@ -1,0 +1,1 @@
+Use the public ``PyEval_SetProfileAllThreads`` API on Python 3.12, rather than walking the interpreter's thread list ourselves and calling a private CPython function for each thread state.
--- a/src/memray/_memray/compat.cpp
+++ b/src/memray/_memray/compat.cpp
@@ -6,7 +6,7 @@ void
 setprofileAllThreads(Py_tracefunc func, PyObject* arg)
 {
     assert(PyGILState_Check());
-#if PY_VERSION_HEX >= 0x030D0000
+#if PY_VERSION_HEX >= 0x030C00F0
     PyEval_SetProfileAllThreads(func, arg);
 #else
     PyThreadState* this_tstate = PyThreadState_Get();


pablogsal (approved):

pablogsal:
@eqbech can you add a small news entry?

eqbech:
> @eqbech can you add a small news entry?

I put it under **bugfix** as the issue it resolves is marked as a **bug**. Let me know if you would rather have it elsewhere 😄

pablogsal (approved):

pablogsal:
Thanks a lot for the fix @eqbech!

PR #1003, opened 2026-08-21
Reduce tracking overhead with sys.monitoring

pablogsal:
On Python 3.12 and newer, use `sys.monitoring` instead of the legacy profile hook to maintain Memray's Python shadow stack. The profile hook makes CPython materialize a `PyFrameObject` and enter the general profiling callback for every Python call and return, and it also reports C-call events that Memray ignores. `sys.monitoring` passes the existing code object and lets Memray subscribe only to the six Python events that can change the stack. Removing that work from every function transition makes the benchmark workloads below 20% to 43% faster while tracking. If the monitoring slot is occupied, Memray keeps using the profile hook as a fallback.

Allocator hooks validate the shadow stack through CPython's interpreter bookkeeping, avoiding calls that can allocate or execute Python while an allocation is already being recorded. This keeps the hot path self-contained and lets Memray discard stale shadow stacks when monitoring is disabled or the tool slot changes ownership.

| Workload | Source | Mode | Faster |
|---|---|---|---:|
| Async CPU | benchmarks/ | Normal | 30.0% |
| Async CPU | benchmarks/ | Native | 29.9% |
| Async I/O | benchmarks/ | Native | 23.2% |
| Async memoization | benchmarks/ | Normal | 28.9% |
| Async memoization | benchmarks/ | Native | 27.0% |
| Async mixed | benchmarks/ | Native | 20.7% |
| Fannkuch | benchmarks/ | Normal | 24.7% |
| Fannkuch | benchmarks/ | Native | 24.0% |
| Pretty-print | benchmarks/ | Normal | 42.4% |
| Pretty-print | benchmarks/ | Native | 42.7% |

### DIFF
--- a/src/memray/_memray/compat_internal.c
+++ b/src/memray/_memray/compat_internal.c
@@ -1,0 +1,108 @@
+#define PY_SSIZE_T_CLEAN
+#include <patchlevel.h>
+
+#if PY_VERSION_HEX >= 0x030C0000
+#    define Py_BUILD_CORE_MODULE
+#endif
+#include <Python.h>
+
+#if PY_VERSION_HEX >= 0x030C0000
+#    include "internal/pycore_frame.h"
+#    if PY_VERSION_HEX >= 0x030E0000
+#        include "internal/pycore_interp_structs.h"
+#        include "internal/pycore_interpframe.h"
+#    else
+#        include "internal/pycore_interp.h"
+#    endif
+#    undef Py_BUILD_CORE
+#    undef Py_BUILD_CORE_MODULE
+#endif
+
+#include "compat_internal.h"
+
+int
+memray_compat_is_current_or_caller_frame(PyFrameObject* frame)
+{
+#if PY_VERSION_HEX >= 0x030C0000
+    PyThreadState* tstate = PyGILState_GetThisThreadState();
+    if (!tstate) {
+        return 0;
+    }
+
+#    if PY_VERSION_HEX < 0x030D0000
+    _PyInterpreterFrame* current = tstate->cframe ? tstate->cframe->current_frame : NULL;
+#    else
+    _PyInterpreterFrame* current = tstate->current_frame;
+#    endif
+    current = _PyFrame_GetFirstComplete(current);
+    if (!current) {
+        return 0;
+    }
+    if (current->frame_obj == frame) {
+        return 1;
+    }
+    _PyInterpreterFrame* caller = _PyFrame_GetFirstComplete(current->previous);
+    return caller && caller->frame_obj == frame;
+#else
+    (void)frame;
+#endif
+    return 0;
+}
+
+int
+memray_compat_is_parent_frame(PyFrameObject* parent, PyFrameObject* frame)
+{
+#if PY_VERSION_HEX >= 0x030C0000
+    _PyInterpreterFrame* previous = frame->f_frame->previous;
+    previous = _PyFrame_GetFirstComplete(previous);
+    return previous && previous->frame_obj == parent;
+#else
+    (void)parent;
+    (void)frame;
+#endif
+    return 0;
+}
+
+int
+memray_compat_is_monitoring_tool_active(int tool_id, PyObject* tool_name, PyObject* callbacks)
+{
+#if PY_VERSION_HEX >= 0x030C0000
+    static const int events[] = {
+            PY_MONITORING_EVENT_PY_START,
+            PY_MONITORING_EVENT_PY_RESUME,
+            PY_MONITORING_EVENT_PY_RETURN,
+            PY_MONITORING_EVENT_PY_YIELD,
+            PY_MONITORING_EVENT_PY_UNWIND,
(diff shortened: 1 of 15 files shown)


src/memray/_memray/compat_internal.c
```diff
@@ -0,0 +1,108 @@
...
+    const uint8_t tool = 1U << tool_id;
+    const _Py_GlobalMonitors* monitors = &interp->monitors;
+    for (Py_ssize_t i = 0; i < event_count; ++i) {
+        if (interp->monitoring_callables[tool_id][events[i]] != PyTuple_GET_ITEM(callbacks, i)
+            || !(monitors->tools[events[i]] & tool))
+        {
+            return 0;
+        }
+    }
+    return 1;
```
godlygeek, line 101: Why can't we just call `sys.monitoring.get_tool(2)` and check if it returns non-`None`? This looks like we're looking at a whole lot of interpreter privates to get something that's exposed through public APIs...

src/memray/_memray/tracking_api.cpp
```diff
@@ -442,29 +452,54 @@ PythonStackTracker::populateShadowStack()
...
-PythonStackTracker::handleTraceEvent(int what, PyFrameObject* frame)
+PythonStackTracker::handleProfileEvent(int what, PyFrameObject* frame)
 {
-    installGreenletTraceFunctionIfNeeded();
+    if (what == PyTrace_CALL) {
+        if (d_stack && !d_stack->empty() && d_stack->back().isCurrentFrame(frame)) {
+            d_stack->back().ignoreProfileCall();
+            return;
+        }
+        handlePush(frame);
```
godlygeek, line 516: Claude caught what I'm pretty sure is a real bug here:

> The duplicate check runs against `d_stack->back().d_frame`, but a frozen frame has `d_frame == nullptr` (`isFrozen()` is literally `d_frame == nullptr`). So if a thread's first profile event after a new Tracker is installed is one of these Cython duplicate CALLs, the check misses: `handlePush` → `replaceFrozenStackIfNeeded()` rebuilds the stack and returns early (so no double push), but nothing records the ignore — and the matching Cython `RETURN` then pops a live frame.

That seems accurate to me - we can't rely on `d_stack->back()` until after we've replaced the frozen stack with a "real" stack that reflects what profile functions would see, but we're checking it before (we're checking in `isCurrentFrame`, and the replacement with a "real" stack happens inside `handlePush`).

godlygeek (commented):
I want to like this PR, but this is so many lines of new, complex code, in areas that we've had lots of bugs in over the years, _and_ we get to keep both implementations indefinitely. Ugh.

I wonder whether we can come up with any ways to simplify this, or to rely less on CPython internals...

I _really_ wish we had a better way to make Cython stop being weird... or some better way to detect the misleading frames that Cython is giving us...

pablogsal:
> I want to like this PR, but this is so many lines of new, complex code, in areas that we've had lots of bugs in over the years, _and_ we get to keep both implementations indefinitely. Ugh.
> 
> I wonder whether we can come up with any ways to simplify this, or to rely less on CPython internals...
> 

Yeah I feel you. To be honest we can only support this in the future once 3.12 is the minimum supported version but I am with you your concerns are real.

What do you think we should do with this then?

godlygeek:
I'm not sure, I need to think about it...

On the one hand, this is a lot of complexity, but on the other hand it leaves a lot of performance on the table if we don't merge it. I'm tempted to spend some time on it to see if I can find any ways to simplify and shed some complexity, but realistically I think I've got higher priorities right now and can't justify the effort.

I'm leaning towards not landing this in its current form just because I'm worried about how difficult it will be to review and support, and the likelihood that there's very subtle bugs lurking in it. But, a 40% speedup is _huge_... 😔

If we can find ways to reduce the matrix of weird cases that we need to support, I'd be a lot more comfortable, but with

- `sys.setprofile` vs `sys.monitoring`
- Cython's `CYTHON_USE_SYS_MONITORING=0` vs `CYTHON_USE_SYS_MONITORING=1`
- the legacy implementation of profile hooks from 3.11 and earlier vs the compatibility shim for them implemented in terms of `sys.monitoring` in 3.12+
- stacks populated by walking backwards from `PyThreadState_GetFrame` vs ones built incrementally by the Python VM reporting calls
- normal Python frames vs Cython's weird-and-lying profiling support

there's just a ton of different factors here, that interact with each other in nontrivial ways... And both of those latter two are places where we've had extremely subtle bugs in the past.

src/memray/_memray/tracking_api.cpp
pablogsal: Yep, good catch. Fixed it and added regression tests.

src/memray/_memray/compat_internal.c
pablogsal: Switched the parent lookup to PyFrame_GetBack. We still need the direct monitoring checks here since get_tool could allocate inside the allocation hook.

PR #1006, opened 2026-08-28
Add --split-threads option to memray transform csv

### LINKED ISSUE #996: Add --split-threads support to memray transform csv

- [x] I have searched the existing proposals

`memray transform csv` includes `tid` and `thread_name` columns in its output schema, but the transform command currently retrieves allocation records with threads merged.

When allocations with the same stack occur on multiple threads, the exported record can therefore contain `tid = -1` and `thread_name = "merged thread"` instead of preserving the original thread identities.

This makes the CSV output difficult to use for per-thread analysis of multi-threaded Python or C++ workloads. It also differs from reporters such as flamegraph and table, which expose a `--split-threads` option.

Add a `--split-threads` option to `memray transform`, initially at least for the CSV format.

When enabled for CSV output, the command would retrieve allocation records with `merge_threads=False` and allow `TransformReporter` to process those records. The existing `tid` and `thread_name` columns would then preserve the original thread information.

The default behavior would remain unchanged.

Would you prefer `--split-threads` to be:

1. supported for CSV only, since its schema already represents thread identity; or
2. supported across transform formats, with Speedscope potentially emitting separate profiles for each thread?

I would be happy to implement the narrower CSV-only version first, including CLI integration tests and reporter tests, if that is the preferred scope.

Users can inspect per-thread information through the HTML reporters, but those reports do not provide a convenient machine-readable export for downstream analysis.

BhargavKumarNath:
Closes: #996

## Summary

`memray transform csv` has `tid` and `thread_name` columns, but `memray transform` always retrieved records with threads merged, so allocations sharing a call stack across multiple threads collapsed into one row attributed to a single representative thread (with summed `num_allocations` / `size`), making per-thread analysis of the CSV impossible.

This adds a `--split-threads` option matching the flamegraph and table reporters. When passed, records are retrieved with `merge_threads=False` and each thread's allocations stay on their own row with that thread's real `tid` and `thread_name`.

Scope is CSV-only, as offered in the issue: `--split-threads` with `gprof2dot` or `speedscope` exits with an error, since those formats aggregate purely by call stack and have no per-thread dimension.

## Changes

- `commands/transform.py`: new `--split-threads` flag; rejected for non-csv formats
- `reporters/transform.py`: `TransformReporter.render` no longer blocks unmerged threads for the csv format
- Unit tests (`test_transform_reporter.py`) + CLI integration tests (`test_main.py`, including a multithreaded fixture)
- `docs/transform.rst`: documents the option and the merged-thread default
- `news/996.feature.rst`: changelog fragment

## Testing

- `python -m pytest tests/unit tests/integration/test_main.py -k Transform`
- `prek run --all-files`, `mypy -p memray --strict`, `sphinx -b html -W docs`

### DIFF
--- a/docs/transform.rst
+++ b/docs/transform.rst
@@ -98,6 +98,12 @@ The available columns are:
   first), where each stack frame in the list has the following format:
   ``<function_name>;<file_name>;<line_number>``.
 
+Unlike memray's human-readable reporters, the ``csv`` format does not merge
+allocations across threads. Two threads that allocate memory at the same call
+stack produce two separate rows, each carrying that thread's own ``tid`` and
+``thread_name``, so the output can be grouped or filtered by thread with other
+data analysis tools.
+
 CLI Reference
 -------------
 
--- a/news/996.feature.rst
+++ b/news/996.feature.rst
@@ -1,0 +1,1 @@
+``memray transform csv`` now keeps allocations separated by thread instead of merging them across threads, so the ``tid`` and ``thread_name`` columns always identify the thread that performed each allocation. This makes the CSV output easier to group and filter by thread. Other ``memray transform`` formats are unaffected and continue to merge allocations across threads.
--- a/src/memray/commands/transform.py
+++ b/src/memray/commands/transform.py
@@ -38,6 +38,12 @@ class TransformCommand(HighWatermarkCommand):
                 f"Format not supported: {args.format}", exit_code=1
             )
 
+        # The csv format exists to be machine-parsed, and it has dedicated tid
+        # and thread_name columns, so we keep each thread's allocations on their
+        # own row instead of merging them across threads. Every other format is
+        # meant for human consumption and always merges threads.
+        args.split_threads = the_format == "csv"
+
         self.suffix = suffix
         self.reporter_name = the_format
         super().run(args, parser)
--- a/src/memray/reporters/transform.py
+++ b/src/memray/reporters/transform.py
@@ -211,8 +211,11 @@ class TransformReporter:
         inverted: bool,
         no_web: bool = False,
     ) -> None:
-        if not merge_threads:
-            raise NotImplementedError("TransformReporter only supports merged threads.")
+        if not merge_threads and self.format != "csv":
+            raise NotImplementedError(
+                "TransformReporter does not support split threads for the"
+                f" {self.format!r} format."
+            )
         if inverted:
             raise NotImplementedError(
                 "TransformReporter does not support inverted argument."
--- a/tests/integration/test_main.py
+++ b/tests/integration/test_main.py
@@ -1,4 +1,5 @@
 import contextlib
+import csv
 import json
 import os
 import platform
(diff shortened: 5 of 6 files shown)


docs/transform.rst
```diff
@@ -88,16 +88,31 @@ The available columns are:
...
-  location by this thread and not deallocated before the high water mark.
+  location and not deallocated before the high water mark.
 * ``size``: the total size in bytes of the allocations performed at this
-  location by this thread and not deallocated before the high water mark.
-* ``tid``: the thread id of the thread that performed the allocations.
-* ``thread_name``: the name of the thread that performed the allocations.
+  location and not deallocated before the high water mark.
+* ``tid``: the thread id of the thread that performed the allocations. When
+  allocations from multiple threads are merged (the default; see
+  ``--split-threads`` below), this is the id of one of those threads.
```
godlygeek: Hm, picking a random one of those threads to show its ID isn't ideal... in other reporters we've just dropped the column when threads are merged, but doing that with CSV is sort of risky since it's likely to be machine processed - I think we'd have to consider it a backwards incompatible change if we did that, while we've gotten away with it for reports that are meant to be human-readable...

I wonder if maybe we should do a space-separated list of thread ids for all of the threads we saw the allocation in? Or, honestly, maybe we should just leave the column blank - keep it in the file for backwards compatibility, but don't put any value in it because it doesn't make sense anymore...

Actually, I think I want to take a different approach entirely here: I think we shouldn't add a `--split-threads` option at all, and we should instead change the _default_ to split threads. As #996 says, the point of the CSV output format is to be machine-parseable, and so I think merging allocations from different threads by default is worse than leaving them separate, especially given that the documentation above says that we've always meant to include the tid and thread name in the CSV output...

@pablogsal What do you think?

godlygeek: OK, @BhargavKumarNath would you mind changing the PR to tackle it that way instead? It seems like the best path forward to me.

BhargavKumarNath:
Thanks @godlygeek, that reasoning makes sense to me. Since the CSV output is meant to be
fed into other tools rather than read by a person, collapsing several threads
into one row with a `tid`/`thread_name` that only represents one of them is
more surprising than helpful, and it never matched what the column docs
describe. Making split-threads the default is the cleaner fix.

I've reworked the PR along those lines:

- Dropped the `--split-threads` option entirely. `memray transform csv` now
  retrieves records with `merge_threads=False` by default, so each thread's
  allocations stay on their own row with that thread's real `tid` and
  `thread_name`.
- `gprof2dot` and `speedscope` are untouched, they have no per-thread
  dimension, so they keep merging.
- Updated `docs/transform.rst` and the news fragment, and reworked the tests:
  the reporter/CLI tests now assert the split-by-default behavior, plus a check
  that `--split-threads` is no longer a recognized argument.

mylee04:
Thanks for implementing #996! I like the split-by-default approach. Would it be useful to give the worker threads explicit names and also check their exported names and allocation sizes in the integration test? That seems like a helpful end-to-end check for the thread identity use case behind the issue.

BhargavKumarNath:
@mylee04 Good suggestion. I've updated the integration test so each worker thread now has an explicit name (`worker-0`, `worker-1`, `worker-2`). The test also verifies both the exported thread name and allocation size for each worker's CSV row, rather than only checking that the workers appear on separate rows.

PR #1008, opened 2026-08-28
Scan capture file padding in blocks

pablogsal:
Uncompressed captures written through the mmap-backed file sink are grown ahead of the write position with posix_fallocate. That leaves a zero-filled tail after the trailer. FileSource supports interrupted captures by finding the last non-zero byte, but currently does that with one peek and seek per byte of padding.

I reproduced this from a by recording ten million 64-byte malloc/free pairs at the same call site, without compression:

    python -m memray run --no-compress -q -o /tmp/memray-padding.bin -c 'from collections import deque; from memray._test_utils import MemoryAllocator; a = MemoryAllocator(); deque(((a.malloc(64), a.free()) for _ in range(10_000_000)), maxlen=0)'

That produced a 76,042,240-byte capture containing about 20 million allocation records. The valid prefix ended at byte 70,008,450, leaving 6,033,790 zero bytes. I timed reader initialization, which is where the padding scan happens, with:

    /usr/bin/time -f '%e seconds' python -c "from memray import FileReader; FileReader('/tmp/memray-padding.bin')"

On the base branch that took 4.04 and 4.33 seconds. With this change it took 2.24 and 2.26 seconds. Removing the padding from the benchmark file makes both cases take about 2.2 seconds, so the difference is specifically the tail scan rather than parsing.

### DIFF
--- a/src/memray/_memray/source.cpp
+++ b/src/memray/_memray/source.cpp
@@ -1,6 +1,7 @@
 #define PY_SSIZE_T_CLEAN
 #include <Python.h>
 
+#include <array>
 #include <cerrno>
 #include <cstring>
 #include <iostream>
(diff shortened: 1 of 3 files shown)


godlygeek:
I'm not against making this change, _but_ I think we should make the changes in #1012 first, and then see if you still think this change is needed. If we make the changes in #1012, then the change in this PR only helps in the case where the process died while tracking without a graceful shutdown, and the maximum savings we can get from searching for the end of the padding in chunks is smaller than what you observed here (because the maximum amount of padding will be capped to less than you observed here).

godlygeek:
Also, if we land #1012, this should be changed to only scan in blocks if the last byte of the file is 0. If it's non-zero, we've already truncated, and so reading 64kib is just a waste.

Do you still think this optimization makes sense to do if it's only used in the case where tracking didn't end gracefully?

src/memray/_memray/source.cpp
```diff
@@ -87,16 +176,35 @@ FileSource::findReadableSize()
     // in order to recover from the file truncation. To ignore these, we count
     // the zeroed bytes at the end of the file, and make calls to read() and
     // getline() fail if they read into those bytes.
-    d_raw_stream->seekg(-1, d_raw_stream->end);
-    while (*d_raw_stream) {
-        char c = d_raw_stream->peek();
-        if (c != 0x00) {
-            d_readable_size = d_raw_stream->tellg() + std::streamoff(1);
+    constexpr size_t buffer_size = 64 * 1024;
+    std::array<char, buffer_size> buffer;
```
godlygeek, line 180: Huh, do you really want to put 64kib on the stack? I'd think it's probably better to use a vector here... That's a pretty big buffer to put on the stack, IMO

pablogsal: Yep, switched to a vector. It only grows if there’s padding.

PR #1009, opened 2026-08-28
Aggregate temporary allocations while reading

pablogsal:
Temporary-allocation reports currently keep every matching allocation until the whole capture has been read, then aggregate them by call site. That makes memory usage grow with the number of events even when the final report contains one row.

I reproduced this with ten million 64-byte `malloc`/`free` pairs from the same Python call site (twenty million records). The capture is uncompressed and uses buffered file IO so mmap padding is not part of this measurement:

```console
python -m memray run --buffered-file-io --no-compress -q -o /tmp/temporary.bin -c 'from memray._test_utils import MemoryAllocator; from collections import deque; allocator = MemoryAllocator(); deque(((allocator.malloc(64), allocator.free()) for _ in range(10_000_000)), maxlen=0)'
/usr/bin/time -f 'elapsed=%e maxrss=%M' python -m memray table --temporary-allocations -o /tmp/temporary.html /tmp/temporary.bin
```

The 67 MiB capture takes 12.27s and 1.18 GiB peak RSS on `main`, versus 11.39s and 55 MiB here.

This aggregates matches as they are found and retains only one record and ordering entry per unique call site, while preserving split/merged thread behavior and capture order.

### DIFF
--- a/src/memray/_memray/snapshot.cpp
+++ b/src/memray/_memray/snapshot.cpp
@@ -224,7 +224,17 @@ TemporaryAllocationsAggregator::addAllocation(const Allocation& allocation)
                     });
 
             if (alloc_it != it->second.end()) {
-                d_temporary_allocations.push_back(*alloc_it);
+                const Allocation& record = alloc_it->allocation;
+                auto loc_key = LocationKey{record.frame_index, record.native_frame_id, record.tid};
+                auto [temporary_it, inserted] = d_temporary_allocations.emplace(loc_key, *alloc_it);
+                if (inserted) {
+                    d_temporary_allocation_order.push_back(loc_key);
+                } else {
+                    temporary_it->second.allocation.size += record.size;
+                    temporary_it->second.allocation.n_allocations += record.n_allocations;
+                    temporary_it->second.sequence_number =
+                            std::min(temporary_it->second.sequence_number, alloc_it->sequence_number);
+                }
             }
             break;
         }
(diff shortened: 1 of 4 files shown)


src/memray/_memray/snapshot.cpp
```diff
@@ -235,18 +245,22 @@ TemporaryAllocationsAggregator::addAllocation(const Allocation& allocation)
...
 {
+    if (!merge_threads) {
+        return d_temporary_allocations;
+    }
+
     reduced_snapshot_map_t stack_to_allocation{};
 
-    for (const auto& snapshot_allocation : d_temporary_allocations) {
+    for (const auto& key : d_temporary_allocation_order) {
+        const auto& snapshot_allocation = d_temporary_allocations.at(key);
```
godlygeek, line 258: Either I'm missing something, or this doesn't make sense. `reduced_snapshot_map_t` is an unordered map, so I don't understand why you're going out of your way to populate it in the same order as `d_temporary_allocations` was populated. You added a `test_temporary_allocations_remain_in_capture_order` test, but that test still passes for me if I reverse the order of this for loop and instead do:
```cpp
    for (auto it = d_temporary_allocation_order.rbegin(), end = d_temporary_allocation_order.rend(); it != end; ++it) {
        const auto& key = *it;
```
I think you're just testing implementation details of the hash table?

pablogsal: > `reduced_snapshot_map_t` is an unordered map, so I don't understand why you're going out of your way to populate it in the same order as `d_temporary_allocations` was populated.

Yeah. The idea here was that this was for when `merge_threads` removes the thread from the key and several entries collapse into one. The first entry supplies the full allocation, while the others only update the size, count, and sequence number, so the order decides which allocator and TID we keep.

> that test still passes for me if I reverse the order of this for loop

Yeah, this was supposed to exercise the thread thing, but it only uses one thread, so it doesn’t test that at all. The final list is sorted by `sequence_number`, which is why I assume reversing the loop still passes.

> I think you're just testing implementation details of the hash table?

Yeah let me clean this up a bit

godlygeek (approved):
LGTM

PR #1011, opened 2026-08-28
Buffer capture file parsing

pablogsal:
Capture parsing currently routes every tiny read through the stream interface. Large capture files contain billions of small fields, so that repeated stream setup becomes a significant part of report generation.

This keeps a 64 KiB buffer in `FileSource` and serves reads and delimited strings from memory, refilling from the underlying stream only when the buffer is exhausted. Varints are decoded directly from those buffered bytes instead of calling `read()` once per byte. Socket reads keep the existing byte-by-byte fallback, while compressed files use the same buffered path over the decompressed stream.

On a 17.54 GB capture containing 3.38 billion allocation events, this reduced flamegraph generation from 7m38.7s to 4m23.6s: 42.5% less time, or 1.74x faster. Peak RSS stayed effectively unchanged at about 376 MB.

### DIFF
--- a/src/memray/_memray/record_reader.cpp
+++ b/src/memray/_memray/record_reader.cpp
@@ -126,25 +126,7 @@ RecordReader::readHeader(HeaderRecord& header)
 bool
 RecordReader::readVarint(uint64_t* val)
 {
-    *val = 0;
-    int shift = 0;
-
-    while (true) {
-        unsigned char next;
-        if (!d_input->read(reinterpret_cast<char*>(&next), sizeof(next))) {
-            return false;
-        }
-
-        *val |= (static_cast<uint64_t>(next & 0x7f) << shift);
-        if (0 == (next & 0x80)) {
-            return true;
-        }
-
-        shift += 7;
-        if (shift >= 64) {
-            return false;
-        }
-    }
+    return d_input->readVarint(val);
 }
 
 bool
--- a/src/memray/_memray/source.cpp
+++ b/src/memray/_memray/source.cpp
@@ -42,30 +42,69 @@ FileSource::FileSource(const std::string& file_name)
 }
 
 bool
-FileSource::read(char* stream, ssize_t length)
+FileSource::refillBuffer()
 {
-    if (d_stream->read(stream, length).fail()) {
+    if (d_buffer_pos != d_buffer_end) {
+        return true;
+    }
+
+    std::streamsize bytes_to_read = d_buffer.size();
+    if (d_readable_size) {
+        if (d_bytes_read >= d_readable_size) {
+            return false;
+        }
+        bytes_to_read =
+                std::min(bytes_to_read, static_cast<std::streamsize>(d_readable_size - d_bytes_read));
+    }
+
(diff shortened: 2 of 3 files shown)


godlygeek:
> On a 17.54 GB capture containing 3.38 billion allocation events, this reduced flamegraph generation from 7m38.7s to 4m23.6s: 42.5% less time, or 1.74x faster.

Whoa, that's huge!

> Socket reads keep the existing byte-by-byte fallback

Why is that difference necessary?

> Varints are decoded directly from those buffered bytes instead of calling `read()` once per byte.

Huh - why does this matter? It seems like it should be equivalent - except, I guess, that if `RecordReader` calls `d_input->read(&buf, 1)` in a loop it's N virtual calls, vs 1 virtual call if it calls `d_input->readVarint(val)` and that function can be implemented without hopping through the vtable. If that's the only difference, and that difference is enough to make a big performance difference, I wonder if we should consider using CRTP here instead of using virtual dispatch for reading from the source - it ought to make an even bigger difference...

pablogsal:
> Socket reads keep the existing byte-by-byte fallback

I haven’t spent much time trying to optimize the full socket path, but `SocketBuf` already reads 4 KiB chunks and I don’t expect this particular change to make a big difference there.

> Varints are decoded directly from those buffered bytes instead of calling `read()` once per byte.

Your guess is partly right, but my measurements show that around 80% of the win comes from skipping the generic read loop and tiny `memcpy` for every byte, while reducing the number of virtual calls explains around 20%.

> I wonder if we should consider using CRTP here instead of using virtual dispatch for reading from the source

I prototyped a branch-free static `FileSource` path and measured another 3% improvement, so it looks useful, but I think it should be a separate change because source selection needs to move outside the hot parsing loop.

godlygeek:
> my measurements show that around 80% of the win comes from skipping the generic read loop and tiny `memcpy` for every byte

I can't think of any reason why `memcpy(&dst, src + idx, 1); idx += 1` would be any slower than `dst = src[idx++]` - that's weird.

I'd expect it to inline into exactly the same machine code, as long as we're not hopping through the vtable to do it.



> > I wonder if we should consider using CRTP here instead of using virtual dispatch for reading from the source
> 
> I prototyped a branch-free static `FileSource` path and measured another 3% improvement, so it looks useful, but I think it should be a separate change because source selection needs to move outside the hot parsing loop.

Alright, I think we should consider CRTP here, but you're right, that could be a future improvement.

godlygeek (approved):
LGTM. I don't love that the design shifts varint reading into the `Source` and breaks the abstraction that a `Source` just fetches bytes, but if that abstraction is making things take 75% longer to run it's obviously not worth it 😅

I think there's some future improvements we might be able to make here. Looking at this PR makes me wonder if it's worth switching to CRTP for the Reader/Source boundary. It also makes me wonder if it's worth switching from delimited C-style strings in the capture file to length-prefixed Pascal-style strings, to stop us from needing to look at every byte as we read it...

But either of those are things that can be done with a future improvement, and I see no reason not to land this as it is in the meantime.

PR #1012, opened 2026-08-29
Reduce overallocation in memory-mapped capture files

godlygeek:
When tracking ends normally, shrink a capture file down to the size that was written, dropping any overallocation.

Limit the maximum amount we overallocate by to no more than 4MiB + the page size.

### DIFF
--- a/src/memray/_memray/sink.cpp
+++ b/src/memray/_memray/sink.cpp
@@ -120,6 +120,7 @@ FileSink::writeAll(const char* data, size_t length)
         d_bufferNeedle += toCopy;
         data += toCopy;
         length -= toCopy;
+        d_maxWrittenPos = std::max(d_maxWrittenPos, d_bufferOffset + (d_bufferNeedle - d_buffer));
     }
     return true;
 }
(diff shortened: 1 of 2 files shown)


src/memray/_memray/sink.cpp
```diff
@@ -266,6 +278,9 @@ FileSink::~FileSink()
         if (0 != munmap(d_buffer, BUFFER_SIZE)) {
             LOG(ERROR) << "Failed to unmap output file: " << strerror(errno);
         }
+
+        ::ftruncate(d_fd, d_maxWrittenPos);
```
pablogsal: Can we retry on `EINTR` and log other `ftruncate` errors? Otherwise we silently keep the padding.

pablogsal (approved):

src/memray/_memray/sink.cpp
godlygeek: I figured it was fine to be silent because it does impact correctness, just reporter performance. But I suppose I can log the error and just say in the message that it will slow down report generation.

Good point about EINTR, though.

godlygeek: Addressed both concerns with https://github.com/bloomberg/memray/compare/9921222f75b499dac3a4db9c28980ca46bfdf0fc..9b5262c7552ad967ad473024b9021b6b12ce6f34

PR #1014, opened 2026-08-31
Handle multiarch tuples in the injecter filename

godlygeek:
As of Python 3.15, it's possible (though not required) for an abi3 extension module to also encode a platform tuple in its filename.

Handle this case by checking for both possible filenames, to support both builds that include the new platform info and ones that don't.

See https://github.com/python/cpython/issues/122931

### DIFF
--- a/news/1014.bugfix.rst
+++ b/news/1014.bugfix.rst
@@ -1,0 +1,1 @@
+Fix ``memray attach`` when Memray is built from source for the latest release candidates of Python 3.15.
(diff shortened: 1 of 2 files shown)


sarahmonod:
Also see #995 that shows our CI would fail without this for 3.15

sarahmonod (approved):
LGTM

godlygeek:
> Also see #995 that shows our CI would fail without this for 3.15

And indeed, CI for #995 passes after landing this and rebasing the PR, so the fix is confirmed to work.

PR #1023, opened 2026-09-14
Handle reference tracer replacement during object lifetime tracking

KRRT7:
When another tool replaces or removes Memray's Python reference tracer while `track_object_lifetimes=True`, object destruction events are missed and tracker exit dereferences stale object pointers. `tracemalloc.start()` does this on every supported version, since tracemalloc has used the reference tracer since Python 3.13.

Detect an interrupted session before collecting the surviving objects and raise `RuntimeError`, while still restoring the Python hooks and releasing the active tracker:

- The tracer is registered with the `Tracker` as its data, and the callback only acts for the `Tracker` that installed it. Another tool can hand a stale `(callback, data)` pair back to CPython after that `Tracker` is gone, so the data is treated as an opaque identity.
- On Python 3.15, the `PyRefTracer_TRACKER_REMOVED` notification sets a flag, so a tracer that was removed and later restored is still reported. The event is not in 3.14's headers, despite the docs.
- `getSurvivingObjects` stops the world (a no-op before 3.14, as `createTracker` already does), then raises if the flag is set or `PyRefTracer_GetTracer` no longer returns our callback. On 3.13 and 3.14 the ownership check is the only detection, so a replacement is only caught if it is still in effect at exit. Documented.
- Cleanup only calls `PyRefTracer_SetTracer(NULL, NULL)` if we still own the slot. On main, a tracker exiting after `tracemalloc.start()` silently removes tracemalloc's tracer.
- `__exit__` restores the profile functions and thread descriptors even if collecting the survivors raises.

One residual window remains on free-threaded builds: `PyRefTracer_SetTracer` is stop-the-world plus two plain stores with no compare-and-swap, so a tracer that another thread installs between our ownership check and our set is overwritten. CPython has no atomic way to remove only our tracer, so this is documented as a known limitation rather than worked around. Interposing `PyRefTracer_SetTracer` was evaluated and does not help: tracemalloc's call is a direct branch inside libpython, which the symbol patcher cannot see.

Subprocess tests cover removal, replacement by tracemalloc, removal followed by restoration (3.15), normal cleanup, and a later tracking session, with and without native traces.

Validation:

- Python 3.14.6: the new test file and `test_object_tracking.py`, 21 passed and 4 skipped; full integration suite, 321 passed and 31 skipped.
- Python 3.15.0rc1: 22 passed and 3 skipped.
- Python 3.15.0rc1 free-threaded: 22 passed and 3 skipped.
- Free-threaded race reproducer, 200 iterations of `tracemalloc.start()` during tracker exit: 1 lost replacement without native traces, 0 with, matching the documented window.
- Pre-commit hooks including clang-format 17.0.6, and mypy on the test file, pass.


godlygeek:
This is a considerably bigger change than I'm willing to make to fix a very niche issue. Certainly the changes to how we run coverage analysis are well out of scope for this PR; I'm not interested in adding free-threaded Python 3.15 to our coverage analysis at this time. 

I think what I can get behind is this:

For 3.14 or higher, handle the `PyRefTracer_TRACKER_REMOVED` event by setting a flag, and when `getSurvivingObjects` is called, raise an exception if that flag is set.

For 3.13, *either* just document this as a known shortcoming and position it as a limitation of the CPython APIs present in 3.13, _or_ interpose `PyRefTracer_SetTracer` in the same way as we interpose `PyGILState_Ensure` in hooks.h / hooks.cpp and set the same flag as `PyRefTracer_TRACKER_REMOVED` sets if it's ever called with something other than our tracer function or null.

Hm...

In fact I suppose we could just unconditionally interpose it, rather than have two separate paths for 3.13 vs 3.14+ - so amending those options, I think I'd be fine with either:

1. For 3.14 or higher, handle the `PyRefTracer_TRACKER_REMOVED` event by setting a flag, and when `getSurvivingObjects` is called, raise an exception if that flag is set. For 3.13, just document that bad things will happen if another reference tracer replaces ours part way through tracking.

2. Or, interpose `PyRefTracer_SetTracer` and set a flag if it's ever called with something other than our tracer or a null pointer, and raise an exception if `getSurvivingObjects` is called with that flag set.

1 is cleaner and simpler, 2 is more complex but also handles 3.13 - I guess there's a 1b option as well, where instead of just documenting it as unsafe we actually disable reference tracking on 3.13 entirely and say that it requires 3.14 instead. Though we might want a major version bump if we did that...

godlygeek:
Oh, and - if there's a race on free-threaded builds due to the possibility that something is destroying objects while we're trying to freeze our set of leaked objects, I'm fine with adding a stop-the-world in `getSurvivingObjects` to deal with that.

KRRT7:
since I'm a memray user, in this particular case I'd want it to work as described in #2, will apply.

KRRT7:
Thanks for the direction. I tried option 2 first and it can't detect the case that actually matters: `tracemalloc.start()` calls `PyRefTracer_SetTracer` from inside libpython, and that call is a direct branch within the same image (`bl _PyRefTracer_SetTracer` in `_PyTraceMalloc_Start`, `b _PyRefTracer_SetTracer` in `_PyTraceMalloc_Stop` on the 3.15 dylib; optimized Linux builds use `-fno-semantic-interposition` and distro/uv interpreters link libpython statically). The ELF/Mach-O import-table patching we use for `PyGILState_Ensure` only sees calls from *other* images, so it would never fire for tracemalloc, and tracemalloc has used the ref tracer since 3.13.

So I've reworked this into option 1 plus the ownership check that is available on 3.13:

- The tracer is registered with `data = this`. `trackObject` ignores events whose `data` is not the current owner, since another tool can hand our old `(func, data)` pair back to `PyRefTracer_SetTracer` after that `Tracker` is gone.
- On 3.15+ `PyRefTracer_TRACKER_REMOVED` sets a flag. Note the event is only in 3.15's headers, not 3.14's, despite the docs saying `versionadded:: 3.14`.
- `getSurvivingObjects` stops the world (a no-op before 3.14, same as `createTracker`), then raises `RuntimeError` if the flag is set or `PyRefTracer_GetTracer` no longer returns our callback. On 3.13/3.14 the ownership check is the only detection, so "removed and restored before exit" is not caught there; that's documented.
- Cleanup only calls `PyRefTracer_SetTracer(NULL, NULL)` if we still own the slot. On main, a tracker exiting after `tracemalloc.start()` silently removes tracemalloc's tracer.
- `__exit__` restores the profile hooks even if collecting the survivors raises.

One residual window on free-threaded builds: `PyRefTracer_SetTracer` is stop-the-world plus two plain stores with no compare-and-swap, so if another thread installs a tracer between our ownership check and our set, ours wins and theirs is dropped. Reproduced locally on 3.15.0rc1t with a 200-iteration `tracemalloc.start()`-during-exit loop: 1 lost replacement without native traces, 0 with. An earlier revision avoided that by leaving an inactive callback installed after exit; I dropped it in favour of the simpler check-then-set and documented the limitation instead. `tracemalloc.stop()` has the same clobber in the other direction in-tree, so this really needs a conditional setter in CPython to be fully closed.

All coverage CI and `pyproject.toml` coverage changes are gone. Validated on 3.14.6, 3.15.0rc1, and 3.15.0rc1 free-threaded: the new test file and `test_object_tracking.py` pass on all three, and the full integration suite passes on 3.14.6.

